PR #7, opened 2018-06-28
Dev loc

dithillobothrium:
rectangular matrices test + c_interface.


alazzaro:
There are several things to fix before any merge:
- Makefile.inc is missing the CXXFLAGS. What's # -Werror=c-binding-type?
- We can make the C interface to be generated whenever the CXX is declared, therefore we can drop the option CINT. 
- examples/Makefile cannot include anything from the library. You should consider the examples like an independent package.
- dbcsr.h has to be copied in the directory install when we run "make install", if the file exists.  Consequently, the example should compile only if dbcsr.h exists.
- I don't like the idea to mix the C interface in a C++ example. Could you complete the DBCSR namespace in dbcsr.h?
- Please do not leave any commented part
- Changes in src/mm/dbcsr_mm.F and src/mm/dbcsr_mm_cannon.F are now useless
- rectangular test can be executed with dbcsr_performance_driver

dithillobothrium:
> Makefile.inc is missing the CXXFLAGS. What's # -Werror=c-binding-type?
I thought to use the same flags as CFLAGS. 

> We can make the C interface to be generated whenever the CXX is declared, therefore we can drop the option CINT.
I think CXX and C-interface are not so related, because we can use pure C.

> examples/Makefile cannot include anything from the library. You should consider the examples like an independent package.
Then Makefile will be partially a copy of Makefile.inc? 

> Changes in src/mm/dbcsr_mm.F and src/mm/dbcsr_mm_cannon.F are now useless
> rectangular test can be executed with dbcsr_performance_driver
Revert changes will be enough?

Makefile.inc
```diff
@@ -32,6 +34,10 @@ endif
 LDFLAGS     = $(FCFLAGS)
 LIBS        = -L${SCALAPACK_PATH}/lib -lscalapack -lreflapack -lrefblas
 
+ifneq ($(CINT),)
+FCDEBFLAGS := $(subst -std=f2003,-std=f2008ts,$(FCDEBFLAGS))
```
dev-zero: btw, the GCC devs made it easier and deprecated the f2008ts flag (now part of f2008)

src/dbcsr.h
```diff
@@ -0,0 +1,54 @@
...
+        MPI_Fint fcomm = MPI_Comm_c2f(comm);
+        c_dbcsr_finalize_lib_aux(&fcomm);
+    }
+  
+    void c_dbcsr_distribution_new(void** dist, MPI_Fint* fcomm, int* row_dist, int row_dist_size,
+                                  int* col_dist, int col_dist_size);
+
+    void c_dbcsr_distribution_release(void** dist);
+
+    void test(char* str);
```
dev-zero: what is that `test(...)` function?

src/dbcsr.h
```diff
@@ -0,0 +1,54 @@
+#ifndef DBCSR
+#define DBCSR
+
+#include "mpi.h"
```
dev-zero: shouldn't it be `#include <mpi.h>`?

examples/dbcsr_example_3.cpp
```diff
@@ -0,0 +1,155 @@
...
+   
+     
+    auto fill_matrix = [&](void*& matrix)
+    {
+        int max_row_size = *std::max_element(row_blk_sizes.begin(),row_blk_sizes.end());
+        int max_col_size = *std::max_element(col_blk_sizes.begin(),col_blk_sizes.end());
+        int max_nze = max_row_size * max_col_size;
+
+        vector<double> block;
+        block.reserve(max_nze);
```
dev-zero: Is there actually a difference between `vector<double> block(max_nze)` and the two-line create+reserve which is relevant for this example? If yes, please document.

Makefile.inc
dithillobothrium: if it so, then good. There was some error which was solved when I changed to the f2008ts format.

examples/dbcsr_example_3.cpp
```diff
@@ -0,0 +1,155 @@
...
+                    std::generate(block.begin(), block.end(), [&](){return (double)std::rand()/(double)RAND_MAX;});
+                    c_dbcsr_put_block_d(matrix, i, j, block.data(), block.size());
+                }
+            }
+        }
+    };
+   
+    // create and fill matrix a 
+    void* matrix_a = nullptr;
+    c_dbcsr_create_new_d(&matrix_a, (char*)"matrix a", dist, 'N', row_blk_sizes.data(), row_blk_sizes.size(),
```
dev-zero: Why don't we modify the API to actually take a `const char*` instead of casting the const away? Why not use the C++ `const_cast`?

examples/dbcsr_example_3.cpp
dithillobothrium: yes, I can change. Initially probably I wanted to use push_back.

src/dbcsr.h
dithillobothrium: OK

src/dbcsr.h
dithillobothrium: I will delete this. It was debug function.

examples/dbcsr_example_3.cpp
dithillobothrium: To change to ```const char*``` will be better.

examples/dbcsr_example_3.cpp
```diff
@@ -0,0 +1,155 @@
...
+vector<int> random_dist(int dist_size, int nbins)
+{
+    vector<int> dist(dist_size);
+
+    for(int i=0; i < dist_size; i++)
+    {
+        dist[i] = (nbins-i+1) % nbins;
+    }
+
+    return std::move(dist);
```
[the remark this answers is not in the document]
dithillobothrium: Because move usually can be ambiguous more often. Do I need to use std everywhere?

Makefile.inc
dev-zero: I guess we have to keep f2008ts for the moment, they changed it for gcc >= 7, sorry

examples/dbcsr_example_3.cpp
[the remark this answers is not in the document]
dev-zero: I personally would use `std::` always, but that's personal style

PR #11, opened 2018-07-05
Feature/fypp by submodule

dev-zero:
The advantage is that it is obvious which version of Fypp we use and it
can be easily updated. The disadvantage is that the release process gets
more complicated since we have to create a tarball which includes a
clone of the Fypp submodule, see:

  https://help.github.com/articles/creating-releases/
  https://stackoverflow.com/questions/34719785/how-to-add-submodule-files-to-a-github-release

On the other hand, this can possibly be automated using some CI system
whenever a new tag gets added.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "tools/build_utils/fypp"]
+	path = tools/build_utils/fypp
+	url = https://github.com/aradi/fypp.git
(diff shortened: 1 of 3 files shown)


alazzaro:
There is a clear advantage of using submodules and we should use them. Then we will find a way to solve the other problems...

PR #25, opened 2018-08-21
libcusmm: specify GPU version 

shoshijak:
- Specify GPU version (i.e. K20, P100, ...) instead of ARCH_NUMBER in Makefile.inc
- Determine ARCH_NUMBER from GPUVER
- libcusmm/generate_parameters.py: read the correct parameter file depending on ARCH_NUMBER
- libcusmm/tune.py: choose compilation options (-arch sm_xx) depending on ARCH_NUMBER and cleanup macros rendered obsolete by new hash

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,9 +25,27 @@ LIBNAME      := dbcsr
 LIBRARY      := lib$(LIBNAME)
 default_target: $(LIBRARY)
 
-# Read the configuration ============================================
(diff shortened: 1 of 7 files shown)


alazzaro:
I have updated the P100 parameters to the latest version from CP2K. BTW, I made as default the P100 parameters when GPUVER is not set in Makefile.inc.

Issue #45, opened 2018-10-10
CMake: install target missing


davydden:
I am sure you are aware of this, but CMake build lacks install target. 

Probably, this should be (in case of a general out-of-source build):

- [ ] `<CMAKE_BUILD_DIR>/src/*.mod -> prefix/include/*`
- [ ] `<CMAKE_BUILD_DIR>/src/*.a|so|dylib -> prefix/lib/*`
- [ ] `<CMAKE_SOURCE_DIR>/src/*.h -> prefix/include/*`

dev-zero:
yes, I am aware. Unfortunately I didn't have the time yet to write that one (PR appreciated ;-))

davydden:
that's usually done via
```
SET(TARGETLIBEXPORT "dbcsr")
# 1. Install defined library target
install (TARGETS ${TARGETLIB} EXPORT ${TARGETLIBEXPORT}
         LIBRARY DESTINATION lib)
# 2. Install content of include dir
install (DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/"
         DESTINATION include)
```
however I never did anything with Fortran in Cmake. So I keep this code snippet here in case it helps later.

dev-zero:
thanks, fixed it with #48

PR #48, opened 2018-10-18
add install target for CMake

davydden:
fixes https://github.com/cp2k/dbcsr/issues/45

~~Currently, this works only when building with `-DBUILD_SHARED_LIBS=ON`. I did not dig into CMake documentation to figure out how to fix this for static builds, but I will have a look...~~


davydden:
superseeded by https://github.com/cp2k/dbcsr/pull/55

PR #53, opened 2018-10-18
Feature/remove python eval

shoshijak:
- Change parameters_*.txt files to JSON format and subsequently adapt the autotuning infrastructure to this change. Using 'eval' in python is considered bad practice. 
- Introduce class hierarchy for Kernel_dnt_* classes to avoid code duplication and ease polymorphism

### DIFF
--- a/src/acc/libsmm_acc/libcusmm/collect.py
+++ b/src/acc/libsmm_acc/libcusmm/collect.py
@@ -9,11 +9,11 @@ import json
 from ast import literal_eval
 from kernels.cusmm_dnt import kernel_algorithm
 
-re_mnk    = re.compile("tune_(\d+)x(\d+)x(\d+)_")
-re_winner = re.compile("\nWINNER: \d+ (.+)\n")
-re_gflops = re.compile("# ([0-9.]+) GFlop/s")
-re_errors = re.compile("Number of errors: (\d+)\n")
-re_kernel_descr = re.compile("Kernel_dnt_(\w+)(\(.*\)) , # (\d+\.\d+) GFlop/s")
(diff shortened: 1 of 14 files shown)


src/acc/libsmm_acc/libcusmm/collect.py
```diff
@@ -73,7 +73,7 @@ def process_log(log_fn, mnk, winners):
         return
 
     old_gflops = 0.0
-    if(winners.has_key(mnk)):
+    if(mnk in winners.keys()):
```
dev-zero: no need for the brackets and the `.keys()`:

```python
if mnk in winners:
```

src/acc/libsmm_acc/libcusmm/collect.py
```diff
@@ -5,11 +5,26 @@
...
 import re
+import json
+from ast import literal_eval
+from kernels.cusmm_dnt import kernel_algorithm
 
 re_mnk    = re.compile("tune_(\d+)x(\d+)x(\d+)_")
 re_winner = re.compile("\nWINNER: \d+ (.+)\n")
 re_gflops = re.compile("# ([0-9.]+) GFlop/s")
 re_errors = re.compile("Number of errors: (\d+)\n")
+re_kernel_descr = re.compile("Kernel_dnt_(\w+)(\(.*\)) , # (\d+\.\d+) GFlop/s")
```
dev-zero: regular expression strings should be prefixed with `r'...'` to avoid any accidental character interpretation

dev-zero:
other than that, looks good to me, and definitely more readable than before

dev-zero:
thank you very much!

shoshijak:
thank you for the review!

PR #55, opened 2018-10-19
Feature/cmake install

dev-zero:
@davydden this PR includes your commit from #48 and fixing some additional things.
For example: you can then build a project using DBCSR with the following cmake snippet:

```cmake
cmake_minimum_required (VERSION 3.10)
set(CMAKE_MODULE_PATH "/usr/local/lib64/dbcsr") # adjust to where you installed DBCSR
include(DBCSRTargets)
include(DBCSRCTargets)

enable_language(Fortran CXX)
set(CMAKE_Fortran_FLAGS "-ffree-form -ffree-line-length-none -std=f2003 -fopenmp")

add_executable(dbcsr_example dbcsr_example.F)
target_link_libraries(dbcsr_example dbcsr)

add_executable(dbcsr_example dbcsr_example.cpp)
target_link_libraries(dbcsr_example_cpp dbcsr_c ${MPI_CXX_LIBRARIES})
target_compile_features(dbcsr_example_cpp PRIVATE cxx_std_14)
target_compile_options(dbcsr_example_cpp PUBLIC ${MPI_CXX_COMPILE_FLAGS})
```

CMake will take care of setting proper include & linker flags, even when using static linking.


src/CMakeLists.txt
```diff
@@ -132,14 +133,13 @@ if (WITH_C_API)
...
 if (WITH_C_API)
-  install (TARGETS dbcsr_c EXPORT ${TARGETLIBEXPORT}
-          LIBRARY DESTINATION lib
-          ARCHIVE DESTINATION lib)
-  install (FILES "${CMAKE_SOURCE_DIR}/src/dbcsr.h"
-           DESTINATION include)
+  install(TARGETS dbcsr_c EXPORT dbcsr_c
+    LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+    ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}")
+  install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/dbcsr.h"
```
davydden: it's NOT `CMAKE_CURRENT_SOURCE_DIR`, it should be `CMAKE_SOURCE_DIR`.

davydden: oh, never mind, it looks like you moved the header recently... I tested that patch on release candidate.

davydden: could you check out-of-source build? not that it should matter, but to be sure...

dev-zero: I'm only building out-of-source and would consider in-source building unsupported ;-)
The relation between `CMAKE_CURRENT_SOURCE_DIR` and `CMAKE_SOURCE_DIR` for `src/` is: `CMAKE_CURRENT_SOURCE_DIR=CMAKE_SOURCE_DIR/src`

dev-zero:
@davydden which version of cmake do you have? I'm contemplating on which solution I should pick for solving the parallel-build issue for the tests and the proper solution requires cmake 3.12

src/CMakeLists.txt
davydden: > I'm only building out-of-source and would consider in-source building unsupported ;-)

I fully agree ;-)

> The relation between CMAKE_CURRENT_SOURCE_DIR and CMAKE_SOURCE_DIR for src/ is: CMAKE_CURRENT_SOURCE_DIR=CMAKE_SOURCE_DIR/src

I know that but I am still puzzled how/why `CMAKE_SOURCE_DIR` worked for me in the other PRs... But as I said, does not matter as long as you check that the header is being installed.

davydden:
@dev-zero personally i have `cmake@3.12.2` and it's not a problem to install as I build everything via [Spack](https://github.com/spack/spack) on macOS, Linux desktop and HPC.

However, I have an impression that `3.12` is in general not provided by sysadmins on HPC clusters.

dev-zero:
@davydden thanks, solved it now without 3.12 for the moment. Since we're not installing the tests, it doesn't matter if we get a full library there for the shared objects.

Issue #3, opened 2018-06-22
Update file headers


dev-zero:
Following the discussion on the CP2K-Dev ML: I think we should update the file headers at some point:

1. most of them only mention CP2K, not DBCSR specifically
2. many mention a copyright, but no license

Example from current: https://github.com/cp2k/dbcsr/blob/a6eea29afe755ac0beaa270a637ed63e0724616c/src/dbcsr_api.F#L1-L4

For the first part we need a good wording which declares DBCSR as its own project but retains the link to CP2K. For the second part I guess it would make sense to use a [SPDX License Identifier](https://spdx.org/using-spdx-license-identifier)

oschuett:
+1 for SPDX headers. I've been contemplating this for CP2K as well.

alazzaro:
Do we have an automatic script for that?

dev-zero:
Something like this?

```sh
find . -iname "*.F" | xargs perl -i -0pe 's/^!-+!\n(![^\n]+!\n)+!-+!/NEW HEADER GOES HERE/sm'
```

dev-zero:
Based on the AiiDA header file, how about the following?

```fortran
!--------------------------------------------------------------------------------------------------
! Copyright (c), the CP2K and DBCSR developers groups. All rights reserved.
! This file is part of the DBCSR library.
!
! The code is hosted on GitHub at https://github.com/cp2k/dbcsr
! For further information on the license, see the LICENSE file
! For further information please visit https://dbcsr.cp2k.org
! SPDX-License-Identifier: GPL-2.0+
!--------------------------------------------------------------------------------------------------
```

Too verbose? Drop the `hosted on`?

dev-zero:
simplified it a bit more:
```
!---------------------------------------------------------------------------------------------------
! Copyright (c), the DBCSR developers group. All rights reserved.
! This file is part of the DBCSR library.
!
! For information on the license, see the LICENSE file
! For further information please visit https://dbcsr.cp2k.org
! SPDX-License-Identifier: GPL-2.0+
!---------------------------------------------------------------------------------------------------
```

alazzaro:
Is this change already done?

PR #56, opened 2018-10-19
reformat all file headers

dev-zero:
fixes: #3


dev-zero:
Don't care much about it, but I removed the right-side part since I got tired when editing the banner and it is just decoration.
The dot after the URL on the other hand is really annoying when copy/pasting.

oschuett:
As I said just some suggestions. The final dot on the url makes it (technically) more accurate, but I see that it could confuse people. As for the decoration, I simply like symmetry ;-)

alazzaro:
I don't have any comment, it is fine with me. The only thing is that we have to update the header on the .h, .cu, .py files (and any other code extension).

dev-zero:
ok, updated the PR based on the discussion, also added headers for `.py`, `.cu` and `.cpp`, please merge if you agree

oschuett:
Thanks for keeping the decoration :-)

PR #57, opened 2018-10-22
Feature/autotuning parameter space

shoshijak:
- Document parameter choices for autotuning (hard constraints and heuristics) 
- Allow autotuning procedure to adapt bounds to different GPU cards (previously: only P100)


src/acc/libsmm_acc/libcusmm/tune.py
```diff
@@ -42,6 +42,7 @@ def main():
     param_fn = options.params
     assert param_fn in arch_number.keys(), "Cannot find compute version for file " + param_fn
     arch = arch_number[param_fn]
+    gpu_properties = json.load(open('kernels/gpu_properties.json'))["sm_" + str(arch)]
```
dev-zero: I would recommend something like:

```python
with open('kernels/gpu_properties.json') as fhandle:
    gpu_properties = json.load(fhandle)["sm_{}".format(arch)]
```

to make sure we close that file again at a defined point.

src/acc/libsmm_acc/libcusmm/kernels/cusmm_dnt_largeDB2.py
```diff
@@ -55,20 +54,21 @@ def promising_parameters(m, n, k, gpu):
...
                         # Miniumum number of threads required to have one thread per tile
+                        # i.e., cover the result matrix
                         min_threads = cmax * rmax
                         if threads < min_threads:
                             continue
                         if (min_threads < (threads - 32)):
                             continue  # heuristic: too many threads unused during calculation
 
-                        for w in range(4, (k + 1)/2, 2):  # heuristic: only even numbers
+                        for w in range(4, (k + 1)/2, 2):  # heuristic: even numbers yield better performance
```
dev-zero: warning: the divison here could yield floating point numbers in Python 3, not sure whether that was intended.
Point being: the behavior may differ between Python 2 and 3. Possibly add `from __future__ import division` as a first import to prevent that.

src/acc/libsmm_acc/libcusmm/tune.py
shoshijak: good point!

src/acc/libsmm_acc/libcusmm/kernels/cusmm_dnt_small.py
```diff
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
shoshijak: I made all the divisions floor divisions with `//`, so there shouldn't be a problem. I'm afraid that import statement is confusing to novices.

PR #59, opened 2018-10-23
Feature/reorder parameters

shoshijak:
- Reorder parameters by (m, n, k) in the JSON parameter files (just for readability) 
- Adapt python scripts `generate_parameters.py` and `merge.py` to JSON format of parameter files


src/acc/libsmm_acc/libcusmm/generate_parameters.py
```diff
@@ -19,7 +19,7 @@ def main(argv):
 
     # Read existing parameters
     print("GPU version:\n", options.gpu_version)
-    param_fn = "parameters_" + options.gpu_version + ".txt"
+    param_fn = "parameters_" + options.gpu_version + ".json"
```
dev-zero: In terms of readability I personally prefer [format-strings](http://pyformat.info) instead:

```python
param_fn = "parameters_{}.json".format(options.gpu_version)
```

... maybe something for next time.

src/acc/libsmm_acc/libcusmm/generate_parameters.py
```diff
@@ -21,9 +22,12 @@ def main(argv):
     print("GPU version:\n", options.gpu_version)
     param_fn = "parameters_" + options.gpu_version + ".json"
     with open(param_fn) as f:
-        content = f.read().splitlines()
-    print("About to process", len(content), "lines from file", param_fn)
-    parameters = get_parameters_from_file(content)
+        all_kernels = [params_dict_to_kernel(**params) for params in json.load(f)]
+    print("About to process", len(all_kernels), "kernels from file", param_fn)
```
dev-zero: note: this `print` will print a tuple on Python 2 since `print` is not a function there.

To get the same output in Python 2 and 3, you must use

```python
from __future__ import print_function
```

as the very first import in this file.

shoshijak: `from __future__ import print_function` is already the first import of this file :)

src/acc/libsmm_acc/libcusmm/generate_parameters.py
shoshijak: ok, I'll keep it in mind for next time

src/acc/libsmm_acc/libcusmm/generate_parameters.py
dev-zero: I'm sorry, scrolled up one too many and clicked the "View" button of another file :-\

PR #64, opened 2018-10-29
[WIP] Extension of C/C++ interface and new c-header 'dbcsr_c.h'

jkn93:
+ new routines:
  c_dbcsr_copy_d, c_dbcsr_add_d, c_dbcsr_finalise_lib_silent,
  c_dbcsr_trace_ab_d, c_dbcsr_trace_a_d, c_dbcsr_set_diag_d,
  c_dbcsr_set_d, c_dbcsr_filter_d, c_dbcsr_get_block_d,
  c_dbcsr_multiply_eps_d, c_dbcsr_gershgorin_estimate_d,
  c_dbcsr_scale_d, c_dbcsr_read_d, c_dbcsr_write_d, c_dbcsr_maxabs_d
+ new header can be included in multiple source-files


src/dbcsr_api_c.F
```diff
@@ -300,4 +300,377 @@ SUBROUTINE c_dbcsr_multiply_d(transa, transb, alpha, c_matrix_a, c_matrix_b, bet
...
+      TYPE(c_ptr), INTENT(in)                            :: c_matrix_a
+
+      TYPE(dbcsr_type), POINTER                          :: matrix_a, matrix_b
+      !REAL(kind=c_double), INTENT(inout), value            :: retval
+      REAL(kind=c_double), INTENT(inout)                   :: retval
+
+      CALL c_f_pointer(c_matrix_a, matrix_a)
+      CALL c_f_pointer(c_matrix_b, matrix_b)
+      retval = 0.0
+      call dbcsr_trace(matrix_a, matrix_b, retval)
```
dev-zero: Please maintain a consistent style wrt mixing all-upper and lower-case, and further down wrt the indentation. Running `make pretty` can help with that.

dev-zero:
@jkn93 can I merge your header file with the already present `src/dbcsr.h`?

jkn93:
'inline' should take care of that. Your test worked fine:

echo '#include <mpi.h>' > test_a.cc     
echo '#include <dbcsr.h>' >> test_a.cc
echo '#include <mpi.h>' > test_b.cc   
echo '#include <dbcsr.h>' >> test_b.cc
mpic++ -Isrc -c test_{a,b}.cc
mpic++ -shared -o test_api.so test_{a,b}.o

dev-zero:
interesting, because by C++ standard should a `constexpr` function be expanded to an inline function

dev-zero:
@jkn93 as a word of precaution: your `dbcsr_c.h` and the existing `dbcsr.h` have to be merged since we can't have two different C API's for DBCSR.
Wrt the C++ header: it might be a good idea to hide some of the implementations away in a `.cxx` file since they will most likely not be inlined. Or, since that would mean tying a binary of the DBCSR lib to a specific C++ compiler, using a private header instead.
For the class design itself, I am missing for example move  constructors, but also the semantics for the constructors are not clear enough:
one hand you write that `Each object of `DistBCSR` contains a pointer to a valid DBCSR_Environment-object`, on the other hand is the default constructor for `DistBCSR` not disabled.
Furthermore is the order of destruction of the `DBCSR_Environment` and `DistBCSR` not enforced (not sure though whether we should do it by using semantics - getting a const pointer - or by using smart pointers), and some of the C++ Core Guidelines seem to get ignored (for example http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c62-make-copy-assignment-safe-for-self-assignment) and I am not sure whether this is intentional.

Since you are clearly working (hard) on it, I'm marking this PR as `[WIP]` for now.

jkn93:
@dev-zero I'm done with the initial version of the DistBCSR-class after adding some further functionality and safety-checks. I've also added another example (DistBCSR_example_3.cpp). With the current features I managed to run Niklassens trace-resetting and implicit purification algorithms successfully.
So far, it's self-contained and still header-only, i.e., no need to compile C++-code for the library itself.

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1192 @@
...
+      ref.free();
+    };
+
+
+    DistBCSR(size_t nrow_in, size_t ncol_in, std::vector<int>& row_dims_in, std::vector<int>& col_dims_in, DBCSR_Environment* dbcsr_env_in,
+             double thr_in=0.e0, const std::string& mname_in="default matrix name"){
+      this->init(nrow_in,ncol_in,row_dims_in,col_dims_in,dbcsr_env_in,thr_in,false,mname_in);
+    };
+
+    DistBCSR(size_t ldim, std::vector<int>& dims_in, DBCSR_Environment* dbcsr_env_in, double thr_in=0.e0,
```
davydden: an observer note: 

there are some places where `const` values are marked as such, but there are also a lot of places where they aren't. 

I bet here `dims_in`, `thr_in` and `add_zero_diag` are all `const`. I don't know what `ldim` means, but most likely it's also const.

Any reason why you pass an environment as a pointer? It would probably make sense to take this object (which wraps MPI comm and its grid?) as `std::shared_ptr<const DBCSR_Environment> dbcsr_env`. Most likely different matrices that use the same grid shall not be allowed to modify it anyhow (`const`-ness again). From the user perspective, it would be good if such things were reflected in the C++ interface.

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1192 @@
...
+    std::string mname;
+    size_t nrow;
+    size_t ncol;
+    std::vector<int> row_dims;
+    std::vector<int> col_dims;
+    double dbcsr_thresh;
+
+  public:
+    DistBCSR(){
+      dbcsr_matrix = nullptr;
```
davydden: it would help reading and documenting the code if you separate declaration from definition within the same header.

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1192 @@
...
+
+/*
+ * Wrapper-class for dbcsr
+ */
+class DistBCSR {
+
+  private:
+    DBCSR_Environment* dbcsr_env;
+
+    dm_dbcsr dbcsr_matrix;
```
davydden: don't know what maintainers plan to do about documentation, but `doxygen` on each member variable and functions of this class would certainly not harm from a user perspective.

jkn93: I've added brief comments to the class-members

src/DistBCSR.hpp
jkn93: It certainly will, but as a first step I wanted to keep my addition compact. It'll also be the first C++-source in the library (I assume), which may also require an overhaul of some make/cmake files.
@dev-zero  I suppose the maintainers will be better suited to make these modifications.

alazzaro:
Thanks for all your tremendous work in including the C++ API in DBCSR! Can I ask you in which project you will use DBCSR?
I will review all your changes. However, this cannot be soon since we are working in release v1.0 of DBCSR. Your changes will go in v2.0. The current ETA is January 2019.

src/DistBCSR.hpp
dev-zero: We already have CUDA files and also C++ already in the tree (as part of the CUDA SMM library), C++ sources are therefore no problem at all.

davydden:
@alazzaro 

>  However, this cannot be soon since we are working in release v1.0 of DBCSR. Your changes will go in v2.0. The current ETA is January 2019.

one way of still merging this is if you branch off for `1.0.release_candidate` and keep accepting PR to `develop` branch.  Then all the bug-fixes will go to both `develop` and `1.0.release_candidate`. One can also do such things with [gitflow workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow).

src/DistBCSR.hpp
davydden: > It certainly will, but as a first step I wanted to keep my addition compact

it would be as compact as now, I only suggested to separate definitions from declarations and still keep things in one header.

davydden: > I'm not even sure how to do this while avoiding multiple definitions.

You are right.

jkn93:
> Thanks for all your tremendous work in including the C++ API in DBCSR! Can I ask you in which project you will use DBCSR?
> I will review all your changes. However, this cannot be soon since we are working in release v1.0 of DBCSR. Your changes will go in v2.0. The current ETA is January 2019.

@alazzaro your're welcome, but then again: it was a selfish act.
I've written a linear-algebra class that handles matrices in different formats (dense/sparse) on different computing architectures (CPU,GPU, distr. linalg via libElemental and DBCSR) and offers some additional functionality for QC-calcs. It's part of a new cooperative project we started with the aim to run large-scale semi-empirical MD. I'll let you know when any publications will emerge.

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1414 @@
...
+      }
+    };
+
+    /*! \brief Gershdoring estimate
+     *
+     * Returns estimate of min/max eigenvalues of matrix
+     * based on Gershgorin discs
+     *
+     */
+    void gershgorin_estimate(double& eps0, double& epsn){
```
davydden: both are `const` i guess.

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1414 @@
...
+      std::vector<double> red_diags(nrow,0.e0);
+    
+      dbcsr::gershgorin_estimate(this->dbcsr_matrix,row_dims.data(),(int)col_dims.size(),(int)nrow,sums.data(),diags.data());
+    
+      // gather results
+      MPI_Allreduce(sums.data(),red_sums.data(), ((int)nrow), MPI_DOUBLE, MPI_SUM, dbcsr_env->dbcsr_group);
+      MPI_Allreduce(diags.data(),red_diags.data(), ((int)nrow), MPI_DOUBLE, MPI_SUM, dbcsr_env->dbcsr_group);
+    
+      double disc_min[2];
+      disc_min[0] =  9999.e0;
```
davydden: `std::numeric_limits<double>::max()`

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1414 @@
...
+      disc_min[1] = -9999.e0;
```
davydden: `std::numeric_limits<double>::lowest`

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1414 @@
...
+      assert(this->dbcsr_env != nullptr);
+      std::vector<double> the_row(ncol);
+      std::vector<double> loc_row(ncol);
+      std::fill(the_row.begin(),the_row.end(),0);
+      std::fill(loc_row.begin(),loc_row.end(),0);
+    
+      int which_row = 0;
+      int rr = 0;
+      while(which_row < (int)row_dims.size()){
+        int bdim_act = row_dims[which_row];
```
davydden: const int &

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1414 @@
...
+        which_row++;
+        rr += bdim_act;
+      }
+      if (which_row == (int)row_dims.size()){
+        printf("\n  Error: Illegal row-index!\n\n");
+        exit(1);
+      }
+      int roff = ((int)row) - rr;
+    
+      int max_dimc = *(std::max_element(col_dims.begin(),col_dims.end()));
```
davydden: const

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1414 @@
...
+      c_dbcsr_get_stored_coordinates(dbcsr_matrix, blk_row, blk_col, &blk_proc);
+      return (blk_proc == dbcsr_env->mpi_rank ? true : false);
+    };
+
+    /*! \brief Returns FORTRAN-handle to dbcsr-matrix
+     */
+    dm_dbcsr& get_dbcsr(){return dbcsr_matrix;};
+    /*! \brief Returns DBCSR_Environment object
+     */
+    std::shared_ptr<const DBCSR_Environment> get_env(){return dbcsr_env;};
```
davydden: const ... get_env() const...

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1414 @@
...
+    /*! \brief Returns number of rows in dense matrix
+     */
+    size_t get_nrow(){return nrow;};
```
davydden: `...get_nrow() const`

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1414 @@
...
+    /*! \brief Returns number of columns in dense matrix
+     */
+    size_t get_ncol(){return ncol;};
```
davydden: `... get_ncol() const...`

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1414 @@
...
+    /*! \brief Returns dimensions of row-blocks
+     */
+    std::vector<int>& get_row_dims(){return row_dims;};
```
davydden: `const ... get_row_dims() const ...`

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1414 @@
...
+    /*! \brief Returns dimensions of column-blocks
+     */
+    std::vector<int>& get_col_dims(){return col_dims;};
```
davydden: `const ... get_col_dims() const...`

src/DistBCSR.hpp
```diff
@@ -0,0 +1,1414 @@
...
+    /*! \brief Returns sparse-threshold
+     */
+    double get_thresh(){return dbcsr_thresh;};
```
davydden: `...get_thresh() const...`

jkn93:
@davydden thanks for the comments, although I didn't consider all of them.

I initially considered `DBCSR_Environment` more like a `struct`, but changed it into a more 'proper' class now. Still, it handles MPI-stuff, so it can't be fashioned to be absolutely safe  (e.g., `MPI_Finalize()`).

Most of your `const`-requests would force me to typecast the variable to non-const in the next line since it's only passed to some Fortran-routine (same for `unsigned int`).
For a user of this class it's certainly important to have `const`-arguments, resting assured that we at least intent not to modify the respective object (whatever happens within the Fortran-routine).
I don't think that `const`-vars within the member-functions are a safety-threat as long as nobody tries to actively sabotage the code.

Just as a side note: You may have noticed that I'm not necessarily a stickler when it comes to C++-conventions, as, for example, I still prefer `printf` over the (in my opinion) awful i/o-streams.
From a purists C++ safety-perspective, I think the biggest concern would be the handling/calling of Fortran-objects/functions.

Finally I've added doxygen-comments.

@dev-zero I've added an option-string `pattern` to `DBCSR_Environment` to allow for different distribution-patterns. So far, only the given default is valid. You may want to add further options based on your experiences.

davydden:
Thanks for incorporating my comments.

> Most of your const-requests would force me to typecast the variable to non-const in the next line since it's only passed to some Fortran-routine (same for unsigned int).
For a user of this class it's certainly important to have const-arguments, resting assured that we at least intent not to modify the respective object (whatever happens within the Fortran-routine).
I don't think that const-vars within the member-functions are a safety-threat as long as nobody tries to actively sabotage the code.

Oh, I see. That’s fine as is then.

dithillobothrium:
> Most of your `const`-requests would force me to typecast the variable to non-const in the next line since it's only passed to some Fortran-routine (same for `unsigned int`).
> For a user of this class it's certainly important to have `const`-arguments, resting assured that we at least intent not to modify the respective object (whatever happens within the Fortran-routine).
> I don't think that `const`-vars within the member-functions are a safety-threat as long as nobody tries to actively sabotage the code.

Will it be possible to use these methods from const instances of the class, or passed by const references or pointers?

jkn93:
@dithillobothrium i've just defined some member-funcs as const that should actually be const. Still,
some require typecasts to non-const for Fortran.

jkn93:
@davydden I've added the consts.
+ mpi-comm/id-string: probably a matter of taste, so far I've only used the id-string, not the comm.
+ inv. Hadamard: it's not a properly defined func. I've only added it since I use it within a project.
+ assert/inv: this check is sort of arbitrary. It will certainly prevent infs within electronic structure codes,
but in general I would assume that users of that function would execute their own checks. I'd leave this decision to the maintainers.
BTW: feel free to submit pull-reqs to my fork...

alazzaro:
Closing it in favour of the #348

PR #70, opened 2018-10-31
Fixed typos among are incorrect routine names

hfp:
Fixed some typos among are incorrect routine names (routineN) for a number of block-ops.


dev-zero:
Thanks!

PR #74, opened 2018-11-02
Joaat hash to produce CRC32 hashes

hfp:
Call libxsmm_hash to implement joaat_hash. This function produces CRC32 hashes and leverages specific CPU instructions. The CRC32 hash may also improve the hash quality and distribution. LIBXSMM provides this function since v1.9.0-6 as part of the Fortran interface (tested in CP2K/intel fork since then). LIBXSMM v.1.10 will be the first release with the Fortran interface for libxsmm_hash.


dev-zero:
@hfp please `squash & merge` when finished

PR #78, opened 2018-11-02
Improved Makefile.inc

hfp:
Improved Makefile.inc (without the goal to make it sophisticated). Though, the CMake based build system is likely where sophistication goes into.

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -36,16 +36,26 @@
 
 #######################################################################################
 #
-# Optional configurations:
+# Optional configuration:
 #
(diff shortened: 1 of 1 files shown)


alazzaro:
I think your changes are reasonable to have two compilers and different versions, however I would not go any further (already people asking for Cray wrappers support)... Cmake (or user level changes) is the way to go for that...

hfp:
Indeed, the change is just to reduce the diffs I have on record. Before throwing it away, I thought it can go into. Thanks for the comment!

PR #80, opened 2018-11-03
Use LIBXSMM (or MKL) for out-of-place and in-place transpose (block-ops).

hfp:
* Functionality becomes available with LIBXSMM v1.10 although the actual interface was there forever. The version barrier was necessary because of portability issues with C_LOC (worked around with libxsmm_ptr1 and libxsmm_ptr2). 
* If LIBXSMM is not used (or because of insufficient version), MKL's omatcopy and imatcopy functions are used (this could be ported to OpenBLAS as well). Note that mkl_trans.fi is not used as the header is obolescent; use an implicit "interface" instead (link-time).
* CMake infrastructure is extended to add __MKL to the list of preprocessor symbols (depends on BLA_VENDOR or list of BLAS_LIBRARIES); this check is done in a backward compatible fashion (older CMake).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,11 @@ endif ()
 # PACKAGE DISCOVERY:
 
 find_package(BLAS REQUIRED)
+
+if (BLAS_VENDOR_FOUND MATCHES "Intel" OR BLAS_VENDOR MATCHES "Intel" OR BLAS_LIBRARIES MATCHES "mkl_")
+  add_definitions(-D__MKL)
+endif ()
+
 find_package(LAPACK REQUIRED)
 find_package(PkgConfig)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ endif ()
 
 find_package(BLAS REQUIRED)
 
-if (BLAS_VENDOR_FOUND MATCHES "Intel" OR BLAS_VENDOR MATCHES "Intel" OR BLAS_LIBRARIES MATCHES "mkl_")
+if (BLA_VENDOR_FOUND MATCHES "Intel" OR BLA_VENDOR MATCHES "Intel" OR BLAS_LIBRARIES MATCHES "mkl_")
   add_definitions(-D__MKL)
 endif ()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,11 +43,6 @@ endif ()
 # PACKAGE DISCOVERY:
 
 find_package(BLAS REQUIRED)
-
-if (BLA_VENDOR MATCHES "Intel" OR BLAS_LIBRARIES MATCHES "mkl_")
-  add_definitions(-D__MKL)
-endif ()
-
 find_package(LAPACK REQUIRED)
 find_package(PkgConfig)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ endif ()
 
 find_package(BLAS REQUIRED)
 
-if (BLA_VENDOR_FOUND MATCHES "Intel" OR BLA_VENDOR MATCHES "Intel" OR BLAS_LIBRARIES MATCHES "mkl_")
+if (BLA_VENDOR MATCHES "Intel" OR BLAS_LIBRARIES MATCHES "mkl_")
   add_definitions(-D__MKL)
 endif ()
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ addons:
 env:
   global:
   # LIBXSMM: update to v1.10 or later
-  - LIBXSMM_VERSION=531c000d1218f5925aaf3ea34dca8eb3033d2dff
+  - LIBXSMM_VERSION=f1171cc08fbe8f65a2aad6d6d9af71ceb975ef91
 
 install:
   # Update cmake to the minimal required version (shamelessly copied from the boost travis config):
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ addons:
 env:
   global:
   # LIBXSMM: update to v1.10 or later
-  - LIBXSMM_VERSION=531c000d1218f5925aaf3ea34dca8eb3033d2dff
+  - LIBXSMM_VERSION=eb4e85c6c7ef080658d1a8b36cc27be13afd9694
   - PKG_CONFIG_PATH=${TRAVIS_BUILD_DIR}/libxsmm-${LIBXSMM_VERSION}/lib
 
 install:
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -111,6 +111,10 @@ if (USE_SMM MATCHES "libxsmm")
   target_compile_definitions(dbcsr PRIVATE __LIBXSMM)
 endif ()
 
+if (BLAS_LIBRARIES MATCHES "mkl_")
+  target_compile_definitions(dbcsr PRIVATE __MKL)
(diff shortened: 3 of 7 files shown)


CMakeLists.txt
```diff
@@ -43,6 +43,11 @@ endif ()
 # PACKAGE DISCOVERY:
 
 find_package(BLAS REQUIRED)
+
+if (BLA_VENDOR_FOUND MATCHES "Intel" OR BLA_VENDOR MATCHES "Intel" OR BLAS_LIBRARIES MATCHES "mkl_")
+  add_definitions(-D__MKL)
```
dev-zero: unless we need this flag also in tests and/or examples I would prefer to move this to `src/CMakeLists.txt` and use the new-style `target_compile_definitions(dbcsr PRIVATE __MKL)` instead.

hfp: Agree, and I tried this first but did not succeed. It appears that `find_package(BLAS REQUIRED)` must run prior to the inserted code. Can the inclusion of CMakeLists.txt be moved post the package search, or do you have a solution?

dev-zero: What I meant is: add the the whole configuration

```cmake
if (BLA_VENDOR_FOUND MATCHES "Intel" OR BLA_VENDOR MATCHES "Intel" OR BLAS_LIBRARIES MATCHES "mkl_")
  target_compile_definitions(dbcsr __MKL)
endif ()
```

to `src/CMakeListst.txt` instead.

Reading the [source of FindBlas](https://github.com/Kitware/CMake/blob/master/Modules/FindBLAS.cmake) it seems that `BLAS_VENDOR_FOUND` is not there. After a bit of googling this might be something found in custom `FindBLAS.cmake` implementions from other projects. Would the following be enough?

```cmake
if (BLAS_LIBRARIES MATCHES "mkl_")
  target_compile_definitions(dbcsr __MKL)
endif ()
```

`BLA_VENDOR` seems to be something set by the user, therefore if the user sets `BLA_VENDOR=intel`, then `BLAS_LIBRARIES` should contain `*mkl* in the end.

hfp: Moving the following block into `src/CMakeListst.txt` does not work (as I said, I tested this before sending the PR). It does not seem to work because none of the condition variables (if-condition) are defined inside of `src/CMakeListst.txt`. It seems that `find_package(BLAS REQUIRED)` runs after `src/CMakeListst.txt` is processed. I will double-check if there is a workaround.

```cmake
if (BLA_VENDOR_FOUND MATCHES "Intel" OR BLA_VENDOR MATCHES "Intel" OR BLAS_LIBRARIES MATCHES "mkl_")
  target_compile_definitions(dbcsr __MKL)
endif ()
```
Indeed, I overlooked that `BLA_VENDOR_FOUND` is not part of the official CMake source code; I will remove it. However if `BLA_VENDOR` matches, it's for sure correct since DBCSR requires BLAS and we would never reach the test if that was failing earlier.

hfp: I checked again, and it's like I said on my side (does not work). I used `make VERBOSE=1`to check if `-D__MKL` appears, and it does not if that block sits in `src/CMakeLists.txt`.

dev-zero:
```
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 4fddb07b..2cb863f9 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -112,6 +112,12 @@ if (USE_SMM MATCHES "libxsmm")
   target_compile_definitions(dbcsr PRIVATE __LIBXSMM)
 endif ()
 
+message("BLA_VENDOR: " ${BLA_VENDOR})
+message("BLAS_LIBRARIES: " ${BLAS_LIBRARIES})
+if (BLA_VENDOR MATCHES "Intel" OR BLAS_LIBRARIES MATCHES "mkl_")
+  target_compile_definitions(dbcsr PRIVATE __MKL)
+endif ()
+
 if (USE_CUDA)
   set(CMAKE_CUDA_FLAGS "-arch=sm_${CUDA_ARCH_NUMBER} --cudart static")
   target_link_libraries(dbcsr PUBLIC ${CUDA_LIBRARY} cuda nvrtc)
```

and then

```console
$ cmake -DBLA_VENDOR=Intel10_64lp .. && make VERBOSE=1
[...]
-- Checking for module 'blas'
--   Found blas, version 0.2.20
-- Found BLAS: /usr/lib/libopenblas_openmp.so (found version "0.2.20") 
-- Checking for module 'lapack'
--   Found lapack, version 3.7.0
-- Found LAPACK: /usr/lib/libreflapack.so;/usr/lib/libopenblas_openmp.so  
-- Found MPI_C: /usr/lib64/libmpi.so (found version "3.1") 
-- Found MPI_CXX: /usr/lib64/libmpi_cxx.so (found version "3.1") 
-- Found MPI_Fortran: /usr/lib64/libmpi_usempif08.so (found version "3.1") 
-- Found MPI: TRUE (found version "3.1") found components:  C CXX Fortran 
-- Using BLAS for Small Matrix Multiplication
-- Found OpenMP_C: -fopenmp (found version "4.5") 
-- Found OpenMP_CXX: -fopenmp (found version "4.5") 
-- Found OpenMP_Fortran: -fopenmp (found version "4.5") 
-- Found OpenMP: TRUE (found version "4.5")  
-- FYPP preprocessor found.
BLA_VENDOR: $BLA_VENDOR
BLAS_LIBRARIES: $BLAS_LIBRARIES
Tests will run with 2 MPI ranks and 2 threads each
-- Configuring done
-- Generating done
-- Build files have been written to: /home/tiziano/work/dbcsr/build
[...]
[ 51%] Building Fortran object src/CMakeFiles/dbcsr.dir/acc/dbcsr_acc_init.F.o
cd /home/tiziano/work/dbcsr/build/src && /usr/bin/gfortran -D__MKL -D__MPI_VERSION=3 -D__STATM_TOTAL -D__parallel -D__SHORT_FILE__=\"dbcsr/acc/dbcsr_acc_init.F\" -I/home/tiziano/work/dbcsr/src/base -I/home/tiziano/work/dbcsr/src -I/home/tiziano/work/dbcsr/build/src -I/usr/include -I/usr/lib64  -ffree-form -ffree-line-length-none -std=f2003 -O3 -funroll-loops   -fopenmp -pthread -c /home/tiziano/work/dbcsr/build/src/acc/dbcsr_acc_init.F -o CMakeFiles/dbcsr.dir/acc/dbcsr_acc_init.F.o
[...]
```

which shows two things:

* at least `BLA_VENDOR` is present in `src/CMakeLists.txt`
* even if `BLA_VENDOR=Intel...` is set, CMake may still pick up something else without error'ing out

I therefore modified it to the following:

```
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 4fddb07b..e29e2056 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -112,6 +112,10 @@ if (USE_SMM MATCHES "libxsmm")
   target_compile_definitions(dbcsr PRIVATE __LIBXSMM)
 endif ()
 
+if (BLAS_LIBRARIES MATCHES "mkl_")
+  target_compile_definitions(dbcsr PRIVATE __MKL)
+endif ()
+
 if (USE_CUDA)
   set(CMAKE_CUDA_FLAGS "-arch=sm_${CUDA_ARCH_NUMBER} --cudart static")
   target_link_libraries(dbcsr PUBLIC ${CUDA_LIBRARY} cuda nvrtc)
```

hfp: Thank you, I can reproduce. I will send an updated PR!

PR #83, opened 2018-11-05
Feature/cmake configure file

dev-zero:


dev-zero:
using MPI and OpenMP targets breaks building with CUDA since those flags end up in being passed to `nvcc` which chokes for example on `-fopenmp`, see https://gitlab.kitware.com/cmake/cmake/issues/17929 for a similar issue with `-pthread` (from MPI).
This will be fixed in cmake-3.13.0, and backport is not possible due to the missing `$<COMPILER_LANGUAGE:..>:SHELL` ... hurray

Issue #52, opened 2018-10-18
setup bug-reporting templates


dev-zero:
with examples on how to obtain the required information we need to be able to reproduce the problem for:
* MPI implementation and version
* compiler vendor, version, flags
* OS/platform
* FFT implementation and version
* BLAS/LAPACK implementation and versions
* CUDA compiler version and target platform

alazzaro:
@dev-zero There is no need to use FFT

dev-zero:
hmm, I wonder whether I wanted to open this bug on cp2k/cp2k instead, but yeah, doesn't make sense

PR #89, opened 2018-11-07
Update issue templates

dev-zero:
fixes: #52


dev-zero:
@alazzaro happy with this?

alazzaro:
Looks OK. Would it be useful to mention that we need a stacktrace or it is too technical (in principle we can make on how own if we are able to reproduce the error)?

PR #90, opened 2018-11-09
Require LIBXSMM v1.8.2 or later

hfp:
Including libxsmm_config.h only works with LIBXSMM v1.8.2 and later. Otherwise, older versions of LIBXSMM still work if this header is not included (prior to v1.8.2 the header was not suitable for Fortran).

### DIFF
--- a/src/base/dbcsr_base_uses.f90
+++ b/src/base/dbcsr_base_uses.f90
@@ -50,4 +50,7 @@
 ! Allows macro-toggles (in addition to parameters).
 #if defined(__LIBXSMM)
 #include <libxsmm_config.h>
+#if !defined(LIBXSMM_CONFIG_VERSION)
+#error LIBXSMM v1.8.2 or later is required!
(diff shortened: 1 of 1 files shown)


dev-zero:
shouldn't version checks be part of the build system?

dev-zero:
... ok, we don't have a `configure` step in the pure-make build system

hfp:
In terms of CMake we already require v1.10 (since no earlier release provides `.pc`-files). However, as noticed with https://github.com/cp2k/cp2k/pull/86, if an older version is manually incorporated (per Makefile) - misleading compiler errors are produced. This is helpful in the latter case.

alazzaro:
I think it is good to have it as Hans did (at least from the CP2K point of view...)

hfp:
As far as I can see, the test for codecov/project seems to be a false positive (-0.01%). I think I will merge if nobody objects.

PR #97, opened 2018-11-19
Ensure DBCSR_ASSERT is used without side-effects

hfp:
Introduced CMake option USE_NDEBUG=ON|OFF (default: OFF). Test that DBCSR_ASSERT can be elided without side-effects (see dbcsr_base_uses.f90).


hfp:
Fails due to missing the watermark for code-coverage by 0.01% (perhaps codecov can be configured to use a range to avoid fluctuation a.k.a. hysteresis).

dev-zero:
yeah, this can be safely ignored (it's annoying though).
FYI, I'm going to restructure the Travis configuration again to enable testing on OSX: https://github.com/cp2k/dbcsr/pull/98/commits/00c67923e0b54052c5a0f55df7e436cb85b94231

Does it make sense to add the `NDEBUG`-test as another axis in the build matrix? Possibly excluding some combinations?

dev-zero:
wrt codecov precision, it seems we can only reduce the precision like that:

```yml
coverage:
  precision: 1
  round: down
  range: "57...100"
```

from https://docs.codecov.io/docs/codecov-yaml

hfp:
>>> Does it make sense to add the NDEBUG-test as another axis in the build matrix? Possibly excluding some combinations?

Maybe it does. However, we quickly approach Travis' resource grant for Open Source projects (which is already generous). I think it's about ~45m (if I am not mistaken). What I did for another project is "mixing" different tests to stay below the limit, and have separate/full coverage on own/additional resources. Setting up test mixtures may prefer "exotic" setups to make Travis as picky as possible, and to get the most out of these tests.

hfp:
Thanks for the feedback!

I will update the PR to always define NDEBUG i.e., USE_NDEBUG=ON rather than OFF.

hfp:
I will also remove the NDEBUG-specific test. As this concern will be covered by Make-based builds/tests.

dev-zero:
@alazzaro it's not only that those tests are dead code, but they are rarely side-effect free if you are not super careful.
@alazzaro will do

CMake currently does never build with `-DNDEBUG`, should I change that?

hfp:
> CMake currently does never build with -DNDEBUG, should I change that?

FYI, the updated PR just does this.

alazzaro:
> CMake currently does never build with `-DNDEBUG`, should I change that?

Yes, please. Let's only keep it on the makefile for CHECKS=1. In principle, if CI found a bug, then we can switch to drop NDEBUG, otherwise I will keep the default configuration.

dev-zero:
... and should we expose `NDEBUG` really as separate option in CMake or tie it to the `BUILD_TYPE`(Release, Debug, Coverage)?

alazzaro:
> ... and should we expose `NDEBUG` really as separate option in CMake or tie it to the `BUILD_TYPE`(Release, Debug, Coverage)?

Well, I will go for the later, which is a nice match with CHECKS=1  of the makefile...

hfp:
I will close the PR.

dev-zero:
@hfp sorry :(

hfp:
No worries!

Btw, good news for DBCSR is that it passes with and without NDEBUG defined. Bad news is (at least with my tests), CP2K does not pass with `-DNDEBUG`.

alazzaro:
@hfp Just to understand, is -DNDEBUG used for DBCSR only when compiling with CP2K or is it used all over in CP2K? I mean, DBCSR is still not fully tested in the CI and we have to rely on CP2K to have a better coverage, so I wonder if you see something wrong in CP2K because of DBCSR...

hfp:
The change I proposed in the CMake-file was private to DBCSR's translation units. Promoting the setting to for instance "Release builds" (using CMake) is also private. I beleive the latter is the current plan. Regarding GNU Make, everyting is possible as well. It is even possible to define NDEBUG for CP2K's source, but not define it for DBCSR (requires to enumerate source file by name or glob by sub/directory).

alazzaro:
Sorry, my question was related to your comment:

> Btw, good news for DBCSR is that it passes with and without NDEBUG defined. Bad news is (at least with my tests), CP2K does not pass with -DNDEBUG.

alazzaro:
To be more specific, in your tests NDEBUG is used in CPASSERT?

hfp:
My tests were based on CP2K with the version of DBCSR which is currently part of CP2K/master. I used `-DNDEBUG` for the entire build incl. the test-drivers (unit test, etc.). And yes, my tests elided CPASSERT in case NDEBUG was defined.

hfp:
See also https://github.com/cp2k/cp2k/blob/master/src/base/base_uses.f90#L24

Just add `-DNDEBUG` to your preferred ARCH-file and run all regtests for CP2K/master, which fails on my end for the majority of tests. I found this true for all tool chains that I checked with (GFortran and IFORT).

alazzaro:
OK, and do you know if it is something wrong with DBCSR (DBCSR_ASSERT), or it is due to a CPASSERT failure? A way to test it is to add -DNDEBUG in the DBCSR makefile directly and not use it in the CP2K arch file...

hfp:
Since DBCSR passes its own tests with NDEBUG defined, I suspect the issue is in CP2K source code (CPASSERT).

> A way to test it is to add -DNDEBUG in the DBCSR makefile directly and not use it in the CP2K arch file

Exactly, that was my point (aka good news). Building stand-alone DBCSR and running its tests passes.

alazzaro:
Hans, DBCSR-CI only covers a fraction set of DBCSR functions and test cases (I know, we have to improve it by importing CP2K test cases). At the moment, we still rely on CP2K to fully test DBCSR...

PR #103, opened 2018-11-20
Fixed typos

hfp:
Full pass over all sources files; fixed typos in code-comments. There are very few renamed subroutine arguments and hence the interface is not affected.


src/core/dbcsr_config.F
```diff
@@ -186,7 +186,7 @@ SUBROUTINE dbcsr_set_config( &
       accdrv_min_flop_process, &
       accdrv_stack_sort, &
       accdrv_min_flop_sort, &
-      accdrv_do_inhomogenous, &
+      accdrv_do_inhomogeneous, &
```
dev-zero: this most likely breaks compilation with CP2K since it is referenced in CP2K's `src/cp_dbcsr_cp2k_link.F`

hfp: Indeed, `src/cp_dbcsr_cp2k_link.F` uses the following calls:

```fortran
CALL dbcsr_get_default_config(accdrv_do_inhomogenous=ldefault)
CALL dbcsr_set_config(accdrv_do_inhomogenous=lval)
```

dev-zero: @alazzaro we're not wrapping functions completely in `dbcsr_api.F`, just re-export. Changing the signature of a subroutine/function/structure exported in `dbcsr_api.F` = changing the API.

alazzaro: I see... this is something we have to change for the functions...

PR #108, opened 2018-11-21
Convert docs to ford

dev-zero:


alazzaro:
I've just reviewed this PR, it looks good to me. Things to do are:
1) Include the documentation creation into Makefile (and explicitly set the path to fypp and cpp in the file .ford-preprocessor.sh. Ideally, we will have a target `doc`
2) Extend the README.md to explain the new target
3) Remove all remaining doxygen targets (e.g. \brie , \return) At this point I will definitely drop doxygen syntax and move to the default FORD syntax, i.e. put the description of the functions after the function name
4) Does it make sense to extend the documentation to the test and example directories?
5) Why search is false? I tried on my version and it is very nice...
6) I have included the .cu and .h files (we can also include .py?). Unfortunately, there is a bug, see

https://github.com/Fortran-FOSS-Programmers/ford/issues/249

As far I can see, it doesn't recognize two characters for the comments (// in our case), but we can still use it (ignoring comments by using a single /).
 
6) In the future, we can decide to automatically publish the documentation somewhere for every new release...

I can work on 1)--3) and push to your branch.
Definitely, FORD deserves more attention ;) I see a lot of options that we can explore in the future...

dev-zero:
> 1. Include the documentation creation into Makefile (and explicitly set the path to fypp and cpp in the file .ford-preprocessor.sh. Ideally, we will have a target `doc`

sure, doable

> 2. Extend the README.md to explain the new target
> 3. Remove all remaining doxygen targets (e.g. \brie , \return) At this point I will definitely drop doxygen syntax and move to the default FORD syntax, i.e. put the description of the functions after the function name

The question here is: do we want to do this in this PR (possibly a lot of work) or do we want a pre-commit check which forces people to update docs on files they touch?

> 4. Does it make sense to extend the documentation to the test and example directories?

Sure, we can do that. On the other hand there aren't many functions, subroutines or custom structures in the `test/` and `example/` directories which would need structured documentation, which is why I left them out for now.

> 5. Why search is false? I tried on my version and it is very nice...

Didn't have time to test it and required a lot of memory in an older version of Fypp.

> 6. I have included the .cu and .h files (we can also include .py?). Unfortunately, there is a bug, see
> 
> [Fortran-FOSS-Programmers/ford#249](https://github.com/Fortran-FOSS-Programmers/ford/issues/249)

As long as FORD does not officially support a minimal set for other files I don't see the point in enabling that.

> As far I can see, it doesn't recognize two characters for the comments (// in our case), but we can still use it (ignoring comments by using a single /).
> 
> 1. In the future, we can decide to automatically publish the documentation somewhere for every new release...

Sure, we can extend the current Travis Deploy step to [publish to GitHub pages](https://docs.travis-ci.com/user/deployment/pages/).

> I can work on 1)--3) and push to your branch.
> Definitely, FORD deserves more attention ;) I see a lot of options that we can explore in the future...

Ok. I guess we also have to include the [custom target for CMake](https://cmake.org/cmake/help/v3.10/command/add_custom_target.html) as well.

alazzaro:
@oschuett Now that the plan is to convert to FORD for the documentation, I wonder if the CP2K API is still updated: https://apidoc.cp2k.org/index.html
I see that it is mentioning dbcsr, so is it updated?

oschuett:
Unfortunately, the apidoc stopped working some time ago. And without Sebastiano there isn't really anybody to maintain it :-/ 
So, if the transition of DBCSR to FORD succeeds then CP2K should probably follow.

oschuett:
Now that I got started thinking about how to adopt FORD for CP2K, I that have a few comments:

1. I actually really like these boxed before each subroutine. I think they give the files visual structure and invite for longer descriptions. Maybe one could keep the boxes and use the FORD syntax only for the dummy arguements?
2. I also like the `author` and `history` entries. I know this information is stored in git, but this way it's readily available when browsing the files. And I it serves the human desire to sign one's own artwork.

dev-zero:
> 1. Include the documentation creation into Makefile (and explicitly set the path to fypp and cpp in the file .ford-preprocessor.sh. Ideally, we will have a target `doc`

Target done for CMake.

> 2. Extend the README.md to explain the new target

Done (with CMake).

> 3. Remove all remaining doxygen targets (e.g. \brie , \return) At this point I will definitely drop doxygen syntax and move to the default FORD syntax, i.e. put the description of the functions after the function name

Added a pre-commit hook (which also gets executed by Travis for changed or added files). This way anyone touching a file also has to convert that file's documentation to FORD.

> 5. Why search is false? I tried on my version and it is very nice...

Done.

> 1. In the future, we can decide to automatically publish the documentation somewhere for every new release...

Included as "Push to Github-Pages after tagging". Needs testing, though.

dev-zero:
Some more thoughts for generating the docs: CMake generates implicit `<FILE>.i` targets which contain the preprocessed source. We can therefore easily remove all the pre-processing inside FORD (the `.ford-preprocessor.sh` script) and run FORD on the `.i` files inside the CMake build dir.

dev-zero:
Ok, it seems that we got some type member renames which were not reflected in the documention.

My conversion script gives me the following, where often when it says `TYPE` it actually is a problem that the documented parameter does not exist anymore (likely got renamed).

```
Parser errors occurred for 31 Doxygen headers:
src/data/dbcsr_data_types.F: Missing parameter 'refcount' in instruction 'TYPE dbcsr_scalar_type'
src/mm/dbcsr_mm_hostdrv.f90: Missing parameter 'params' in instruction 'SUBROUTINE xsmm_process_mm_batch_${nametype1}$ (stack_descr, params, &
                                                  stack_size, a_data, b_data, c_data, used_smm)'
src/mm/dbcsr_mm_hostdrv.f90: Missing parameter 'params' in instruction 'SUBROUTINE xsmm_process_mm_stack_${nametype1}$ (stack_descr, params, &
                                                  stack_size, a_data, b_data, c_data, used_smm)'
src/mm/dbcsr_mm_multrec.F: Missing parameter 'max_m' in instruction 'TYPE dbcsr_mm_multrec_type'
src/ops/dbcsr_csr_conversions.F: Missing parameter 'dbcsr_nblkcols_total,' in instruction 'TYPE csr_mapping_data'
src/ops/dbcsr_operations.F: Missing parameter 'result' in instruction 'SUBROUTINE dbcsr_dot_sd(matrix_a, matrix_b, trace)'
src/tensors/dbcsr_tensor_split.F: Missing parameter 'block_size_1' in instruction 'SUBROUTINE dbcsr_t_split_blocks_generic(tensor_in, tensor_out, ${varlist("blk_size")}$, nodata)'
src/tensors/dbcsr_tensor_test.F: Missing parameter 'mp_comm_2d' in instruction 'SUBROUTINE dbcsr_t_test_formats(ndims, mp_comm, unit_nr, verbose, &
                                   ${varlist("blk_size")}$, &
                                   ${varlist("blk_ind")}$)'
src/tensors/dbcsr_tensor_test.F: Missing parameter 'blk_ind_1' in instruction 'SUBROUTINE dbcsr_t_setup_test_tensor(tensor, mp_comm, enumerate, ${varlist("blk_ind")}$)'
src/tensors/dbcsr_tensor_types.F: Missing parameter 'map_grid' in instruction 'FUNCTION new_dbcsr_tas_blk_size_t(blk_size, map_blks, which_dim)'
src/tensors/dbcsr_tensor_types.F: Missing parameter 'nd_dist_1' in instruction 'SUBROUTINE dbcsr_t_distribution_new(dist, pgrid, map1_2d, map2_2d, ${varlist("nd_dist")}$, own_comm)'
src/tensors/dbcsr_tensor_types.F: Missing parameter 'blk_size_1' in instruction 'SUBROUTINE dbcsr_t_create_new(tensor, name, dist, map1_2d, map2_2d, data_type, &
                                 ${varlist("blk_size")}$)'
src/tensors/dbcsr_tensor_types.F: Missing parameter 'blks_local_1' in instruction 'SUBROUTINE dbcsr_t_get_info(tensor, nblks_total, &
                               nfull_total, &
                               nblks_local, &
                               nfull_local, &
                               pdims, &
                               my_ploc, &
                               ${varlist("blks_local")}$, &
                               ${varlist("proc_dist")}$, &
                               ${varlist("blk_size")}$, &
                               ${varlist("blk_offset")}$, &
                               distribution, &
                               name, &
                               data_type)'
src/tensors/dbcsr_array_list_methods.F: Missing parameter 'data_1' in instruction 'SUBROUTINE create_array_list(list, ndata, ${varlist("data")}$)'
src/tensors/dbcsr_array_list_methods.F: Missing parameter 'data_1' in instruction 'SUBROUTINE get_arrays(list, ${varlist("data")}$, i_selected)'
src/tensors/dbcsr_tensor_block.F: Missing parameter 'blk_ind_1' in instruction 'SUBROUTINE dbcsr_t_reserve_blocks_index(tensor, ${varlist("blk_ind")}$)'
src/tensors/dbcsr_tensor_io.F: Missing parameter 'proc' in instruction 'SUBROUTINE dbcsr_t_write_block(name, blk_size, blk_index, proc, unit_nr, &
                                  ${varlist("blk_values",nmin=2)}$, write_int)'
src/core/dbcsr_log_handling.F: Missing parameter 'print_keys' in instruction 'TYPE dbcsr_logger_type'
src/core/dbcsr_types.F: Missing parameter 'myproc' in instruction 'TYPE dbcsr_mp_type'
src/core/dbcsr_types.F: Missing parameter 'myproc' in instruction 'TYPE dbcsr_mp_obj'
src/core/dbcsr_types.F: Missing parameter 'row_dist,' in instruction 'TYPE dbcsr_distribution_type'
src/core/dbcsr_types.F: Missing parameter 'row_image,' in instruction 'TYPE dbcsr_imagedistribution_type'
src/core/dbcsr_types.F: Missing parameter 't' in instruction 'TYPE dbcsr_block_buffer_obj'
src/core/dbcsr_types.F: Missing parameter 'rbs,' in instruction 'TYPE dbcsr_iterator'
src/core/dbcsr_types.F: Missing parameter 'tree_s,' in instruction 'TYPE dbcsr_mutable_type'
src/core/dbcsr_types.F: Missing parameter 'm' in instruction 'TYPE dbcsr_work_type'
src/core/dbcsr_types.F: Missing parameter 'initialized' in instruction 'TYPE dbcsr_type'
src/core/dbcsr_types.F: Missing parameter 's' in instruction 'TYPE dbcsr_1d_array_obj'
src/core/dbcsr_types.F: Missing parameter 's' in instruction 'TYPE dbcsr_2d_array_obj'
src/core/dbcsr_types.F: Missing parameter 'data_windows,' in instruction 'TYPE dbcsr_1d_array_type'
src/mpi/dbcsr_mpiwrap.F: Missing parameter 'file_size' in instruction 'SUBROUTINE mp_file_get_position(fh, pos)'
```

oschuett:
It's great that we're getting apidocs again!
However, I'd very much prefer to put the comments **before** the code which they are documenting. Fortunately, Ford now supports this out of the box. It would make this change much less invasive as we could keep our nice boxes.

dev-zero:
now I also preserve/conserve the `\note` and `\par` as well as integrate the `\return` tags and I'm down to 20 headers to be converted manually

oschuett:
I totally agree that Ford is a big improvement.

I'm only proposing that we put the doc strings **before** the code. This is what we're all used to and it's also the de-facto standard in Fortran and C. Again, Ford supports this out of the box via [predocmark](https://github.com/Fortran-FOSS-Programmers/ford/wiki/Project-File-Options#predocmark).

So, then one would write code like this:

```
!> Performs a multiplication of two dbcsr_type matrices,
!> as  C := alpha * op( A ) * op( B ) + beta * C.
SUBROUTINE dbcsr_multiply_generic(transa, transb, ...)
      !> specifies the form of op( A ) to be used in the matrix multiplication ...
      CHARACTER(LEN=1), INTENT(IN)                       :: transa, transb
```

The vertical bars for the boxes could later be inserted automatically by a script:

```
!***********************************************************************************
!> Performs a multiplication of two dbcsr_type matrices,
!> as  C := alpha * op( A ) * op( B ) + beta * C.
!***********************************************************************************
SUBROUTINE dbcsr_multiply_generic(transa, transb, ...)
      !> specifies the form of op( A ) to be used in the matrix multiplication ...
      CHARACTER(LEN=1), INTENT(IN)                       :: transa, transb
```

dev-zero:
> I totally agree that Ford is a big improvement.
> 
> I'm only proposing that we put the doc strings **before** the code. This is what we're all used to and it's also the de-facto standard in Fortran and C. Again, Ford supports this out of the box via [predocmark](https://github.com/Fortran-FOSS-Programmers/ford/wiki/Project-File-Options#predocmark).

And in Python the de-facto standard is to put the documentation for the function after the declaration.
From what I can see is that pre-doc hurts in multiple ways:

* refactoring: always having to make sure that you're actually getting the whole thing
* folding: while we can expect folding in editors to work for a function body, documentation is not necessarily folded as well. And besides, I don't want to look at documentation of a function I'm not interested in.
* navigation: when I jump to implementation I can easily setup my editor to align the line to top, but when documentation is on top no matter how much context I allow, chances are that it's not enough and I have to scroll up again
* title/abstract/content: that's how people are used to read stuff. If the function name is not sufficient to serve as a title you might want to rename it, abstract is the short documentation following the title and the declaration of dummy arguments, content is the implementation

oschuett:
It's unfortunate that you decided not see this discussion through to the end.

The code you changed is written in Fortran and it should therefore adhere to Fortran conventions. While you might find them inconvenient, these conventions make the project more accessible to others.

alazzaro:
My two cents.
I would prefer to keep as it is (default FORD) and move forward. Then, we can see how to improve. Documentation is a big lack in DBCSR, we have to start somewhere...

PR #112, opened 2018-11-30
Makefile change s.t. mod files are not recompiled unnecessarily

pseewald:
The proposed changes make sure that we do not recompile all modules using a.F if a.F has changed, recompilation only happens if a.mod has changed (gfortran does not update a .mod file if it does not change).

For more information: http://lagrange.mechse.illinois.edu/f90_mod_deps/

CP2K Makefile solves this in the same way.

Needed for cp2k/cp2k#86, in order to to recompile cp2k only if dbcsr_api.mod has changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -422,10 +422,12 @@ vpath %.cpp   $(ALL_SRC_DIRS)
 
 FYPPFLAGS ?= -n
 
-%.o %.mod: %.F
-	@rm -f $*.mod
+%.o: %.F
 	$(FYPPEXE) $(FYPPFLAGS) $< $*.F90
 	$(FC) -c $(FCFLAGS) -D__SHORT_FILE__="\"$(notdir $<)\"" -I'$(dir $<)' -I'$(SRCDIR)' $*.F90 $(FCLOGPIPE)
+
+%.mod: %.o
+	@true
 
 %.o: %.c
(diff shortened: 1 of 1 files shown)


alazzaro:
Well, we had a long discussion about what will be the right way to handle that case. I dropped the CP2K solution just for safety (the library is small enough and it can get recompiled). Indeed, you can delete a mod file and it will not be recreated. I'm not sure why this change should be related with CP2K, DBCSR is independent...

pseewald:
> I'm not sure why this change should be related with CP2K, DBCSR is independent...

What we want is that CP2K gets rebuilt only if `dbcsr_api.mod` has been updated, to avoid full recompilation of CP2K for every DBCSR change. But due to the current DBCSR build dependencies, `dbcsr_api.mod` is almost always updated when you change something in DBCSR.

> Indeed, you can delete a mod file and it will not be recreated.

Why should this case occur, and even if it does, it should not cause any real problems (build will fail and you need to `make clean` but you will never end up with wrongly compiled executables). However I do consider slow build times a problem (1 min vs. 1 s for `make` after `touch src/base/dbcsr_kinds.F`).

alazzaro:
Ah, I see, so you are going to change the dependency in CP2K, instead of the library... OK, then go ahead and merge this PR

PR #114, opened 2018-12-02
Removed padded data

hfp:
Removed padded data (over-allocation) for MATMUL, BLAS, and LIBXSMM code path. The SMM-layer is supposed to never cause OOBs (unless padding is clearly specified/determined). Previously, the +10-pad was cumbersome.


hfp:
The code coverage report makes not much sense: -46% for 3 lines of changed code does not look correct.

PR #127, opened 2018-12-18
Makefile: fix compatibility issue with cp2k toolchain

shoshijak:
- The Makefile variable `NVFLAGS` can already contain the correct gpu
architecture option depending on the variables included into the
makefile. If the option is set twice, compilation errors arise.
Therefore check whether the option is already set before setting
it in DBCSR Makefile.
- This fixes the following issue in cp2k repo: https://github.com/cp2k/cp2k/issues/128

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,10 @@ else
 endif
 
 ifneq ($(ARCH_NUMBER),)
+#if "-arch" has not yet been set in NVFLAGS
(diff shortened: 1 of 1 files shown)


alazzaro:
@shoshijak There is a bug in DBCSR that I have to fix. Then I will update DBCSR in CP2K

Issue #129, opened 2018-12-18
Verify numerical primitives


dev-zero:
Out of interest I was looking into the code of CP2K to check how the L2-norm of vectors is calculated.
I got the 3 equivalent forms `sqrt(x(1)*x(1) + ...)`, `sqrt(x(1)**2 + ...` and `sqrt(dot_product(x,x))` but not many usages of the F2008 built-in `norm2`.
Being curious I looked at the machine code a typical compiler generates (see https://godbolt.org/z/c5H1Ih for a comparison of the different variants) and noticed that the built-in `norm2` is much more elaborate.

Not being an expert on assembler my guess is that this is the implementation of the standard-optional `It is recommended that the processor compute the result without undue overflow or underflow.` (see 13.7.123 of the Fortran standard).

A quick search in the DBCSR source code led me to the following (and it is likely not the only place):
https://github.com/cp2k/dbcsr/blob/d03f22f5ad4b497a13e4017f8d0ba5c77801c319/src/ops/dbcsr_operations.F#L2125-L2130
which seems to me a rather careless way on how to calculate the norm, given that compilers take more care when calculating the norm via an intrinsic and after a quick look into https://www.springer.com/us/book/9780817647056

Am I missing something here?
@hfp maybe you could share some insight?

hfp:
@dev-zero Fortran built-ins are preferable. I can only find two reason for not using the intrinsics:

* Insufficient functionality (e.g., Kahan compensation needed, or low-accuracy approximation sufficient)
* Limited portability (e.g., not available in F2008 standard)

For the latter point, we can use `__F2008` preprocessor definition (at least somewhat introduced in CP2K's sources). However, this may spread compile-time toggles all over the code if the code-snippet is not wrapped into a (utility-)function. Btw, there is more legacy code for functions such as `ERF` (at least CP2K implements this among some others) which are meanwhile covered by F-intrinsics. Regarding code style (as pointed out in your example code above), even array expressions may be preferred sometimes over element-wise loop enumeration.

For the `norm2` alike, we did some pass over CP2K's source some (longer) time ago and got rid of things like `IF (norm2(x) .LE. norm2(y)) THEN` (an expression where `sqrt` is superfluous).

dev-zero:
@hfp thank you very much for the details. So it is indeed something which should be done at some point.
[...]

PR #136, opened 2019-01-16
switch to F2008

dev-zero:
based on the discussion in #129, https://github.com/cp2k/cp2k/issues/142 and the fact that PR https://github.com/cp2k/cp2k/pull/158 was merged for CP2K this PR adds a check for F2008's `norm2` support in the compiler when using CMake.


alazzaro:
You forgot to update the Makefile.inc... Let me do it ;)

alazzaro:
@dev-zero Concerning this PR, should we update the intrinsics in DBCSR?

PR #137, opened 2019-01-21
Feature/libcusmm optimal kernel parameters prediction

shoshijak:
Currently, libcusmm’s matrix-matrix multiplication kernels are autotuned in order to find the parameters with which they perform best. This means libcusmm can only handle the multiplication of (m, n, k)-triplets that have been autotuned, while other multiplications are handed back to the CPU. Unfortunately, exhaustive autotuning of the entire space of all (m, n, k)-triplets is prohibitively expensive. 

This PR introduces predictive modeling for libcusmm’s optimal matrix-matrix multiplication parameters. A decision tree model is built from the data obtained by autotuning certain (m, n, k)-triplets and is used to predict optimal parameters for unseen (m, n, k)-triplets.

For a detailed description of changes, see below. 

----------------------------

Developed as a part of the PASC proposal "Sparse Tensor Linear Algebra Library"

----------------------------

**Build system**
- Complete the output of target `toolversions` (commit 1b5a22ba)
- Add target `toolflags` (commit 1b5a22ba)

**`libcusmm` kernel code**
- Improve CUDA error handling (commit 6ac7436)

**`libcusmm` autotuning**
- Rename tuning scripts so that they’re prefixed with `tune_` (commit 9660c79a)

**`libcusmm` predictive modeling**
- Add Python scripts for the entire training procedure (prefixed with `predict_`), from gathering data to generating parameter files

**`libcusmm` parameters** 
- Add predicted optimal parameters for the P100 (commit a8430374)

**`libcusmm` notebooks**
- Add two Jupyter notebooks to explore autotuned parameter sets and collected autotuning data (commit 7363d0a9, e396ede4 and 82e00f05)

**`libcusmm` tests**
- More reliable DBCSR performance tests (commit e91e176)
- Generate `libcusmm_timer_multiply.cu` and `libcusmm_unittest_multiply.cu` at DBCSR’s compile time. Run kernel multiplications sampled from both autotuned and predicted parameter sets (commit 3de1c75)
- libcusmm timer has two modes: autotuned or predicted 

**`libcusmm` python scripts in general** 
- Reformat using [Black](https://github.com/ambv/black) and made flake8 compatible (commit ce0f6700, 5728c9af and 24b66aad)
- Improve comments

**`libcusmm` documentation** 
- Add a libcusmm README & kernels README (commit e2c18022, 18b660ab, 7ede6ed2 and 840b0e01)
- Documentation on predictive modelling goals and procedure (`predict.md`)
- Complete comments throughout the code
- Add Python package requirements (commit 49d6ca7d)


shoshijak:
The CI fails because two scripts belonging to the autotuning and predicting workflows:
- `src/acc/libsmm_acc/libcusmm/tune_setup.py` and 
- `src/acc/libsmm_acc/libcusmm/predict_genpars.py`

contain syntax that isn’t python2 compatible. 

I understand the wish to keep non-optional python scripts compatible with python2, but could we relax this requirement for scripts that will only be used by more advanced users, e.g. the scripts belonging to the autotuning and predicting workflows? 

This should be relatively easy to do since those scripts are all in the `src/acc/libsmm_acc/libcusmm` directory and are all prefixed with `tune_` and `predict_` respectively. I’m not very familiar with travis and pre-commit hooks however, so I’m not sure what the best way to do this would be.

dev-zero:
Cool stuff! Takes some time to review, though ;-)

> The CI fails because two scripts belonging to the autotuning and predicting workflows:
> 
> * `src/acc/libsmm_acc/libcusmm/tune_setup.py` and
> * `src/acc/libsmm_acc/libcusmm/predict_genpars.py`
> 
> contain syntax that isn’t python2 compatible.
> 
> I understand the wish to keep non-optional python scripts compatible with python2, but could we relax this requirement for scripts that will only be used by more advanced users, e.g. the scripts belonging to the autotuning and predicting workflows?

With Python 2.7 being supported for less than a year now I agree.
The required changes have to be done in `.pre-commit-config.yaml`:

```yaml
- repo: https://github.com/pre-commit/pre-commit-hooks
  rev: v2.0.0
  hooks:
  - id: flake8
  - id: check-ast
    stages: [manual]
    language_version: "python2"
  - id: check-ast
    stages: [manual]
    language_version: "python3"
```

should become something like

```yaml
- repo: https://github.com/pre-commit/pre-commit-hooks
  rev: v2.0.0
  hooks:
  - id: flake8
  - id: check-ast
    stages: [manual]
    language_version: "python2"
    exclude: '^src/acc/libsmm_acc/libcusmm/(tune|predict)_'
  - id: check-ast
    stages: [manual]
    language_version: "python3"
```

You can test it locally by running:
```bash
pip install pre-commit  # on non-Ubuntu you may have to add --user as well
# you may have to add ~/.local/bin to your $PATH before the following commands work
pre-commit install
pre-commit run --hook-stage manual --files src/acc/libsmm_acc/libcusmm/*```

**Comment by dev-zero** [comment]
Btw, when going py3-only, I suggest to take a look at [Python f-strings](https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python) since you're doing quiet some string-formatting

**Comment by shoshijak** [comment]
> Cool stuff! Takes some time to review, though ;-)

of course, no rush.

> With Python 2.7 being supported for less than a year now I agree.
The required changes have to be done in .pre-commit-config.yaml:

Thanks a lot for the indications on the pre-commit hooks, I just pushed a commit fixing this issue.

> Btw, when going py3-only, I suggest to take a look at Python f-strings since you're doing quiet some string-formatting

I just took a look, f-strings are really cool, I didn't know about them. Thanks!

File: src/acc/libsmm_acc/libcusmm/predict_helpers.py [hunk 1]
```diff
@@ -0,0 +1,223 @@
...
+
+
+def safe_pickle_load(file_path):
+    max_bytes = 2**31 - 1
+    bytes_in = bytearray(0)
+    input_size = os.path.getsize(file_path)
+    with open(file_path, "rb") as f:
+        for _ in range(0, input_size, max_bytes):
+            bytes_in += f.read(max_bytes)
+    return pickle.loads(bytes_in)
```
dev-zero: Can you elaborate on what kind of problem you're trying to solve here and why `pickle.dump`/`pickle.load` does not work?

The reading part looks rather inefficient and could probably benefit from `mmap` if `pickle.load` can't cope with large amount of data, without having to manually build a `bytearray` with the complete data:

```
import mmap
from contextlib import contextmanager

@contextmanager
def mmap_open(name, mode='r'):
    access = mmap.ACCESS_READ if mode == 'r' else mmap.ACCESS_WRITE
    with open(name, mode + 'b') as fhandle:
        fmapped = mmap.mmap(fhandle.fileno(), 0, access=access)
        yield fmapped
        fmapped.close()

with mmap_open(fpath) as fmapped:
    return pickle.loads(fmapped)
```

src/acc/libsmm_acc/libcusmm/predict_helpers.py
```diff
@@ -0,0 +1,223 @@
...
+# I/O helpers
+def safe_pickle(data, file):
+    """
+    Pickle big files safely by processing them in chunks
+    :param data: data to be pickled
+    :param file: file to pickle it into
+    """
+    max_bytes = 2**31 - 1  # Maximum number of bytes to pickle in one chunk
+    pickle_out = pickle.dumps(data)
+    n_bytes = sys.getsizeof(pickle_out)
```
dev-zero: are you sure that `len(pickle_out)` doesn't work? You have a `bytes` object and you're writing the contained `bytes`, while `sys.getsizeof` gives the size of the python object including the metadata. Compare:

```python
import pickle
import sys

d = {'foo': 123}
print(len(pickle.dumps(d)))
print(sys.getsizeof(pickle.dumps(d)))
```

src/acc/libsmm_acc/libcusmm/predict_train.py
```diff
@@ -0,0 +1,1165 @@
...
+
+# ===============================================================================
+def main():
+    """
+    This script is part of the workflow for predictive modelling of optimal libcusmm parameters.
+    For more details, see predict.md
+
+    """
+
+    parser = OptionParser()
```
dev-zero: while still working, [`OptionParser`](https://docs.python.org/2/library/optparse.html) was deprecated in Python 2.7 and got replaced with [argparse](https://docs.python.org/3/library/argparse.html).

src/acc/libsmm_acc/libcusmm/predict_helpers.py
shoshijak: `pickle.dump`/`load` with Python3 on OSX has trouble handling very large data (see https://bugs.python.org/issue24658). These two functions are meant as a safety wrapper around `pickle.dump`/`load` to avoid this problem.

shoshijak: I'm not too concerned about the performance of the load since this function is called only in the beginning of the training (`predict_train.py`) (optionally, for evaluating a pre-trained model) and the parameter generation (`predict_genpars`), both long procedures which would not be impacted much by reducing the time of loading they do once. 

I'd prefer sticking with a simpler and more readable package.

src/acc/libsmm_acc/libcusmm/predict_helpers.py
shoshijak: Good catch! 
I believe my error doesn't necessarily introduce buggy behaviour though, since `sys.getsizeof` returns a value larger than `len` and `f.read` in `safe_pickle_load` only reads until EOF.
Thanks for that!

src/acc/libsmm_acc/libcusmm/predict_train.py
shoshijak: Thanks!
`OptionParser` is not used just here though, but throughout python scripts in DBCSR.
I'll change that ...

dev-zero:
Looks good to me :+1:

src/acc/libsmm_acc/libcusmm/predict_collect.py
```diff
@@ -0,0 +1,314 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
shoshijak: thanks!

alazzaro:
@shoshijak Thanks for your massive commit! Sorry, I didn't have time to review it in the last days, so I did a fast glimpse on it and it looks very good (and interesting)! 
We still don't have a way to test on GPU with CI (@dev-zero any update on that?), but I would propose to move to v1.1 and push the new release to CP2K (we can include #140 ).
Question: is cmake supporting GPU build?

dev-zero:
:-( no update yet, I'm sorry. Yes, CMake is able to build for GPU.

shoshijak:
@alazzaro thanks a lot :) 
Note PR #140 is not quite ready yet, I still want to fix some potential issues in the build system pointed out by @dev-zero .

alazzaro:
@shoshijak BTW, no strictly related, I think your work is really a good candidate for a paper. Are you planning to submit a paper somewhere?

shoshijak:
@alazzaro I don't know yet

PR #140, opened 2019-02-01
Fix/perf issue

shoshijak:
### Motivation
Investigate libcusmm's performance following a [performance issue in CP2K](https://github.com/cp2k/cp2k/issues/155).

----------------------
### PR changes
- Add a performance test loosely corresponding to H2O simulations in CP2K (commit
 94326fb, file `tests/inputs/test_H2O.perf`) 

- Fix a performance issue in libcusmm (see commit 7b808c09 for description of the fix)

----------------------
### JIT-ing kernels 

is performed only once per node for each (m,n,k)-multiplication triplet or each (m,n)-transpose pair which occurs in the program. This can be verified by running a test with [nvprof](https://devblogs.nvidia.com/cuda-pro-tip-nvprof-your-handy-universal-gpu-profiler/) as in the following example:

Sample output of `nvprof` for DBCSR's performance driver with `tests/inputs/test_H2O.perf` on 1 node, 12 threads: 

```
            Type  Time(%)      Time     Calls       Avg       Min       Max  Name
 GPU activities:   50.13%  26.574ms        73  364.02us  14.368us  469.85us  void cusmm_dnt_medium<int=23, int=23, int=23, int=3, int=2, int=96, int=4, int=4>(int const *, int, double const *, double const *, double*)
                   32.09%  17.009ms        84  202.49us  1.0230us  2.5687ms  [CUDA memcpy HtoD]
                   16.99%  9.0034ms        37  243.33us  3.8400us  254.11us  [CUDA memcpy DtoH]
                    0.73%  384.73us         3  128.24us  127.39us  129.02us  void transpose_d<int=23, int=23>(int*, double*)
                    0.07%  35.424us        37     957ns     768ns  2.4320us  [CUDA memset]
      API calls:   19.70%  273.86ms        36  7.6073ms  3.0762ms  9.3789ms  cudaStreamSynchronize
...
                    3.08%  42.758ms         2  21.379ms  17.946ms  24.812ms  cuModuleLoadDataEx
```

- The [NVRTC](https://docs.nvidia.com/cuda/nvrtc/index.html) API function `cuModuleLoadDataEx`, which is the longest function to execute as part of the JITting, (see `src/acc/libsmm_acc/libcusmm/libcusmm.cpp`) is called only twice (once for (23,23,23)-multiplication and once for (23,23)-transpose), though transposition (`transpose_d`) is called `3` times and multiplication (`cusmm_dnt_medium`) is called `73` times.
- JIT-ing kernels rather than generating them and compiling them at DBCSR's compile time (as was the case before PR #15) is therefore a one-off cost at runtime.
- The overhead of JIT-ing a single kernel is about `300-400 ms`, a negligible overhead for long runs. 

----------------------

### Performance numbers in DBCSR 
**Running performance driver with `tests/inputs/test_H2O.perf` on 1 node, 12 threads (cscs-daint-gpu)**

| commit | mean [s] | minmin [s] | maxmax [s] |
| ------------- |-------------:|-------------:|-------------:|
| `7b808c09` (this PR) | 0.311 | 0.035 | 0.860 |
| `78a915c2` (JIT-PR #15) | 1.45 | 0.036 | 4.29 |
| `5844932` (before JIT-PR #15) | 0.133 | 0.036 | 0.325 |

(each cell value is averaged over 3 runs of the performance driver) 

- The "minmin" values are very similar in all three DBCSR versions, in alignment with the fact that kernel-JITing incurs a one-off cost.
- Comparing the maxmax values, we see that the JIT overhead difference between `7b808c09` and `5844932` is of about `530 ms` for 2 kernels ((23,23,23)-multiplication and (23,23)-transpose), i.e. about `270 ms` per kernel in this case. These numbers are to be taken with a grain of salt though, since the first kernels launched on the GPU come with some GPU-startup overhead that is difficult to estimate. 
- This overhead is reduced between `7b808c09` and `78a915c2` as explained in commit 7b808c09.

----------------------

### Performance numbers in CP2K 
**Running CP2K with `tests/QS/benchmark_DM_LS/H2O-dft-ls.inp` with `NREP=1` on 1 node, 1 rank, 12 threads (cscs-daint-gpu)**

| CP2K commit | DBCSR commit | runtime [s] | 
| ------------- | ------------- |-------------:|
| `53314e2` (DBCSR separate from CP2K) | `7b808c09` (this PR) | 5.38 |
| `53314e2` (DBCSR separate from CP2K) | `78a915c2` (JIT-PR #15) | 7.45 |
| `e9292bad5` (DBCSR not yet separated from CP2K) | `5844932` (no JIT)| 4.82 | 

(values averaged over 6 runs)

**Running CP2K with `tests/QS/benchmark_DM_LS/H2O-dft-ls.inp` with `NREP=3` on 8 nodes, 2 ranks/node, 12 threads/rank (same config as `CRAY_XC50_gpu (Performance)` on the [dashboard](https://dashboard.cp2k.org/), cscs-daint-gpu)**

| CP2K commit | DBCSR commit | runtime [s] | 
| ------------- | ------------- |-------------:|
| `53314e2` (DBCSR separate from CP2K) | `7b808c09` (this PR) | 112.77 |
| `53314e2` (DBCSR separate from CP2K) | `78a915c2` (JIT-PR #15) | 119.16 |
| `e9292bad5` (DBCSR not yet separated from CP2K) | `5844932` (no JIT)| 109.27 | 

(values averaged over 3 runs)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -486,7 +486,7 @@ FYPPFLAGS ?= -n
 	$(CXX) -c $(CXXFLAGS) $<
 
 libcusmm.o: libcusmm.cpp parameters.h cusmm_kernels.h
-	$(CXX) -c $(CXXFLAGS) -fopenmp -DARCH_NUMBER=$(ARCH_NUMBER) $<
+	$(CXX) -c $(CXXFLAGS) $(CXXOMPFLAGS) -DARCH_NUMBER=$(ARCH_NUMBER) $<
 
 %.o: %.cu
 	$(NVCC) -c $(NVFLAGS) -I'$(SRCDIR)' $<
--- a/Makefile
+++ b/Makefile
@@ -486,7 +486,7 @@ FYPPFLAGS ?= -n
 	$(CXX) -c $(CXXFLAGS) $<
 
 libcusmm.o: libcusmm.cpp parameters.h cusmm_kernels.h
-	$(CXX) -c $(CXXFLAGS) -DARCH_NUMBER=$(ARCH_NUMBER) $<
+	$(CXX) -c $(CXXFLAGS) -fopenmp -DARCH_NUMBER=$(ARCH_NUMBER) $<
 
 %.o: %.cu
 	$(NVCC) -c $(NVFLAGS) -I'$(SRCDIR)' $<
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,14 @@ endif ()
 
 get_filename_component(Fortran_COMPILER_NAME ${CMAKE_Fortran_COMPILER} NAME)
 
+# if the fortran compiler command is a CRAY wrapper, find out what the underlying compiler is
+if (Fortran_COMPILER_NAME MATCHES "ftn")
+  execute_process(COMMAND ftn -craype-verbose OUTPUT_VARIABLE ftn_verbose_OUT ERROR_QUIET)
+  separate_arguments(ftn_verbose_OUT NATIVE_COMMAND "${ftn_verbose_OUT}")
+  list(GET ftn_verbose_OUT 0 Fortran_COMPILER_NAME)
+endif ()
+
+# Depending on the fortran compiler, set the appropriate compiler flags
 if (Fortran_COMPILER_NAME MATCHES "gfortran.*")
   set(CMAKE_CXX_FLAGS              "-std=c++11")
   set(CMAKE_CXX_FLAGS_RELEASE      "-O3 -funroll-loops")
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -34,11 +34,12 @@
 #
 # For convenience, some variables can be set during compilation,
 # e.g. make `VARIABLE=value` (multiple variables are possible):
-# 1) MPI=0    : disable MPI compilation
-# 2) GNU=0    : disable GNU compiler compilation and enable Intel compiler compilation
-# 3) CHECKS=1 : enable GNU compiler checks and DBCSR asserts
-# 4) CINT=1   : generate the C interface
-# 5) GPU=1    : enable GPU support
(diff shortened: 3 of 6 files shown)


Makefile
```diff
@@ -439,7 +439,7 @@ FYPPFLAGS ?= -n
 	$(CXX) -c $(CXXFLAGS) $<
 
 libcusmm.o: libcusmm.cpp parameters.h cusmm_kernels.h
-	$(CXX) -c $(CXXFLAGS) -DARCH_NUMBER=$(ARCH_NUMBER) $<
+	$(CXX) -c $(CXXFLAGS) -fopenmp -DARCH_NUMBER=$(ARCH_NUMBER) $<
```
dev-zero: Please note that this most likely breaks compilation with PGI compilers (which we do not support yet since they are unable to compile DBCSR, but their devs are working on it).
Plus we might also have to add a check to the `CMakeLists.txt` that OMP is required when building with CUDA support.

shoshijak: Right, good point ... thanks for that! 
- Keeping the option of building **with** CUDA support but **without** OMP support is probably desirable, I'll make this possible using appropriate `#ifdef`s in `libcusmm.cpp`. 
- To fix this foreseeable PGI issue, we should probably let the OMP flags be defined appropriately in the `Makefile.inc` rather than hard-coded in the `Makefile`. I'll fix this.

shoshijak:
Please note this PR is **not ready** to be merged.

A race condition appears when two threads try to JIT two different `(m, n, k)`s simultaneously (observed, for example, when running `dbcsr_unittest3` with #ranks >=2, #threads >= 12). I'm still working on this. 

This race condition is introduced by this PR, there is no race condition in the current `dbcsr/develop`.

PR #132, opened 2019-01-08
tests: add dbcsr_unittest3 to test libcusmm-backend

shoshijak:
Add a matrix-matrix multiplication test which features block-sizes commonly used in libcusmm.
Closes issue #131 

Sample of test output: 
```
 -------------------------------------------------------------------------------
 -                                                                             -
 -                                DBCSR STATISTICS                             -
 -                                                                             -
 -------------------------------------------------------------------------------
 COUNTER                                    TOTAL       BLAS       SMM       ACC
 flops    13 x     4 x     4             59938944      74.0%      0.0%     26.0%
 flops     4 x    13 x     4             60162752      74.3%      0.0%     25.7%
 flops     4 x     4 x    13             60968960      73.9%      0.0%     26.1%
 flops    25 x     4 x     4            114982400      73.9%      0.0%     26.1%
 flops     4 x    25 x     4            115664800      74.2%      0.0%     25.8%
 flops     4 x     4 x    25            117375200      73.2%      0.0%     26.8%
 flops    14 x    14 x    14            162516144      73.8%      0.0%     26.2%
 flops    13 x     4 x    13            195892632      74.3%      0.0%     25.7%
 flops     4 x    13 x    13            197118896      73.9%      0.0%     26.1%
 flops    13 x    13 x     4            198487120      73.9%      0.0%     26.1%
 flops    14 x    29 x    14            324215360      74.2%      0.0%     25.8%
 flops    14 x    14 x    29            326750424      73.8%      0.0%     26.2%
 flops    29 x    14 x    14            328535200      74.2%      0.0%     25.8%
 flops     4 x     4 x     4            338922368      74.0%      0.0%     26.0%
 flops    14 x    32 x    14            365833216      74.3%      0.0%     25.7%
 flops    32 x    14 x    14            367213056      74.4%      0.0%     25.6%
 flops    14 x    14 x    32            370800640      73.5%      0.0%     26.5%
 flops    13 x    25 x     4            374740600      74.2%      0.0%     25.8%
 flops    25 x     4 x    13            376090000      74.4%      0.0%     25.6%
 flops     4 x    13 x    25            376633400      73.6%      0.0%     26.4%
 flops    13 x     4 x    25            377325000      74.0%      0.0%     26.0%
 flops    25 x    13 x     4            380484000      73.8%      0.0%     26.2%
 flops     4 x    25 x    13            381056000      73.9%      0.0%     26.1%
 flops     4 x     5 x     4            399021760      74.2%      0.0%     25.8%
[...]

PR #143, opened 2019-02-06
Fix build tests

shoshijak:
Fixes to the building of tests

**cmake** 
- add `dbcsr_unittest3` to the DBCSR tests (really should have been done in PR #132 , where I updated the `Makefile` but forgot about `cmake` ... )
- add the generation of `libcusmm_unittest_multiply.cu` and `libcusmm_timer_multiply.cu` (really should have been done in PR #137 , but I again forgot about `cmake` ...)

**Makefile**
- add forgotten dependency on the template

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -34,6 +34,7 @@ endforeach ()
 set(DBCSR_TESTS
   dbcsr_unittest1
   dbcsr_unittest2
+  dbcsr_unittest3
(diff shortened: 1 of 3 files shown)


shoshijak:
Checks fail because this PR adds `dbcsr_unittest3` to the list of unit tests in cmake, and since this test is significantly longer that the other `dbcsr_unittest`s, travis runs out of time.

PR #148, opened 2019-03-06
libcusmm-autotune: small fixes

shoshijak:
Fixes minor issues, aesthetic details and documentation in scripts related to autotuning and collecting data.

### DIFF
--- a/.pre-commit/check_header.py
+++ b/.pre-commit/check_header.py
@@ -44,10 +44,8 @@ def check_header(header_dir, files, verbose=False):
     header_len = {}
 
     for headertype in TYPES:
-        with open(path.join(header_dir, headertype), 'r') as fhandle:
+        with open(path.join(header_dir, headertype), 'rb') as fhandle:
             header_content = fhandle.read()
-            if not hasattr(header_content, "decode"):
-                header_content = bytes(header_content, 'utf-8')
             header_re[headertype] = re.compile(re.escape(header_content))
             header_len[headertype] = len(header_content)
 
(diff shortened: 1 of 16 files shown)


shoshijak:
@dev-zero 

> we could remove yapf completely for python2

I tried doing that. When I do so, travis fails (presumably because yapf requires py3.6.4, I cannot find any other explanation), and when I specify `py3.6.4` in `.pre-commit-config.yaml`, it fails because `py3.6.4` is not available.

shoshijak:
I just confirmed locally on my laptop, the error that occurs when travis runs yapf (namely: `lib2to3.pgen2.parse.ParseError: bad input: type=12, value=',', context=('', (95, 32))`) is due to travis running yapf with python3.5 which yapf requires python3.6.4+

shoshijak:
@dev-zero I fixed the travis-yapf interaction as well as the compatibility of `check_headers.py` with python3.6. It should be all good now.

.pre-commit/check_header.py
```diff
@@ -46,6 +46,8 @@ def check_header(header_dir, files, verbose=False):
     for headertype in TYPES:
         with open(path.join(header_dir, headertype), 'r') as fhandle:
             header_content = fhandle.read()
+            if not hasattr(header_content, "decode"):
+                header_content = bytes(header_content, 'utf-8')
```
dev-zero: ok, this was obviously my bad: I've implemented a regex against bytes with a str, sorry :(

Wouldn't it work to open the file as a bytes stream right from the beginning? So:
```python
with open(path.join(header_dir, headertype), 'rb') as fhandle:
```

shoshijak: No worries, it was an easy catch.
You're right, what you suggest is a better solution, I just incorporated it.

PR #152, opened 2019-04-09
libcusmm/tune: add autotuning procedure from cp2k.org

shoshijak:
Move the instructions for autotuning libcusmm cuda kernels from cp2k.org to github repo for a better separation of cp2k and dbcsr's documentation.


alazzaro:
Great, thanks!

PR #154, opened 2019-04-10
Bug/misc fixes

dev-zero:


.pre-commit-config.yaml
```diff
@@ -4,23 +4,20 @@ default_language_version:
...
+  - id: black
     language_version: "python3"
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v2.0.0
   hooks:
-  - id: flake8
-    language_version: "python2"
-    exclude: '^src/acc/libsmm_acc/libcusmm/(tune|predict)_'
-  - id: flake8
-    language_version: "python3"
```
shoshijak: Why remove `flake8` for py2 but keep check-ast for py2?

.pre-commit-config.yaml
```diff
@@ -4,23 +4,20 @@ default_language_version:
 exclude: '^tools/(prettify/fprettify|build_utils/fypp)'
 fail_fast: false
 repos:
-- repo: https://github.com/pre-commit/mirrors-yapf
-  rev: v0.24.0
+- repo: https://github.com/ambv/black
+  rev: stable
   hooks:
-  - id: yapf
+  - id: black
```
shoshijak: I'm very happy about switching from yapf to black!

.pre-commit-config.yaml
dev-zero: `flake8` is only about style, while `check-ast` validates that it can still be parsed with Python 2 (neither of them ensure that it still runs on Python 2 and if it does that it gives the same result as with Python 3), which is why I would argue that the `flake8` on Python 2 check doesn't add anything?

.pre-commit-config.yaml
dev-zero: cool :)

.pre-commit-config.yaml
shoshijak: That makes sense. 
In the back of my mind, I was hoping that you wanted to drop support for python2 all together, and that you had just forgotten to remove `check-ast` for python2.

dev-zero: Want, yes, but there are (according to some survey) still CP2K users relying on Python 2.
So, 1.1.2020 it is for now ;-)

PR #155, opened 2019-04-11
cmake: add OpenMP compiler flags if OpenMP found

shoshijak:
This PR fixes the `cmake`-compilation of DBCSR with when `OpenMP` is used by setting the appropriate flags.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,9 @@ project(dbcsr)
 
 include(CMakeDependentOption)
 
+############################################################################
+# OPTIONS DEFINITION:
+
 option(USE_MPI "Build with MPI support" ON)
 option(USE_OPENMP "Build with OpenMP support" ON)
 cmake_dependent_option(WITH_C_API "Build the C API (ISO_C_BINDINGS)" ON "USE_MPI" OFF) # the ISO_C_BINDINGS require MPI unconditionally
@@ -43,6 +46,7 @@ if (USE_CUDA)
 endif ()
 
 
+############################################################################
 # PACKAGE DISCOVERY:
 
 find_package(BLAS REQUIRED)
@@ -84,6 +88,7 @@ if (NOT CMAKE_BUILD_TYPE)
     FORCE)
 endif ()
 
+############################################################################
 # COMPILER CONFIGURATION:
 
 get_filename_component(Fortran_COMPILER_NAME ${CMAKE_Fortran_COMPILER} NAME)
@@ -141,6 +146,15 @@ Please open an issue at https://github.com/cp2k/dbcsr/issues with the reported c
   set(CMAKE_Fortran_FLAGS_DEBUG    "-O0 -g")
 endif ()
 
+# Add OpenMP flags if OpenMP has been found
+if(OPENMP_FOUND)
+  set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${OpenMP_Fortran_FLAGS}")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
+endif()
+
+############################################################################
 file(STRINGS VERSION VERSION_INFO)
 foreach(line ${VERSION_INFO})
   if (${line} MATCHES "^([^#].*)=[ \t]*(.*)$")


dev-zero:
hmm, what was the error you were seeing?

When I build `dbcsr` I get the following:

```
make -j VERBOSE=1
[...]
cd /users/tiziano/work/dbcsr/dbcsr/build/src && /usr/bin/gfortran -DNDEBUG -D__MPI_VERSION=3 -D__STATM_TOTAL -D__parallel -D__SHORT_FILE__=\"dbcsr/data/dbcsr_data_operations.F\" -I/users/tiziano/work/dbcsr/dbcsr/src/base -I/users/tiziano/work/dbcsr/dbcsr/src -I/users/tiziano/work/dbcsr/dbcsr/build/src -I/usr/lib64/mpi/gcc/openmpi3/include -I/usr/lib64/mpi/gcc/openmpi3/lib64  -ffree-form -ffree-line-length-none -std=f2003 -O3 -funroll-loops   -fopenmp -pthread -c /users/tiziano/work/dbcsr/dbcsr/build/src/data/dbcsr_data_operations.F -o CMakeFiles/dbcsr.dir/data/dbcsr_data_operations.F.o
[...]
```

so, `-fopenmp` is properly passed along via the `target_link_libraries(dbcsr PRIVATE OpenMP::OpenMP_Fortran)` in `src/CMakeLists.txt`.

Especially with the OpenMP flags we have to be careful since the `nvcc` wrapper didn't like them at all (see the `cuda_convert_flags` for CMake < 3.13).

shoshijak:
When I build DBCSR on Summit, I don't get the `-fopenmp` flag without this patch.

shoshijak:
Since I used to build DBCSR with the Makefile on other platforms, I assumed this was an oversight in the `CMakeLists.txt`. But now I'm curious to know why you get the `-fopenmp` and I don't ...

shoshijak:
What `cmake` version are you using? 
I'm using `3.11.3`

dev-zero:
you should get the following while `cmake`ing:

```
-- Found OpenMP_C: -fopenmp (found version "4.5") 
-- Found OpenMP_CXX: -fopenmp (found version "4.5") 
-- Found OpenMP_Fortran: -fopenmp (found version "4.0") 
-- Found OpenMP: TRUE (found version "4.5")
```

which version of CMake?

shoshijak:
> you should get the following while cmakeing:
> ```
> -- Found OpenMP_C: -fopenmp (found version "4.5") 
> -- Found OpenMP_CXX: -fopenmp (found version "4.5") 
> --  Found OpenMP_Fortran: -fopenmp (found version "4.0") 
> -- Found OpenMP: TRUE (found version "4.5")
> ```

I only get this if I apply the proposed patch. 
Cmake version `3.11.3`

dev-zero:
I've verified it with `3.10.2` (default openSUSE Leap 15.0), but I can reproduce on daint

shoshijak:
Okay, I obtain the correct behavior without the proposed patch if I use `cmake version 3.13.4`.

shoshijak:
So, to summarise: 

| cmake version | Correct build with OpenMP? |
| ------------- |-------------|
| `3.10.2` | dev-zero, openSUSE ✅  shoshijak, OSX ✅  | 
| `3.11.2` | shoshijak, Summit ❌   | 
| `3.12.0` | dev-zero & shoshijak, Daint ✅  | 
| `3.13.4` | shoshijak, Summit ✅  | 

Shall we leave the CMakeList option setting as is (i.e. drop my proposed patch) and just ban `3.11.2`?

dev-zero:
I did some more tests: without CUDA I get `-fopenmp` in the command line flags properly on Daint (cmake 3.12.0) and openSuse (cmake 3.10.x). With CUDA `-fopenmp` is stripped completely, not only for the calls to `nvcc` but also from the calls to `gfortran` (on both daint and tcgpu1).

dev-zero:
ok, I found the issue:

```
# from https://cliutils.gitlab.io/modern-cmake/chapters/packages/CUDA.html
function(CUDA_CONVERT_FLAGS EXISTING_TARGET)
    get_property(old_flags TARGET ${EXISTING_TARGET} PROPERTY INTERFACE_COMPILE_OPTIONS)
    if(NOT "${old_flags}" STREQUAL "")
        string(REPLACE ";" "," CUDA_flags "${old_flags}")
        set_property(TARGET ${EXISTING_TARGET} PROPERTY INTERFACE_COMPILE_OPTIONS
            "$<$<BUILD_INTERFACE:$<COMPILE_LANGUAGE:CXX>>:${old_flags}>$<$<BUILD_INTERFACE:$<COMPILE_LANGUAGE:CUDA>>:-Xcompiler=${CUDA_flags}>"
            )
    endif()
 endfunction()
```

Here I'm accidentally stripping all Fortran flags from the given component (only setting CXX and CUDA options again). So, neither `-fopenmp` nor `-pthread` (from MPI) were passed along anymore to Fortran files.

shoshijak:
Thank you for the testing and finding the issue. 
Good catch :)

shoshijak:
Closing because superseded by #158

Issue #147, opened 2019-02-20
CI for GPU


alazzaro:
This is a place holder for discussion, we really need to implement a CI for GPU.

shoshijak:
I very much agree.
Are there any current obstacles that need to be lifted in order to get a CI for GPU and what needs to be done?

dev-zero:
my plan is/was:

* Jenkins CI at CSCS to run on Daint GPU partition: Jenkins is ready, only needs configuration. Since it is tied to projects (s238, g90), only me, Jürg and Alfio have access to those instances atm (+Patrick for the s238).
* Jenkins CI at HutterGroup infra: Host is there, only have to install a Docker image of Jenkins and configure it. Testing on tcgpu1 or tcgpu2 using the Nvidia Docker (packages for openSUSE [are already prepared](https://github.com/NVIDIA/nvidia-docker/pull/790)). This I started last week, but due to deadlines approaching for other projects, I will not be able to continue until end of next week.

shoshijak:
Great. If there's something I can help with, do let me know. 
Good luck with the approaching deadlines :)

dev-zero:
Progress report: I have a PoC running with the following Jenkins (@CSCS) pipeline configuration:

```groovy
node {
   stage('checkout') {
        checkout([$class: 'GitSCM',
            userRemoteConfigs: [[url: 'https://github.com/cp2k/dbcsr.git']],
            branches: [[name: '*/develop']],
            browser: [$class: 'GithubWeb', repoUrl: 'https://github.com/cp2k/dbcsr'],
            doGenerateSubmoduleConfigurations: false,
            extensions: [[$class: 'SubmoduleOption',
                disableSubmodules: false,
                parentCredentials: false,
                recursiveSubmodules: true,
                reference: '',
                trackingSubmodules: false]],
            submoduleCfg: []
        ])

   }

   stage('build&test') {
        sh 'sbatch --account="${JOB_NAME%%/*}" --job-name="${JOB_BASE_NAME}" --wait /users/timuel/job.sh'
   }
}
```

```bash
#!/bin/bash -l

#SBATCH --export=ALL
#SBATCH --exclusive
#SBATCH --constraint="gpu"
#SBATCH --partition="cscsci"
#SBATCH --time="1:00:00"
#SBATCH --nodes=1
#SBATCH --ntasks-per-node=6
#SBATCH --cpus-per-task=2
[...]

PR #159, opened 2019-04-11
ci: initial configuration for Jenkins@CSCS

dev-zero:
fixes one part of #147 once properly hooked up

### DIFF
--- a/.ci/daint.cscs.ch/build.sh
+++ b/.ci/daint.cscs.ch/build.sh
@@ -1,0 +1,36 @@
+#!/bin/bash -l
+
+#SBATCH --export=ALL
+#SBATCH --exclusive
+#SBATCH --constraint="mc"
+#SBATCH --partition="cscsci"
+#SBATCH --nodes=1
+#SBATCH --ntasks-per-node=4
+#SBATCH --cpus-per-task=3
+#SBATCH --ntasks-per-core=1 # 1=no HT, 2=HT
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+module swap PrgEnv-cray PrgEnv-gnu
+module load daint-gpu cudatoolkit CMake/3.12.0
+module unload cray-libsci_acc
+
+set -o xtrace  # do not set earlier to avoid noise from module
+
+umask 0002  # make sure group members can access the data
+
+mkdir --mode=0775 -p "${SCRATCH}/${BUILD_TAG}"
+cd "${SCRATCH}/${BUILD_TAG}"
+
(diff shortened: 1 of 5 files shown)


alazzaro:
LGTM. I would make this test as optional, is it possible (+ include the coverage)?

dev-zero:
Unfortunately it seems we can't add it to PRs :(

Reason being that CSCS only supports the "github pull requests builder plugin" for PR based builds and that plugin does not work with the modern style Declarative Pipelines yet:

https://github.com/jenkinsci/ghprb-plugin/issues/528
https://github.com/jenkinsci/ghprb-plugin/issues/617

.ci/daint.cscs.ch/build.sh
```diff
@@ -0,0 +1,36 @@
...
+
+set -o xtrace  # do not set earlier to avoid noise from module
+
+umask 0002  # make sure group members can access the data
+
+mkdir --mode=0775 -p "${SCRATCH}/${BUILD_TAG}"
+cd "${SCRATCH}/${BUILD_TAG}"
+
+cmake \
+    -DUSE_CUDA=ON \
```
shoshijak: If these tests are only for the sake of correctness, this is fine. If we want to use them for monitoring performance though, we should link against `libxsmm` instead of `BLAS` (`-DUSE_SMM=libxsmm`)

dev-zero: Good point. Do you happen to know how/where is libxsmm available on Daint?

alazzaro: We use for the regtests: /store/uzh/uzh1/CP2K_ENVIRONMENT/DAINT/libs/PrgEnv-gnu/6.0.4_daint-gpu/libxsmm

It is replicated on scratch too: 

/scratch/snx3000/alazzaro/CP2K_ENVIRONMENT/DAINT/libs/PrgEnv-gnu/6.0.4_daint-gpu/libxsmm

It is 1.9.0, I can install 1.10, although I don't think it is essential...

dev-zero: Will do the libxsmm build in a later commit, especially since we need libxsmm 1.10 for the `pkg-config` file used in `CMake` to correctly link against it.
Hope that's ok.

dev-zero: there is issue #161 as a reminder for that

alazzaro: Fine, I can install libxsmm 1.10 in the next days...

PR #164, opened 2019-04-15
Fix/generated tests cmake

shoshijak:
**Problem**
The `generate_libcusmm_{unittest,timer}_multiply.py` write their output files at the same location as the template, which is `tests/`. This is contrary to CMake convention, where you usually try to contain all writing to the build directory.

**Solution**
- Add a flag to `generate_libcusmm_{unittest,timer}_multiply.py` to explicitly set the output directory
- Set this flag to `$CMAKE_CURRENT_BINARY_DIR` in `CMakeLists`


tests/CMakeLists.txt
```diff
@@ -75,40 +75,48 @@ foreach (dbcsr_test ${DBCSR_TESTS})
...
-    COMMENT "Generate tests/libcusmm_timer_multiply.cu"
-  )
-
-  foreach (libcusmm_test ${LIBCUSMM_TESTS_BUILD})
-    add_executable(${libcusmm_test} ${CMAKE_CURRENT_SOURCE_DIR}/${libcusmm_test}.cu)
+  # Add custom commands for the test files that need to be generated from a template
+  foreach (libcusmm_test ${LIBCUSMM_TESTS_TO_GENERATE})
+    add_custom_command(
+      OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${libcusmm_test}.cu
+      COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/generate_${libcusmm_test}.py --base_dir ${CMAKE_CURRENT_SOURCE_DIR}/.. --out_dir ${CMAKE_CURRENT_BINARY_DIR} --gpu_version=${WITH_GPU}
```
dev-zero: According to the [cmake documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_BINARY_DIR.html) `$CMAKE_CURRENT_BINARY_DIR` is a full path. But below you're expecting `--out-dir` to be relative to `--base_dir`.

It seems this works because Python will return only the second argument of `os.path.join(a, b)` if `b` is an absolute path.

shoshijak: Solved by calculating the relative path using [`file`](https://cmake.org/cmake/help/v3.4/command/file.html)

dev-zero: ah, okay, sorry, didn't see that 🙈

shoshijak: No no, you were right. I just fixed it in my last commit. 
Thanks for the review!

shoshijak:
Should this be merged or should I wait until after version 2.0.0?

alazzaro:
Yes, please, merge.

shoshijak:
Merged already :)

PR #166, opened 2019-04-15
Adjusted way to nullify output-pointers (IFORT)

hfp:
This fixes compilation errors saying that the initialization used is not applicable to the object (IFORT 18.0.5).


hfp:
This should fix build errors with IFORT including CP2K dashboard (e.g., PSI / Merlin5).

hfp:
Well, CP2K may use DBCSR alpha release without this fix.

alazzaro:
Well, the error at PSI seems due to the fact then it doesn't upload the newest DBCSR version (from the log I see that it is still using 1.1.0)... OK, just saw your last comment, I was goint to ask you if we need to do a new release for CP2K... OK, no new release ;)

alazzaro:
Uhm, well, not it gives an error related to NULL pointers:

https://www.cp2k.org/static/regtest/trunk/Linux-x86-64-intel-regtest/popt/regtest-0

Is this what you are referring to @hfp ? If so, we have to release a new alpha version...

hfp:
Yes, this is what I meant and also what's fixed by this PR.

alazzaro:
OK, @pseewald could you make an alpha.2 version and update CP2K?

pseewald:
> OK, @pseewald could you make an alpha.2 version and update CP2K?

done

PR #171, opened 2019-04-17
[WIP] Refactor of dbcsr_api.F grid initialization, timers, and mp_env

dithillobothrium:

Issue #102, opened 2018-11-19
Do not unconditionally print DBCSR STATISTICS and DBCSR MESSAGE PASSING PERFORMANCE 


davydden:
Looking at [dbcsr_example_3.cpp](https://github.com/cp2k/dbcsr/blob/develop/examples/dbcsr_example_3.cpp) and in [dbcsr.h](https://github.com/cp2k/dbcsr/blob/develop/src/dbcsr.h) I did not notice any way to disable printing

```
-------------------------------------------------------------------------------
 -                                                                             -
 -                                DBCSR STATISTICS                             -
 -                                                                             -
 -------------------------------------------------------------------------------
<whole lot of output>
```
and
```
 -------------------------------------------------------------------------------
 -                                                                             -
 -                      DBCSR MESSAGE PASSING PERFORMANCE                      -
 -                                                                             -
 -------------------------------------------------------------------------------
<whole-lot-of-output>
```

I hope this is just a missing C/C++ interface, as opposed to hard-coded unconditional output somewhere in Fortran source code.

alazzaro:
We have introduced a new function for printing statistics:

dbcsr_print_statistics()

Close via 8ff5de28bb3294c80916d7636c8718dc540165ae

PR #178, opened 2019-04-30
Fixed dbcsr_finalize_lib when initialized per dbcsr_finalize_lib

hfp:
Fixed crash in dbcsr_finalize_lib when initialized per dbcsr_init_lib_hooks (tested with CP2K+PR330). The root cause of the issue was following flow (stacktrace):

> dbcsr_finalize_lib  
dbcsr_print_timers  
timings_report_print  
collect_reports_from_ranks  
-> "timer_env => get_timer_env()".

In get_timer_env, "timer_env => list_peek(timers_stack)" fails due to the "timers_stack" not being setup/allocated. The only function that does that i.e., initializing an internal private SAVE variable is "add_timer_env".

PR #171 seems to intent to explicitly omit add_timer_env in case of a hook-based initialization (dbcsr_init_lib_hooks), likely assuming this is controlled externally. This leaves the conclusion that "dbcsr_print_timers" can only be called in dbcsr_finalize_lib in case of the non-hook initialization. The existing "test" of non-hook initialization has been reused (ASSOCIATED(logger)).

Please carefully review and revise this change.

### DIFF
--- a/src/core/dbcsr_lib.F
+++ b/src/core/dbcsr_lib.F
@@ -87,14 +87,18 @@ CONTAINS
       INTEGER, INTENT(IN), OPTIONAL :: io_unit
 
       CALL dbcsr_init_lib_pre(mp_comm, io_unit)
-      ! Declare loggers and timers
+      !
+      ! Declare loggers
       CALL dbcsr_logger_create(logger, mp_env=mp_env, &
                                default_global_unit_nr=ext_io_unit, &
                                close_global_unit_on_dealloc=.FALSE.)
       CALL dbcsr_add_default_logger(logger)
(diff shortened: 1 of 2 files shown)


alazzaro:
Great spot and thanks for fixing the bug!

hfp:
Btw, this patch fixed the crash but the dbcsr statistic vanished from CP2K's output. I believe, DBCSRv2 changes need more testing which is probably best with an updated CP2K integration.

alazzaro:
You have to use: dbcsr_print_statistics()
See #102

hfp:
Makes sense, Thank you!

PR #183, opened 2019-05-10
Check the consistency between dbcsr and libcusmm when building with openmp

### LINKED ISSUE #180: Fortran and C++ OpenMP threads
As suggested here https://github.com/cp2k/cp2k/issues/338#issuecomment-489568559

> As a safe guard you could compare omp_get_num_threads() on the C++ and the Fortran side, e.g. during dbcsr_init().

shoshijak:
In DBCSR's initialisation, if DBCSR is compiled with OpenMP support, add a call checking whether libcusmm was compiled with OpenMP support as well.

See [discussion](https://github.com/cp2k/cp2k/issues/338) in CP2K for the need behind such a check.

Fixes #180

### DIFF
--- a/src/acc/libsmm_acc/include/libsmm_acc.h
+++ b/src/acc/libsmm_acc/include/libsmm_acc.h
@@ -18,6 +18,8 @@ int libsmm_acc_process (void *param_stack, int stack_size,
 int libsmm_acc_transpose (void *trs_stack, int offset, int nblks,
     void *buffer, int datatype, int m, int n, void* stream);
 
+bool libsmm_acc_libcusmm_is_thread_safe ();
+
 #ifdef __cplusplus
  }
 #endif
--- a/src/acc/libsmm_acc/include/libsmm_acc.h
+++ b/src/acc/libsmm_acc/include/libsmm_acc.h
@@ -18,7 +18,7 @@ int libsmm_acc_process (void *param_stack, int stack_size,
 int libsmm_acc_transpose (void *trs_stack, int offset, int nblks,
     void *buffer, int datatype, int m, int n, void* stream);
 
-bool libsmm_acc_libcusmm_is_thread_safe ();
+int libsmm_acc_libcusmm_is_thread_safe ();
 
 #ifdef __cplusplus
  }
--- a/src/acc/libsmm_acc/libcusmm/libcusmm.cpp
+++ b/src/acc/libsmm_acc/libcusmm/libcusmm.cpp
@@ -364,10 +364,10 @@ extern "C" int libsmm_acc_transpose (void *trs_stack, int offset, int nblks, voi
 
 
 //===========================================================================
-extern "C" bool libsmm_acc_libcusmm_is_thread_safe () {
+extern "C" int libsmm_acc_libcusmm_is_thread_safe () {
 #if defined _OPENMP
-    return true;
+    return 1;  // i.e. true, libcusmm is threaded
 #else
-    return false;
+    return 0;  // i.e. false, libcusmm is not threaded
 #endif
 }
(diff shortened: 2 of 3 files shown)


shoshijak:
(Please note these are the first few lines of Fortran I've written in my life, and review critically)

shoshijak:
Unfortunately, it seems like I do not have the necessary permissions to view the output file. Clicking on "Details" takes me to a 404 even though I'm logged into Jenkins. 

Could someone perhaps copy-paste the error?

shoshijak:
I can also find DBCSR's workspace on Daint, but do not have permission to read the files.

alazzaro:
@shoshijak I think you should be part of the UZH group to access Jenkins...

The error message is:
```
ABORT in dbcsr/core/dbcsr_lib.F:197 DBCSR's and libcusmm's threading levels do not match. DBCSR is compiled with threading support while libcusmm is compiled without threading support.
```

Looking at your changes in the code, this is quite strange... Cmake doesn't compile the file libcusmm.cpp with OpenMP.

alazzaro:
Looking at the cmake build part, I see:

```
[  1%] Building CXX object src/acc/libsmm_acc/libcusmm/CMakeFiles/libcusmm.dir/libcusmm.cpp.o
cd /scratch/snx3000/jenkg90/jenkins-g90-DBCSR-102/src/acc/libsmm_acc/libcusmm && /opt/cray/pe/craype/2.5.15/bin/CC  -DARCH_NUMBER=60 -I/scratch/snx3000/jenkg90/jenkins-g90-DBCSR-102/src/acc/libsmm_acc/libcusmm -I/users/jenkg90/workspace/g90/DBCSR/src/acc/libsmm_acc/libcusmm -isystem /opt/nvidia/cudatoolkit9.1/9.1.85_3.18-6.0.7.0_5.1__g2eb7c52/include  -std=c++11 -O3 -funroll-loops   -o CMakeFiles/libcusmm.dir/libcusmm.cpp.o -c /users/jenkg90/workspace/g90/DBCSR/src/acc/libsmm_acc/libcusmm/libcusmm.cpp
```

Therefore there is no OpenMP, which confirm my initial guess... 
@dev-zero Any clue?

shoshijak:
I'll look at this carefully tomorrow morning, it's probably an oversight on my part ...

dev-zero:
I guess we need the following in `src/acc/libsmm_acc/libcusmm/CMakeLists.txt` (respectively [a workaround of it](https://cmake.org/pipermail/cmake/2017-July/065791.html) unless we want to raise the minimal required CMake version to 3.13+)

```cmake
if (OpenMP_FOUND)
  target_link_libraries(libcusmm PRIVATE OpenMP::OpenMP_Fortran)
endif ()
```

alazzaro:
@dev-zero Well, I would assume that OpenMP should be on for all cpp files in the library...
@shoshijak Besides cpp files, do we have to use OpenMP for cu files? I see that there we use NVFLAGS and we don't set OpenMP in the Makefile.inc

dev-zero:
@alazzaro that's one of the fundamental differences between pure make and cmake: in cmake you specify flags per-target rather than per-filetype. While you could simply add the OpenMP flags to compiler flags in the top-level cmake the "modern" way is to inject them via `target_link_libraries`.
... and since libcusmm has its own target, it needs to be added there as well.

dev-zero:
@shoshijak can you PM me your CSCS username please? I'll forward it to make sure you get added to the corresponding project (g90).

dev-zero:
@alazzaro I think `-fopenmp` is one of the flags which has to be hidden behind a `-Xcompilers` flag to be passed down to the backend-end compiler.

shoshijak:
> But then I remember that I checked back when I added those file CMake files and was under the impression that libcusmm was not using OpenMP at that time, is that possible?

It is indeed possible. I think libcusmm started using OpenMP in PR #140. I updated the Makefile back then, but oversaw the missing flags in cmake, I'm sorry. (Another reason to stick to one build system, cf [issue in cp2k](https://github.com/cp2k/cp2k/issues/351) ?)

> @shoshijak can you PM me your CSCS username please? I'll forward it to make sure you get added to the corresponding project (g90).

I joined the project, thanks a lot!

PR #185, opened 2019-05-12
Bugfix/shared lib build

dev-zero:


alazzaro:
Great patch, thanks!
For the record, I would do the minimal changes in the makefile just needed by CP2K. At some point, we should make cmake as the official building tool and leave makefile just for CP2K...

dev-zero:
@alazzaro the patches for makedep are only there to be in sync with CP2K (which gets the updated `makedep.py` in cp2k/cp2k#353)

alazzaro:
Yes, this is exactly what I mean, thanks.

PR #187, opened 2019-05-15
[cmake] add support for mpifort, mpif77, mpif90

shoshijak:
... and add missing cmake options to README

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,13 @@ if (Fortran_COMPILER_NAME MATCHES "ftn")
   list(GET ftn_verbose_OUT 0 Fortran_COMPILER_NAME)
 endif ()
 
+# if the fortran compiler command is an OPENMPI wrapper, find out what the underlying compiler is
+if (Fortran_COMPILER_NAME MATCHES "mpif*")  # mpif77, mpif90 or mpifort
+  execute_process(COMMAND ${Fortran_COMPILER_NAME} --showme OUTPUT_VARIABLE mpif_verbose_OUT ERROR_QUIET)
+  separate_arguments(mpif_verbose_OUT NATIVE_COMMAND "${mpif_verbose_OUT}")
+  list(GET mpif_verbose_OUT 0 Fortran_COMPILER_NAME)
+endif ()
+
 # Depending on the fortran compiler, set the appropriate compiler flags
 if (Fortran_COMPILER_NAME MATCHES "^gfortran.*")
   set(CMAKE_CXX_FLAGS              "-std=c++11")
--- a/README.md
+++ b/README.md
@@ -75,10 +75,13 @@ The configuration flags are (default first):
     -DUSE_OPENMP=<ON|OFF>
     -DUSE_SMM=<blas|libxsmm>
     -DUSE_CUDA=<OFF|ON>
+    -DUSE_CUBLAS=<OFF|ON>
     -DWITH_C_API=<ON|OFF>
     -DWITH_EXAMPLES=<ON|OFF>
     -DWITH_GPU=<P100|K20X|K40|K80|V100>
     -DTEST_MPI_RANKS=<auto,N>
+    -DTEST_OMP_THREADS=<2,N>
+    -DCMAKE_BUILD_TYPE=<Release|Debug|Coverage>
 
 ## Contributing to DBCSR
 


dev-zero:
The first part about the MPI wrappers for the compilers should not be needed, cf
https://github.com/cp2k/dbcsr/blob/3bd4c2c5b4d99b8a67291b02d5f818702a0805c2/src/CMakeLists.txt#L225-L232

shoshijak:
Hmm, yet when I compile with the `mpifort` wrapper around `gfortran`, compilation fails, and I get the following warning: 

```
CMake Warning at CMakeLists.txt:135 (message):
  Unknown compiler, trying with just '-O2'/'-O0 -g'.

  You will most likely have to pass additonal options for free-form Fortran
  2003/2008 for your compiler!

  Please open an issue at https://github.com/cp2k/dbcsr/issues with the
  reported compiler name and the required flags.
```

The changes in this PR fix the issue, because cmake recognizes that `gfortran` is the underlying compiler and the flags in https://github.com/cp2k/dbcsr/blob/develop/CMakeLists.txt#L101-L110 are added.

shoshijak:
Are you saying the user should not compile with the mpi wrappers? i.e. the user should not do: 

```
CC=mpicc; CXX=mpicxx; FC=mpifort; cmake .. etc.
```

alazzaro:
Well, if I can jump in this discussion, that's what I do, nothing to set...

dev-zero:
> Are you saying the user should not compile with the mpi wrappers? i.e. the user should not do:
> 
> ```
> CC=mpicc; CXX=mpicxx; FC=mpifort; cmake .. etc.
> ```

At least that was what I was aiming for with CMake: that the user is able to build in whatever environment without having to specify anything except the flags to configure DBCSR.

... and if people have different MPI implementations they have to control their environment instead (e.g. loading the appropriate module, setting the PATH variable).

On the other are we not that strict about not allowing compiler wrappers since we accept the Cray wrappers and don't support the OpenMPI or MPICH wrappers.

But I guess this way of configuring might also need some (explicit) documentation.

shoshijak:
> At least that was what I was aiming for with CMake: that the user is able to build in whatever environment without having to specify anything except the flags to configure DBCSR.

> ... and if people have different MPI implementations they have to control their environment instead (e.g. loading the appropriate module, setting the PATH variable).

Great, I understand. 
Why not still allow cmake to handle `CC=mpicc; CXX=mpicxx; FC=mpifort` (i.e. controlling of the environment) gracefully? I don't see any downside. 

> But I guess this way of configuring might also need some (explicit) documentation.
I could add that.

dev-zero:
> > At least that was what I was aiming for with CMake: that the user is able to build in whatever environment without having to specify anything except the flags to configure DBCSR.
> 
> > ... and if people have different MPI implementations they have to control their environment instead (e.g. loading the appropriate module, setting the PATH variable).
> 
> Great, I understand.
> Why not still allow cmake to handle `CC=mpicc; CXX=mpicxx; FC=mpifort` (i.e. controlling of the environment) gracefully? I don't see any downside.

I think there were multiple reasons why I didn't try to support it when I first wrote that stuff:

* several implementations provide compiler wrappers: Intel, OpenMPI and MPICH
* when I checked the flags at that time, the `--showme` flags didn't seem to be standardized or even properly documented (which means it would make sense to add a CMake output line stating which backend compiler was detected)
* I considered augmenting the `$PATH` variable to be more user friendly than having to specify the different compiler variables separately (especially when it comes to the fact that for Fortran one has to use a companion C compiler to get the preprocessing done, which makes inadvertently mixing non-matching C and Fortran compilers more problematic)
* by not supporting certain aspects of the configuration and keeping the configuration very simple I hoped to keep people away from strange usage patterns which would require even more consistency checking logic on our end and make them report configuration errors sooner rather than later which would then allow us to properly track/test/tune the configuration process (and also: the fewer ways available, the fewer the possibilities to break something when changing behavior considered to be internal)
* rather than simply using the compiler wrapper I wanted full control over the process, being able to build (and run) executables without the wrapper (for example tests to determine the compilers' Fortran compatibility)
* ... and after all I didn't want the people to see the CMake command line as another `Makefile.inc`

That being said, if you think this is necessary and useful, go ahead, but I would prefer to keep it outside of the documentation and advertise the simple way to configure DBCSR.

shoshijak:
Thanks for the detailed explanation. 
Let me think about the pros/cons for 1-2 more days, then get back to you here.

shoshijak:
@dev-zero, I've been thinking about the points you made yesterday:

I think it's great that DBCSR's cmake just works out-of-the box ("the user is able to build in whatever environment without having to specify anything except the flags to configure DBCSR"), and I agree that we should "keep people away from strange usage patterns". 

However, in some weird environments and when searching for issues, users may want to specify the wrappers as compilers and run something along the lines of

```
CC=mpicc CXX=mpicxx FC=mpifort cmake .. 
```

which breaks without this PR. I ran into this issue myself when testing DBCSR on Summit as did someone else from ORNL. 
I think that supporting this usage makes sense, so I'll merge.

hfp:
I believe this should be merged as it helps those people who struggle to get things working without local intervention or manual adjustments.

I get the value of scientific code and the compromise that people are willing to make because of the tremendous scientific value... but I found that the "begin of the end" is reached if something is hesitant to work "out-of-the-box" (ingratitude is easily rewarded ~ "Undank ist der Welten Lohn"  -:).

shoshijak:
@hfp I agree, PR is already merged :)

hfp:
ohh, that's what I overlooked... ;-)

PR #188, opened 2019-05-21
Rely on mpi_comm_set_errhandler instead of mpi_errhandler_set

hfp:
Rely on `mpi_comm_set_errhandler` instead of `mpi_errhandler_set`. The conditional compilation checks for MPIv3 (`__MPI_VERSION`) which is reasonable but may be not exactly related. This so far fixes linker errors with OpenMPI 4.x (see [here](https://www.open-mpi.org/faq/?category=mpi-removed) for a list of removed [legacy] functions). It is reasonable to assume that MPI function names are standardized since MPIv3 (portable ABI between different implementations of MPI). In addition, this changes also fixes source code comments (typos).


hfp:
This can be merged - the "issue" is -0.1% lower coverage (CODECOV).

PR #189, opened 2019-05-21
MPI wrapper: CONTIGUOUS keyword (MPI needs cont. buffers anyway), fixes temp. arrays (dummy arguments)

hfp:
Since DBCSR moved to F2008, the CONTIGUOUS keyword can be used. MPI requires contiguous buffers hence the wrapper can safely make that statement. In fact, this fixes occurrences of "created temporary array" (unveiled with related GNU Fortran flags). This change can be taken further by adding more uses of the CONTIGUOUS even in the MPI wrapper module (or anywhere else).


hfp:
Ok, this cannot be merged until F2008 is the default requirement. Note that CP2K moved to F2008 by effectively dropping conditional compilation (`__F2008 `). However, DBCSR at this point applies the `-std=f2003` flag for at least GNU Fortran based compilation. This PR may be updated with adjusted compilation flags.

hfp:
This PR depends on #136.

dev-zero:
@hfp I just merged PR #136, please note that while you can use `CONTIGUOUS` was `IS_CONTIGUOUS` just added to GCC 9

hfp:
@dev-zero Thank you, I will update this PR and merge likely later today!

alazzaro:
retest this please

hfp:
The PR passed all tests (which it never did before due to missing F08 flag). This PR has a similar patch in CP2K's MPI-wrapper and it passes there as well since a few weeks. In summary, this PR seems to be ready.

PR #195, opened 2019-06-04
One more CONTIGUOUS applied to MPI-wrapper

hfp:


hfp:
This case was missing (unveiled by comparing CP2K vs DBCSR MPI-wrapper).

PR #198, opened 2019-06-22
Avoid temporary array when expanding/filling into the first row of an 2d-array

hfp:


hfp:
There will be a complementary PR for CP2K. Otherwise this change in DBCSR is borderline if not obscure. It just happens that CP2K calls the original routine (now called "1d") with an array slice of an 2d-array. Non-trivially, this slice has actually a stride wrt passing it to as 1d-array, hence the compiler creates a temporary array. This PR just overloads the routine and adds a 2d-variant which avoid this intermediate/temporary array-copy.

hfp:
This PR is ready for merge (build issues in Travis were present before).

alazzaro:
Thanks Hans! 
Yes, Travis is failing with the macOS tests.

PR #208, opened 2019-07-11
Fix/207

### LINKED ISSUE #207: std::out_of_range in libcusmm
I just got a [crash](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-47ba94bf_report.txt) in libcusmm while running `LIBTEST/dbcsr_types_04.inp`. Maybe a race condition?
```
 *******************************************************************************
 *                             PERFORMANCE TESTS                               *
 *******************************************************************************
Testing with sizes    340    340    340 min/max block sizes     5    16 transposed? F T
 loop     1 with     2 MPI ranks: using        0.063s       138 Mflops/rank
 loop     2 with     2 MPI ranks: using        0.006s      1570 Mflops/rank
 Final checksums   50815313.963378906        22141711.205875367     
Testing with sizes    340    340    340 min/max block sizes     5    16 transposed? F T
terminate called after throwing an instance of 'std::out_of_range'
  what():  _Map_base::at

Program received signal SIGABRT: Process abort signal.

Backtrace for this error:
#0  0x7fe127f272da in ???
#1  0x7fe127f26503 in ???
#2  0x7fe127374f1f in ???
#3  0x7fe127374e97 in ???
#4  0x7fe127376800 in ???
#5  0x7fe1376d0956 in ???
#6  0x7fe1376d6ab5 in ???
#7  0x7fe1376d6af0 in ???
#8  0x7fe1376d6d23 in ???
#9  0x7fe1376d282e in ???
#10  0x5613ecb8e6ae in _ZNSt8__detail9_Map_baseISt5arrayIiLm3EESt4pairIKS2_10omp_lock_tESaIS6_ENS_10_Select1stESt8equal_toIS2_ESt4hashIS2_ENS_18_Mod_range_hashingENS_20_Default_ranged_hashENS_20_Prime_rehash_policyENS_17_Hashtable_traitsILb0ELb0ELb1EEELb1EE2atERS4_
	at /usr/include/c++/7/bits/hashtable_policy.h:774
#11  0x5613ecb87017 in _ZNSt13unordered_mapISt5arrayIiLm3EE10omp_lock_tSt4hashIS1_ESt8equal_toIS1_ESaISt4pairIKS1_S2_EEE2atERS8_
	at /usr/include/c++/7/bits/unordered_map.h:990
#12  0x5613ecb87017 in _Z18libcusmm_process_dPiiP11CUstream_stiiiPdS2_S2_
	at /workspace/cp2k/exts/dbcsr/src/acc/libsmm_acc/libcusmm/libcusmm.cpp:236
#13  0x5613ecb6f9b0 in __dbcsr_acc_operations_MOD_dbcsr_acc_do_mm_stack
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_acc_operations.F:162
#14  0x5613ecb9a41a in __dbcsr_mm_accdrv_MOD_dbcsr_mm_accdrv_process
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_mm_accdrv.F:655
#15  0x5613ecb92a56 in __dbcsr_mm_sched_MOD_dbcsr_mm_sched_process
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_mm_sched.F:409
#16  0x5613ecb711bb in flush_stacks
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_mm_csr.F:891
#17  0x5613ecb711bb in __dbcsr_mm_csr_MOD_dbcsr_mm_csr_purge_stacks
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_mm_csr.F:851
#18  0x5613eca7ed7c in __dbcsr_mm_multrec_MOD_dbcsr_mm_multrec_multiply
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_mm_multrec.F:386
#19  0x5613eca612a6 in __dbcsr_mm_cannon_MOD_multiply_cannon._omp_fn.3
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_mm_cannon.F:1689
#20  0x7fe12795595d in ???
#21  0x7fe138d156da in ???
#22  0x7fe12745788e in ???
#23  0xffffffffffffffff in ???

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 246682 RUNNING AT 3bbec01c287d
=   EXIT CODE: 6
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
YOUR APPLICATION TERMINATED WITH THE EXIT STRING: Aborted (signal 6)
This typically refers to a problem with your application.
Please see the FAQ page for debugging suggestions
EXIT CODE:  6  MEANING:  RUNTIME FAIL
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

shoshijak:
This PR fixes race conditions in libcusmm, in the JIT-ing of multiplication and transpose kernels. 

*Race condition*: an OpenMP thread could read from the lock-map `kernel_locks` while/after the lock-map is being modified.

*Solution*: This commit gets rid of the lock-map, instead just protecting the JIT-ing and the modify-ing of the kernel-handles-map with a critical section. This means that no two (m,n,k)-kernels (even different ones!) can be jit-ed in parallel by two OpenMP threads. Nevertheless, performance is improved because we get rid of locks.

Fix #207

### DIFF
--- a/src/acc/libsmm_acc/libcusmm/libcusmm.cpp
+++ b/src/acc/libsmm_acc/libcusmm/libcusmm.cpp
@@ -22,6 +22,10 @@
 #include <array>
 #include <iostream>
 
+#if defined _OPENMP
+#include <omp.h>
+#endif
+
 #define dbcsr_type_real_4     1
 #define dbcsr_type_real_8     3
 #define dbcsr_type_complex_4  5
@@ -201,58 +205,29 @@ int libcusmm_process_d(int *param_stack, int stack_size, CUstream stream, int m,
     CUfunction kern_func = NULL;
     int threads, grouping;
     Triplet h_mnk = { m, n, k };
+    static bool cpu_fallback = false;
+    std::unordered_map<std::array<int, 3>, kernel_launcher>::iterator kernel_it;
+
+#if defined _OPENMP
+#pragma omp critical (jit_multiplication)
+{
+#endif
 
     // Look up the kernel in the table of already JITed kernels
-    if (kernel_handles.find(h_mnk) == kernel_handles.end()){  // the kernel has not been JIT-ed yet
+    kernel_it = kernel_handles.find(h_mnk);
+    if (kernel_it == kernel_handles.end()){  // the kernel has not been JIT-ed yet
 
-        static bool cpu_fallback = false;
-
-        // Add lock to table of locks and initialize it
-        // (Some serialization here)
-#if defined _OPENMP
-        #pragma omp critical 
-        {
-            if(kernel_locks.find(h_mnk) == kernel_locks.end()){
-                omp_lock_t lock_kernel_mnk;
-                kernel_locks.emplace(h_mnk, lock_kernel_mnk);
-                omp_init_lock(&kernel_locks.at(h_mnk));
-            }
-        }
-
-        // Set lock if it exists
-        if(kernel_locks.find(h_mnk) != kernel_locks.end()){
-            omp_set_lock(&kernel_locks.at(h_mnk));
-        }
-
-        // JIT the kernel using a single thread
-        if(kernel_handles.find(h_mnk) == kernel_handles.end()){
-#endif
-            add_kernel_handle_to_jitted_kernels(kern_func, stream, h_mnk, threads, grouping, cpu_fallback); 
-#if defined _OPENMP
-        }
-
-        // Unset lock and destroy
-        if(kernel_locks.find(h_mnk) != kernel_locks.end()){
-            omp_unset_lock(&kernel_locks.at(h_mnk));
-        }
-
-        #pragma omp critical 
-        {
-            if(kernel_locks.find(h_mnk) == kernel_locks.end()){
-                omp_destroy_lock(&kernel_locks.at(h_mnk));
-                auto it = kernel_locks.find(h_mnk);
-                kernel_locks.erase(it);
-            }
-        }
-#endif
-
-        if(cpu_fallback)
-            return -2; // fall back to CPU
+        add_kernel_handle_to_jitted_kernels(kern_func, stream, h_mnk, threads, grouping, cpu_fallback);
+        kernel_it = kernel_handles.find(h_mnk);
 
     }  // now the kernel has been jitted
 
-    // Look up the kernel in the table of already JITed kernels
-    auto kernel_it = kernel_handles.find(h_mnk);
+#if defined _OPENMP
+}
+#endif
+
+    if(cpu_fallback)
+        return -2; // fall back to CPU
 
     // Retrieve kernel launching parameters
     kern_func = kernel_it->second.kernel_function;
@@ -330,20 +305,29 @@ int libcusmm_transpose_d(int *trs_stack, int offset, int nblks,
 
     // Look up the kernel in the table of already JITed kernels
     Triplet h_mnk = { m, n, 0 };
-    auto kernel_it = transpose_handles.find(h_mnk);
-    if(kernel_it != transpose_handles.end()){  // the kernel has already been JITed
+    std::unordered_map<std::array<int, 3>, CUfunction>::iterator kernel_it;
 
-        kern_func = kernel_it->second; // retrieve handle
+#if defined _OPENMP
+#pragma omp critical (jit_transpose)
+{
+#endif
 
-    } else { // the kernel has not been JIT-ed yet
+    kernel_it = transpose_handles.find(h_mnk);
+    if(kernel_it == transpose_handles.end()){  // the kernel has not been JIT-ed yet
 
         // JIT and store a kernel for this transposition
         jit_transpose_handle(kern_func, m, n);
         transpose_handles.emplace(h_mnk, kern_func);
+        kernel_it = transpose_handles.find(h_mnk);
 
     }
 
-    // Construct argument pointer list and lauch function
+#if defined _OPENMP
+}
+#endif
+
+    // Construct argument pointer list and launch function
+    kern_func = kernel_it->second; // retrieve handle
     int* trs_stack_ = trs_stack + offset; 
     void *args[] = { &trs_stack_, &buffer};
 
--- a/src/acc/libsmm_acc/libcusmm/libcusmm.h
+++ b/src/acc/libsmm_acc/libcusmm/libcusmm.h
@@ -19,10 +19,6 @@
 #include <unordered_map>
 #include <vector>
 
-#if defined _OPENMP
-#include <omp.h>
-#endif
-
 // Macros for CUDA error handling
 // Wrap calls to CUDA NVRTC API
 #define NVRTC_SAFE_CALL(name, x)                                  \
@@ -51,9 +47,6 @@ struct kernel_launcher {
 };
 
 static std::unordered_map<Triplet, kernel_launcher> kernel_handles;
-#if defined _OPENMP
-static std::unordered_map<Triplet, omp_lock_t> kernel_locks;
-#endif
 
 int libcusmm_process_d(int *param_stack, int stack_size,
     CUstream stream, int m, int n, int k,


alazzaro:
Great! Once merged in DBCSR, we can use https://github.com/cp2k/cp2k/pull/461 to check on CP2K CI

src/acc/libsmm_acc/libcusmm/libcusmm.cpp
```diff
@@ -207,41 +211,15 @@ int libcusmm_process_d(int *param_stack, int stack_size, CUstream stream, int m,
```
oschuett: This read access to `kernel_handles` is still unprotected.

src/acc/libsmm_acc/libcusmm/libcusmm.cpp
```diff
@@ -207,41 +211,15 @@ int libcusmm_process_d(int *param_stack, int stack_size, CUstream stream, int m,
 
         static bool cpu_fallback = false;
 
-        // Add lock to table of locks and initialize it
-        // (Some serialization here)
+        // OpenMP version: make sure we JIT the kernel only once for each (m, n, k) (Some serialization here)
 #if defined _OPENMP
```
oschuett: Without OpenMP aren't you now recompiling every time?

oschuett:
I really like the idea of loosing all those locks. It will probably increase startup times even more, but the simplifications are worth it. (Otherwise we can still use a couple of locks and hashing).

Nevertheless, I still think that **any** access to the shared maps must be protected. Hence, I'd write it like this:
```
kernel_launcher kernel_it;

#pragma omp critical (jit_multiplication) {
   kernel_it = kernel_handles.find(h_mnk);
   if (kernel_it == kernel_handles.end()){
      add_kernel_handle_to_jitted_kernels(kern_func, stream, h_mnk, threads, grouping, cpu_fallback); 
      kernel_it = kernel_handles.find(h_mnk);
   }
}
```

src/acc/libsmm_acc/libcusmm/libcusmm.cpp
shoshijak: No, there is a [lookup](https://github.com/shoshijak/dbcsr/blob/fix/207/src/acc/libsmm_acc/libcusmm/libcusmm.cpp#L210) for the kernel a few lines above.

src/acc/libsmm_acc/libcusmm/libcusmm.cpp
shoshijak: True. But does it really have to be? 
The result of this lookup could be:
- `h_mnk` is found in `kernel_handles`: in which case we can just get the kernel handle and use it
- `h_mnk` is *not* found in `kernel_handles`: in which case the thread will (wait and) enter the critical section in which accesses to `kernel_handles` are protected. By the time it enters the critical section, the kernel might have been jitted already, or not. 

I don't see what problems could arise?

shoshijak:
Let me look into the errors in the CI, then get back to you ...

src/acc/libsmm_acc/libcusmm/libcusmm.cpp
oschuett: The problem arises when another thread calls `find` while the first threads calls `emplace`. During the emplace call the map might be in an inconsistent state, e.g. in the middle of rebalancing itself.

shoshijak:
> Nevertheless, I still think that any access to the shared maps must be protected. Hence, I'd write it like this:

Ok, I see how this is safer than the code I suggested. 
However I think it is important to write optimal code for the most frequent case in production, which is that the kernel has already been jit-ed, and we simply need to lookup the triplet in the map and use the handle to the kernel.
In your suggestion, the critical section is used *every time* that multiplications with libcusmm take place; I'm concerned about performance.

oschuett:
I don't think performance will suffer. Once all kernels have been compiled the threads will spend only a very short time in the critical section. And locking/unlocking of the critical section itself should be pretty cheap too.

dev-zero:
one way to optimize it could be to hide the map behind a `std::atomic`: instead of inserting to the structure you make a copy of it and simply replace the pointer to it atomically. The only lock you need is the part where you insert.

shoshijak:
@dev-zero what you suggest would indeed be ideal, but I don't know how to do it with OpenMP: 

> one way to optimize it could be to hide the map behind a std::atomic 

If I'm not mistaken, `std::atomic` would work for `std::threads`, but not for OpenMP threads... 

> (or a #pragma omp atomic)

Not all read/write/update operations can be made atomic by OpenMP. In particular, it seems like `std::unordered_map<Key,T,Hash,KeyEqual,Allocator>::operator[]` and `emplace` cannot (I get a compilation error)

alazzaro:
OpenMP atomic is very restricted to a small set of operations...
I would rather consider omp_set_lock (they are used in mm/dbcsr_mm_3d.F)

shoshijak:
Right, we were using `omp_set_lock` here before as well, but my use of them was buggy (see #207) and in this PR, I was trying to move away from locks.

hfp:
> [...] the kernel has already been jit-ed, and we simply need to lookup the triplet in the map

This is sort of a Reader-Writer lock, which is not available in OpenMP as a separate primitive. Sounds odd (though it's part of Pthreads), but it's hard to find an implementation of RWLocks that actually scales. OpenMP may use lock-elision under the hood (TSX instructions). IMHO ideal would be TBB's concurrent_map (which scales), but I believe it's not part of the new Parallel STL (although PSTL is often implemented using TBB).

With this, it's hard to implement a well-performing "write-once read-many" lock(free) data structure. Atomics in C++ (std::atomic) were unhandy until recently since one got state+behaviour rather than just behavior (aka C99 atomics). OpenMP atomics are a much better choice, even better than C99 especially if the program already adopted OpenMP.

shoshijak:
@hfp , thanks! So would you recommend ..
a) going with an OpenMP critical section (which would have to be called at every read), as @oschuett suggested [earlier](https://github.com/cp2k/dbcsr/pull/208#issuecomment-510400382)
b) use `omp_set_lock`s
c) ... or something different?

alazzaro:
Sorry, I didn't look at the code and the previous comments...
OK, I see the locks now in libcusmm.cpp
I'm not sure what you want to achieve there...

```
#if defined _OPENMP
        #pragma omp critical 
        {
            if(kernel_locks.find(h_mnk) == kernel_locks.end()){
                omp_lock_t lock_kernel_mnk;
                kernel_locks.emplace(h_mnk, lock_kernel_mnk);
                omp_init_lock(&kernel_locks.at(h_mnk));
            }
        }

        // Set lock if it exists
        if(kernel_locks.find(h_mnk) != kernel_locks.end()){
            omp_set_lock(&kernel_locks.at(h_mnk));
        }

        // JIT the kernel using a single thread
        if(kernel_handles.find(h_mnk) == kernel_handles.end()){
#endif
            add_kernel_handle_to_jitted_kernels(kern_func, stream, h_mnk, threads, grouping, cpu_fallback); 
#if defined _OPENMP
        }

        // Unset lock and destroy
        if(kernel_locks.find(h_mnk) != kernel_locks.end()){
            omp_unset_lock(&kernel_locks.at(h_mnk));
        }

        #pragma omp critical 
        {
            if(kernel_locks.find(h_mnk) == kernel_locks.end()){
                omp_destroy_lock(&kernel_locks.at(h_mnk));
                auto it = kernel_locks.find(h_mnk);
                kernel_locks.erase(it);
            }
        }
#endif
```

In this part of code, you want to avoid threads to JIT the same kernel. 
OK for the critical section to generate the lock. But then you have to add a barrier, right?
I mean, while a thread goes in the critical, another thread can try to read the kernel_locks.
Am I wrong?

hfp:
I recommend option (a) i.e., what Ole suggests.  It's hard to beat OpenMP's critical section. On modern CPUs it might almost perform up as if there was no lock (TSX). This is solide, and quick development. Please, also give a name to the critical section otherwise it's the same/unnamed/global critical section for the entire application.

However, if you want to spend more time, option (b) can be of interest if you partition your buffer such that certain kernels only go into certain partitions ("sharded database"). In that case, you can distribute your lock-contention over several locks.

shoshijak:
@alazzaro 
> In this part of code, you want to avoid threads to JIT the same kernel.

Exactly.

> I mean, while a thread goes in the critical, another thread can try to read the kernel_locks.

Yes, you're right. And we think this is the cause of the race condition in #207.

alazzaro:
To be more specific, what you want to achieve is a lock in kernel_handles (the entire map). Therefore the idea is to create a lock map class, where you put a single lock and replace the std::map
This new class will wrap an internal 

`static std::unordered_map<Triplet, kernel_launcher> kernel_handles;`

and then you have to declare 3 methods:
```
emplace
find
end
```

Then you can remove all the other stuff...

For instance, find becomes:
```
find(...) {
set_lock();
call find on kernel_handles
unset_lock();
}
```

shoshijak:
Thanks, @hfp . It seems like @oschuett  's suggestion is also by far the most maintainable/readable. 
I'll do that

oschuett:
I just wanted to point out that this code path is not actually performance critical, It's called only once per stack, ie per 30000 actual block multiplications.

alazzaro:
Sorry, I'm late in this discussion... I'm not sure how you can implement with a critical section.
My idea is to implement a concurrent (OpenMP-aware) unordered map with a single lock. Easy and clean. Then all the rest can be removed and the list does all the job.

shoshijak:
Sorry @alazzaro , just saw your reply ... 
What you suggest would work as well. 
I just pushed my commits implementing a safe version with openmp critical.

alazzaro:
LGTM
@oschuett ?

oschuett:
Looks good!

Nit: I don't think you need `#if defined _OPENMP` around `#pragma omp`. These pragmas should just get ignored if OpenMP is disabled.

alazzaro:
Some compilers will give a warning, so it is better to leave as it is...

PR #209, opened 2019-07-29
C-API to use void-signature for no-arg functions

hfp:
Adjusted C-API to rely on void-signature in case where no functions-argument is expected (C++ would ask for an ellipsis whereas C implicitly uses an ellipsis if the signature is non-void but empty).


hfp:
There are perhaps some more (small) changes possible. For instance, `void c_dbcsr_print(void* matrix)` is likely better expressed with `void c_dbcsr_print(const void* matrix)`. Also, the pointer/star may go with the type rather than the variable name, or at least should be consistent accross the API.

oschuett:
One could also add some typedefs:
```
typedef void* dbcsr_matrix_t;
typedef void* dbcsr_dist_t;
```

PR #211, opened 2019-07-31
Set executable flag (CI scripts), removed trailing whitespaces and tabs

hfp:


hfp:
False positive from code-coverage script (-0.1%).

PR #213, opened 2019-08-01
mpiwrap: mp_get_processor_name and cleanup

hfp:
**Introduced mp_get_processor_name**  
This may be used to introduce the notion of a node. The simplest use would be to count the number of nodes as opposed to the number of ranks. The node-count may help to recommend/hint better MPI/OpenMP decomposition. The latter can complement https://github.com/cp2k/cp2k/tree/master/tools/plan_mpi_omp.

**Misc**
* Fixed routine names mentioned in error messages.
* Some code cleanup (len->LEN).

### DIFF
--- a/src/mpi/dbcsr_mpiwrap.F
+++ b/src/mpi/dbcsr_mpiwrap.F
@@ -93,6 +93,7 @@ MODULE dbcsr_mpiwrap
 #else
    INTEGER, PARAMETER, PUBLIC :: mp_max_library_version_string = 1
 #endif
+   INTEGER, PARAMETER, PUBLIC :: mp_max_processor_name = MPI_MAX_PROCESSOR_NAME
 
    INTEGER, PARAMETER, PUBLIC :: file_offset = MPI_OFFSET_KIND
    INTEGER, PARAMETER, PUBLIC :: address_kind = MPI_ADDRESS_KIND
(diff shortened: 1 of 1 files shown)


hfp:
Code coverage false positive; can be merged.

alazzaro:
Hans, there is already a function to get the hostname of the processor (which is used in CP2K): https://github.com/cp2k/dbcsr/blob/fedb782c811d74f4d714f2e78f8e2a1b856729af/src/base/dbcsr_machine_posix.f90#L288

What's the benefit of using the MPI function?

Concerning your idea of extracting the number of nodes, time ago I did a prototype which was based on the MPI-3 comm MPI_Comm_split_type and MPI_COMM_TYPE_SHARED. It was a brute-force approach though...

hfp:
Hi Alfio, this function likely returns a host's name but not necessarily. According to MPI, it is an identifier unique on a per-node basis. The idea is to run MPI_Gather for the world followed by counting the unique identifiers (this number should be the node-count).

alazzaro:
OK, the code I'm referring in CP2K is:

https://github.com/cp2k/cp2k/blob/ac91bc3e391836d11dfbe433cf78948e2fca5fda/src/environment.F#L229

Joost wrote that function exactly to have a map of the ranks/nodes in order to debug hardware problems (you can enable the output by adding the flag REPORT_MAXLOC

```
&TIMINGS
     THRESHOLD 0.00000000001
     REPORT_MAXLOC
  &END
```

Anyway, I think the MPI version is better. My only suggestion is to use the m_hosttnm as fall-back in the case MPI is enabled, do you agree?

hfp:
ACK, and thank you for the code!  I will add the fallback when implementing the AllGather thing. I believe this functionality can help to emit an insightful/actionable hint (mpirun command) on how to get potentially higher performance. Btw, a second step might be generating such hint under GPU constraints (number of GPUs per node).

PR #220, opened 2019-08-08
Contraction of sub-tensors

dev-zero:
This generalizes the arguments of dbcsr_multiply first_row, first_column,
last_row, last_column, first_k, last_k to tensor contraction


dev-zero:
ok, it seems that since the same origin branch is used as for the previous PR, GH/Travis still link to the failing job. Here is the link to the newly started CI run, let's hope the one referenced here gets updated once it finishes: https://travis-ci.org/cp2k/dbcsr/builds/569321491

PR #221, opened 2019-08-08
Potentially accelerate MPI-only operation by using OpenMP based libxsmm_gemm_batch

hfp:
A user of the DBCSR-library can chose to define DBCSR_LIBXSMM_GEMM_BATCH as `libxsmm_gemm_batch_omp`, and additionally link against libxsmmext (on top of libxsmm and libxsmmf). This change allows to speedup the runtime when relying on POPT-builds (e.g., CP2K).

The symbol DBCSR_LIBXSMM_GEMM_BATCH allows to switch from between `libxsmm_gemm_batch` to `libxsmm_gemm_batch_omp`. This is a minimal change with no complementary adjustment of the build systems; it does not introduce potentially complicated build-logic (Make and CMake).

Note: `libxsmm_gemm_batch_omp` safely operates on a batch of SMMs even if there are duplicated C-indexes (synchronization).

### DIFF
--- a/src/mm/dbcsr_mm_hostdrv.f90
+++ b/src/mm/dbcsr_mm_hostdrv.f90
@@ -163,10 +163,13 @@
   SUBROUTINE xsmm_process_mm_batch_${nametype1}$ (stack_descr, params, &
                                                   stack_size, a_data, b_data, c_data, used_smm)
 #if ${xsmm_supported[n]}$
+#if !defined(DBCSR_LIBXSMM_GEMM_BATCH)
+#define DBCSR_LIBXSMM_GEMM_BATCH libxsmm_gemm_batch
+#endif
      ! Caution: This dependency is ignored by makedep.py, because libxsmm.F is kinda empty.
      USE libxsmm, ONLY: LIBXSMM_GEMM_PRECISION => ${'LIBXSMM_GEMM_PRECISION_F'+bits1[n]}$, &
                         libxsmm_gemm => libxsmm_${nametype1}$gemm, &
-                        libxsmm_gemm_batch, &
+                        libxsmm_gemm_batch => DBCSR_LIBXSMM_GEMM_BATCH, &
                         libxsmm_ptr0
      REAL(${kind1}$), PARAMETER :: one = 1.0_${kind1}$
      INTEGER :: sp


dev-zero:
and how does a potential user finds out about this option? ;-)

This is exactly something which CMake could and should enable/disable depending on the selected option(s).

hfp:
Good point! This can be enabled in CMake and Make independent of this PR. I have done similar in XCONFIGURE, which enables this for POPT-builds. The requirement is to link libxsmmext in addition to libxsmm and libxsmmf. Normally this can be enabled if OpenMP is disabled. The cumbersome part is, the OpenMP runtime must be linked as well (although OpenMP is not used in DBCSR's code).

Moreover, a new (separate) switch may be desired if people really want to have no OpenMP ("pure POPT"); the default behaviour might be discussed also. A separate switch is useful if most if not all executions are planned to execute on all cores using MPI (then a potential OpenMP thread-count == 1; hence no potential speedup from using libxsmmext/libxsmm_gemm_batch).

dev-zero:
> Good point! This can be enabled in CMake and Make independent of this PR. I have done similar in XCONFIGURE, which enables this for POPT-builds. The requirement is to link libxsmmext in addition to libxsmm and libxsmmf. Normally this can be enabled if OpenMP is disabled. The cumbersome part is, the OpenMP runtime must be linked as well (although OpenMP is not used in DBCSR's code).

The requirement to link to the OpenMP runtime is probably provided by the `libxsmmext.pc`, right?
But it seems this requires another configuration option, blurring the line between the popt/psmp/... categorization even further. Anyway, the build system can ensure that the option is only accepted in case OpenMP is disabled.

hfp:
> The requirement to link to the OpenMP runtime is probably provided by the libxsmmext.pc, right?

Yes, this is implemented by listing `-fopenmp` as part of the link-line recommended by pkg-config (`libxsmmext.pc`). It likely requires to use the compiler as a linker-driver (C/C++ or Fortran compiler with GNU-style flags). The latter is practically the standard (as opposed to use `ld` directly). I can implement `-lgomp`, `-liomp5` (or whatever) if somebody prefers this for a good reason.

PR #222, opened 2019-08-09
Move to using Fypp'ed sources as base for the Coverage to hopefully fix/stabilize the reports.

dev-zero:


alazzaro:
Huge drop in coverage, as expected... I think this is the way to go, looking at FYPP output...

dev-zero:
what puzzles me is that codecov tries to match files to the previous structure. Now the source file path should be `build/src/...`, but it still shows only `src/` and the files still contain Fypp macros.
There is a chance that codecov needs to have access to the original files, which means that the total coverage (which it gets directly from the report) is correct, but what is shown in the interface is still wrong since it tries to match the line info for Fypp'ed files to the original ones

alazzaro:
Ah, I see... Anyway, it was overestimated before... Is there anyway to "force" to replace src with the FYPP generated? (I know, a kind of brute force solution...). Or, can we restart codecov?

dev-zero:
@alazzaro no, it's not possible. Codecov **always** matches a source file against the original in the repository. Similar issue is with Coveralls.io: also they upload only the Coverage summary (file, line, number of executions) to their service but never the full files the report was based on, and therefore also rely on the source being checked in.
Their interface is more than just a nicer view of what `genhtml` does: they take the report and match it against the content of the repository at that time.
The only thing we can do to get more than just the number as an entry in the PR is to fallback to manual `genhtml` run and let the CI publish that somewhere (GH pages).
CodeClimate may behave differently, but they don't list Fortran as a supported language :(

alazzaro:
OK, I see. So, nothing to do if we use source-to-source translation... So, unless we create a fake repository (?), I would say it is better to restore the previous settings (i.e. before fypp)... Would do you suggest?

dev-zero:
ok, as soon as https://gitlab.kitware.com/cmake/cmake/issues/18188 is fixed for CMake+Ninja, we can move over to Ninja: https://github.com/dev-zero/dbcsr/tree/bugfix/code-coverage-ninja

Until then #224 ensures that we pass along the proper flags to Fypp to workaround the gfortran5 line-directive issues.

PR #225, opened 2019-08-21
mm: fix RMA detection for distro-provided OpenMPI

dev-zero:
Previously version checks failed with internal errors when the
patch-component contained additional components, which is often the case
for packages provided by Linux distributions.

Example from OpenSUSE:

    Open MPI v3.0.0.0.1cbdf5d279, ...

### DIFF
--- a/src/mm/dbcsr_mm.F
+++ b/src/mm/dbcsr_mm.F
@@ -998,35 +998,54 @@ CONTAINS
    END SUBROUTINE dbcsr_multiply_generic
 
 ! **************************************************************************************************
-!> \brief ...
+!> \brief Check if RMA is used with buggy OpenMPI (version<1.10.4 or 2.0.1)
 ! **************************************************************************************************
    SUBROUTINE check_openmpi_rma()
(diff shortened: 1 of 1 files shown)


alazzaro:
Great! Actually, as I told you in other cases, it seems that other OpenMPI versions are buggy as well... I have the test ready, but this is not a priority at the moment...

dev-zero:
Well, this fixes the RMA tests on the new tcopt machines. Would be nice if we could test the feature instead of the version.

alazzaro:
This is hard to check, the original error was a "false" instead of "true" in a mutex (see https://github.com/open-mpi/ompi/issues/1869#issuecomment-232370579 )
For the newer OpenMPI versions (3.x), I see some other problems: the code hangs, hard to debug, however it is fixed in 4.x... Which version of OpenMPI are you testing?

dev-zero:
`OpenMPI v3.0.0.0.1cbdf5d279` :)

alazzaro:
Is this ready to merge?

PR #230, opened 2019-09-09
Tensors: better heuristics for parameters affecting parallelism

pseewald:
With these changes, process grids and process sub-groups are optimized internally in tensor contraction based on estimated sparsity of result tensor. Subsequent tensor contractions with tensors of similar shape and sparsity can optionally reuse optimized grid.

Finally (with corresponding changes in CP2K, a pull request will follow), we can get rid of the performance-critical parameters [GROUP_SIZE_3C](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/XC/WF_CORRELATION/IM_TIME.html#GROUP_SIZE_3C) and [GROUP_SIZE_P](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/XC/WF_CORRELATION/IM_TIME.html#GROUP_SIZE_P) in CP2K.

This pr comes with some API changes, probably the last tensor API changes before v2.0.

### DIFF
--- a/src/tas/dbcsr_tas_base.F
+++ b/src/tas/dbcsr_tas_base.F
@@ -416,7 +416,7 @@ CONTAINS
          split_info_prv = split_info
       ELSE
          ! default split heuristic: split into submatrices that have roughly same block dimensions
-         IF(row_dist%nmrowcol >= col_dist%nmrowcol) THEN
+         IF (row_dist%nmrowcol >= col_dist%nmrowcol) THEN
             split_rowcol = rowsplit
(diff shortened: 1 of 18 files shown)


alazzaro:
Patrick, could you also take care of #175 ? One example would help people to understand how to use the tensor API...

pseewald:
@alazzaro yes I did not forget, this pr is a preparation to make tensors more easy to use and adding an example is the next step. let's make that a milestone for 2.0 (if we can wait another couple of days).

Issue #157, opened 2019-04-11
[cmake] verify Python installation


dev-zero:
`CMake` should check all requirements to give meaningful error messages instead of the build failing later. Currently it does not check for Python and/or Numpy:

```
Wrote kernel string to file
-> cusmm_kernels.h
Traceback (most recent call last):
  File "../../../../../src/acc/libsmm_acc/libcusmm/generate_parameters.py", line 18, in <module>
    from kernels.cusmm_predict import params_dict_to_kernel
  File "/users/tiziano/work/dbcsr/dbcsr/src/acc/libsmm_acc/libcusmm/kernels/cusmm_predict.py", line 11, in <module>
    import numpy as np
ImportError: No module named numpy
```

dev-zero:
... furthermore, CMake should favor Python 3 over 2 if both are available but Python 2 is the default interpreter.

Issue #217, opened 2019-08-07
Thread number has changed


hfp:
When building and running CP2K as POPT variant (any workload), DBCSR complains:

```text
 *******************************************************************************
 *   ___                                                                       *
 *  /   \                                                                      *
 * [ABORT]                                                                     *
 *  \___/                        Thread number has changed                     *
 *    |                                                                        *
 *  O/|                                                                        *
 * /| |                                                                        *
 * / \                                                dbcsr_dist_methods.F:610 *
 *******************************************************************************
```

Apparently DBCSR is built with OpenMP even in case of CP2K/POPT. In `dbcsr_dist_methods.F:204`, `dist%d%num_threads` is set according to the number of OpenMP-threads. The suggestion is not to build DBCSR without OpenMP, but to revise the check about whether the "Thread number has changed" in a broken way. It should be possible to build and run DBCSR with and without OpenMP independent of CP2K's OpenMP-flag. I suggest to only warn and terminate if the thread-count transitions from N to M with N.GT.1 and N.NE.M.

alazzaro:
This is a bug in CP2K :( Could you tell me where do you get it? In principle, in the popt you should not have any mention of openmp in the arch file...

Yes, the current model has a *static* and not centralized way to handle threads, which is inefficient. So, we have to adapt CP2K to this model.
Let's leave this issue open for any feature activity about threading.

hfp:
I ran the following `exe/Linux-x86-64-intelx/cp2k.popt tests/QS/benchmark_single_node/dbcsr.inp` (any workload should fail similarly) and built with `make ARCH=Linux-x86-64-intelx VERSION=psmp GNU=1 LIBINTROOT=$HOME/libint/gnu-skx   LIBXCROOT=$HOME/libxc/gnu-skx   ELPAROOT=$HOME/elpa/gnu-skx-omp DBG=1 -j48` (CP2K/master). There was no `-fopenmp` flag present for any of the Fortran translation units and DBCSR was built just as part of CP2K (I double-checked with `tee`-ing the output and `grep`-ing for OpenMP). So for that matter any POPT-Archfile should be fine (you do not have to worry about this "Linux-x86-64-intelx" ;-).

alazzaro:
Well, we don't see such a problem on the dashboard...
So, I see that you are using 
`VERSION=psmp`
[...]

PR #241, opened 2019-10-24
Port DBCSR to AMD GPUs

shoshijak:
Port DBCSR to AMD GPUs by adding a [HIP](https://rocm-documentation.readthedocs.io/en/latest/Installation_Guide/HIP.html) backend alongside the already-existing CUDA backend. This PR adds DBCSR support for AMD GPU architectures, in particular (for the time being) for AMD's [Instinct Mi50](https://www.amd.com/en/products/professional-graphics/instinct-mi50).

------

# Overview of Changes

### New Build Options

- `-DUSE_HIP=<OFF|ON>`: HIP can be compiled to run on NVIDIA or AMD GPUs. If you want to run DBCSR on an AMD GPU, choose this option.
- `-DUSE_HIPBLAS=<OFF|ON>`: HIP's equivalent of "Cublas"
- `-DWITH_GPU=<P100|K20X|K40|K80|V100|Mi50>`: new option value: `Mi50`

### CMakeLists

- Require out-of-source builds
- Add comments and re-group commands thematically
- Add Python & Numpy check (see #157). This adds a new `cmake/FindNUMPY.cmake` module.
  - @dev-zero: I hadn't noticed that you self-assigned that issue and started working on it, sorry. I took the liberty of cherry-picking [a commit from your fork](https://github.com/dev-zero/dbcsr/commit/772f3098f59804e25d5810342356386426fd19c5), then supplemented it a little [in this commit](https://github.com/cp2k/dbcsr/commit/a3f068b4535369c19ae3d8bcea828b38bf0d2076).
- New macro definition value: `__DBCSR_ACC == 3` if compiled with HIPBLAS
	
## GPU backend

What was previously known as "libcusmm" (**lib**rary using **cu**da for **s**mall **m**atrix-**m**atrix multiplication) now becomes "libsmm_acc" (**lib**rary for **s**mall **m**atrix-**m**atrix multiplication on a GPU-**acc**elerator)

### Code structure

The number of files in `src/acc/libsmm_acc/libcusmm/` has been growing in the last year. Adding the HIP backend was a good opportunity to review the code structure and make it more hierarchical: 

**Previously**
```
src/acc/libsmm_acc/
-- include/
-- libcusmm_acc/: 
---- kernels/
---- notebooks/
---- EVERYTHING ELSE (generate_*.py, libcusmm*, predict, tune)
```

**Now**
```
src/acc/libsmm_acc/
-- include/
-- kernels/
-- notebooks/
-- parameters/
-- predict/
-- tune/
-- generate_*.py
-- libsmm_acc*
```

### Macros

A set of macros defined in `src/acc/cuda/acc_cuda.h` and `src/acc/hip/acc_hip.h` allow to translate a GPU API call to CUDA or HIP depending on the chosen compiler:

```
#ifdef __CUDA
#include "cuda/acc_cuda.h"
#else
#include "hip/acc_hip.h"
#endif
```

Thanks to these macros, code logic duplication is avoided: we don't have to re-write the same code logic both in CUDA and in HIP, but rather, we can just write it once, and let the macros expand the code to the right language during pre-processing. This way, DBCSR can be compiled 

a) with CUDA, targeting NVIDIA GPU
b) with HIP, targeting either NVIDIA GPU or AMD GPU

#### Example CUDA code:

```
int cuda_error_check (cudaError_t cudaError){
  if (cudaError != cudaSuccess){
      printf ("CUDA Error: %s\n", cudaGetErrorString (cudaError));
       return -1;
     }
   return 0;
 }
```

#### Example code with macros, which can be expanded either to HIP or to CUDA:

```
int acc_error_check (ACC(Error_t) error){
  if (error != ACC(Success)){
      printf ("GPU BACKEND error: %s\n", ACC(GetErrorString)(error));
       return -1;
     }
   return 0;
 }
```

### GPU Kernels

- Rename GPU kernels "cusmm" -> "smm_acc"
- Add GPU properties for AMD Mi50.
- HIP and CUDA are so similar, that the kernel code itself doesn't need to be changed
- The only thing that does change: up to now, the warp size was hard coded in CUDA kernels. But AMD and NVIDIA GPUs have different warp sizes, so this value needs to be parametrised.
  - Read warp_size in kernels from gpu_properties.json.
    - Why: Some GPU kernels in libsmm_acc need to know the GPU's warp size. Up to now, this information was simply hard-coded in each kernel (`const int warp_size = 32;`). However, AMD GPUs have a warp size of 64, not 32, and this needs to be reflected in the kernel code.
    - How: For the kernels concerned, add a warpSize template parameter. The warpSize is defined in the `parameter.h` header, generated through a python script at compile time, which reads the warp size from `gpu_properties.json`.
    - Add check: at initialization of dbcsr library, check whether the warp size as recorded in the JSON file and the warp size as returned from the CUDA/HIP API are equal, to verify consistency.
    - Adapt the autotuning scripts accordingly.

### libsmm_acc_init.h/cpp

Group `libsmm_acc` startup checks and the like in `src/acc/libsmm_acc/libsmm_acc_init.h/cpp`.

### Parameters

Add `src/acc/libsmm_acc/parameters/parameters_Mi50.json` containing 10 autotuned parameter sets for AMD's Mi50.

### Autotuning and Optimal Parameter Prediction

- Add support for the HIP backend in the relevant scripts (i.e. adapt launcher codes, e.g.).
- `libsmm_acc/tune`: make batch script more general: remove account info and Programming Environment version number.
- tuning: AMD's `hipcc` compiler is still very slow: make smaller autotuning folders (less "launchers" per executable") if autotuning on AMD.
- `tune_setup` has a new argument: `--compiler` to indicate whether to use the HIP or CUDA backend.

## Tests

`tests/H2O performance`: increase repetitions 3 -> 50 for more reliable performance measurements.

## Docs

Fix a few typos in Markdown files


alazzaro:
Fantastic! Please give us some time to review :)

dev-zero:
it seems the `fypp` submodule got updated but I couldn't find a reference to that in one of the commit msgs. Was that intentional?

alazzaro:
After a fast review, All changes LGTM.
From the user perspective, now libcusmm becomes libsmm_acc (makes sense for a general library) and we have a new macro for HIP, i.e. `__DBCSR_ACC == 3`. @shoshijak Could you please document it in the Makefile.inc? (I think it is not there) Makefile.inc is still what we need to build CP2K...

Question: can we test HIP backend at CSCS on the NVIDIA GPUs?

Concerning an ETA, pretty soon (I hope this w-e) we will release v2.0, which is the baseline for tensors development in CP2K. After that, we can merge this PR and where ready we can release v2.1.

dev-zero:
> ### CMakeLists
> * Require out-of-source builds
> * Add comments and re-group commands thematically

Instead of horizontal bars I would rather prefer we start factoring out grouped commands to included files instead, similar to the `CompilerConfiguration.cmake`.

> * Add Python & Numpy check (see #157). This adds a new `cmake/FindNUMPY.cmake` module.
>   
>   * @dev-zero: I hadn't noticed that you self-assigned that issue and started working on it, sorry. I took the liberty of cherry-picking [a commit from your fork](https://github.com/dev-zero/dbcsr/commit/772f3098f59804e25d5810342356386426fd19c5), then supplemented it a little [in this commit](https://github.com/cp2k/dbcsr/commit/a3f068b4535369c19ae3d8bcea828b38bf0d2076).

That's perfectly fine. Thanks for picking it up!

Everything else looks good to me.

dev-zero:
@shoshijak sorry for the conflicts, I have a rebased branch I could force push if you want

shoshijak:
@dev-zero no problem at all, that's what happens when you open a PR then go on vacation 😅.
I will fix the conflicts, I have a couple of little fixes I'm working on right now.

dev-zero:
ok, it would be here: https://github.com/dev-zero/dbcsr/tree/feature/amd-backend
most notable conflict besides the documentation tags in Fortran files is my CXX_STANDARD in `CMakeLists.txt`

shoshijak:
Thanks!

alazzaro:
OK, now v2.0 is out, this PR will make v2.1.
The plan is to do a RC, update CP2K and then test it. 
@shoshijak Let's us know where you are ready!

shoshijak:
About a comment I made in the PR description, namely: 

> ### GPU Kernels
> The only thing that does change: up to now, the warp size was hard coded in CUDA kernels. But AMD and NVIDIA GPUs have different warp sizes, so this value needs to be parametrised.
Read warp_size in kernels from gpu_properties.json.
> ...
> - How: For the kernels concerned, add a warpSize template parameter. The warpSize is defined in the `parameter.h` header, generated through a python script at compile time, which reads the warp size from gpu_properties.json.

It turns out, one can use the built-in variable `warpSize` in kernel code for both `nvcc` and `hipcc`. Therefore, no need to add `warpSize` as a template parameter, I simply replaced the hard-coded value by the built-in variable in the kernel code.

src/CMakeLists.txt
```diff
@@ -190,56 +272,55 @@ if (USE_CUDA)
...
-  target_compile_definitions(dbcsr PRIVATE __NO_STATM_ACCESS)
+# =================================================================================================
+# DBCSR LIBRARY's HIP BACKEND
 
-  if (BLAS_LIBRARIES MATCHES "Accelerate")
-    target_compile_definitions(dbcsr PRIVATE __ACCELERATE)
+if (USE_HIP)
+
+  if (OpenMP_FOUND)
+    set(HIP_HIPCC_FLAGS "${HIP_HIPCC_FLAGS} -fopenmp")
```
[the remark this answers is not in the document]
shoshijak: you're right, thanks! this is fixed now.

shoshijak:
> it seems the fypp submodule got updated but I couldn't find a reference to that in one of the commit msgs. Was that intentional?

No, it was not intentional, thanks for catching that. Fixed now.

shoshijak:
> ... we have a new macro for HIP, i.e. __DBCSR_ACC == 3. @shoshijak Could you please document it in the Makefile.inc? (I think it is not there) Makefile.inc is still what we need to build CP2K...

@alazzaro Right, I did not update `Makefile` nor `Makefile.inc` in this AMD work, I only updated and documented cmake.
This is why one Travis check (" Build and test MPI+OpenMP version with make") is failing.
Maintaining both a Makefile and a CMakeLists for DBCSR is unsustainable, especially in the case of large PRs like this one. 

I would like to  delegate building DBCSR to the toolchain instead of building DBCSR inside CP2K (see [cp2k issue](https://github.com/cp2k/cp2k/issues/351)) in order to prevent this kind of work, and remove the need for documenting compilation options twice, in DBCSR and CP2K.

Moreover, about `__DBCSR_ACC == 2, 3`: see issue #254 on how to solve this.

shoshijak:
> Question: can we test HIP backend at CSCS on the NVIDIA GPUs?

In theory, yes. In practice:
- on our tesbed Ault: some nvrtc headers are missing in the hipcc installation, so I cannot compile the HIP backend targetting NVIDIA GPUs
- on Daint: the hipcc compiler is not installed. I would like to try to install it manually, but haven't gotten around to it yet.

shoshijak:
> Instead of horizontal bars I would rather prefer we start factoring out grouped commands to included files instead, similar to the CompilerConfiguration.cmake.

I see. If you agree, I'd like to not do any further work on CMakelists in this PR, and rather factor out grouped commands when I work on #254  when I'll have to go over the cmake again anyway.

@dev-zero Would that be okay with you?

shoshijak:
About Python & Numpy: 
I was trying to fix fails in Travis and got fed up, and realised that building DBCSR really should not require Numpy. So I took this dependency out in [this commit](https://github.com/cp2k/dbcsr/pull/241/commits/34498824ed6e1e851b6de86c332edc09952395f2). This should make our lives much easier.

shoshijak:
> @shoshijak Let's us know where you are ready!

As far as I'm concerned, I'm ready. @alazzaro and @dev-zero, if you both agree with the answers above, then this is good to be merged.

alazzaro:
OK, concerning DBCSR migration to cmake (and drop Makefile), my opinion is clear on this point (cmake and cmake). I understand the frustration of having two build systems...

However, I cannot make any decision on CP2K. @oschuett and @dev-zero are probably the right persons for that. 
From what I can see, we need to do some work there, for example we should introduce a way to get DBCSR version (nothing special though) and make sure that CP2K compilation is compatible with DBCSR one (for the time being, you will get trouble if you have CP2K w/o OpenMP and DBCSR w/, see https://github.com/cp2k/dbcsr/issues/217).

My two cents:
1. Assuming the need to have a new CP2K release this year, I would not change DBCSR compilation now. The current DBCSR v2.0 is ready for that. 
2. This PR will be part of v2.1-rc1, so I would still keep the current design of the Makefile and cmake, and leave to >=v3.0 for any change in the build system.
 
So, unless we want to go directly for v3.0 (2) and delay CP2K (1), I would like to reiterate my request on adding some a descriptio in Makefile.inc for `-D__DBCSR_ACC=3` for a future reference in CP2K. It can be under 

https://github.com/cp2k/dbcsr/blob/118f2157e9b499a0705c0610f84968493c684acf/Makefile.inc#L66

You just need to add a comment on how people can use ``-D__DBCSR_ACC=3`... I know, it is a waste of time in the long run...

@dev-zero @oschuett Any comment on DBCSR in CP2K?

dev-zero:
> > Instead of horizontal bars I would rather prefer we start factoring out grouped commands to included files instead, similar to the CompilerConfiguration.cmake.
> 
> I see. If you agree, I'd like to not do any further work on CMakelists in this PR, and rather factor out grouped commands when I work on #254 when I'll have to go over the cmake again anyway.
> 
> @dev-zero Would that be okay with you?

sure

alazzaro:
> > Question: can we test HIP backend at CSCS on the NVIDIA GPUs?
> 
> In theory, yes. In practice:
> 
> * on our tesbed Ault: some nvrtc headers are missing in the hipcc installation, so I cannot compile the HIP backend targetting NVIDIA GPUs
> * on Daint: the hipcc compiler is not installed. I would like to try to install it manually, but haven't gotten around to it yet.

No problem, as far as we stay with a RC, we don't need a proper test mechanism in place, mainly for CP2K (currently the way to release an official DBCSR release is to test it in CP2K, and of course any official CP2K release requires an official DBCSR one...). We can open an issue for that as a reminder (HIP testings). For instance, we can try some container options...

shoshijak:
@alazzaro Ok, I understand your point of view on versioning. I'll fix the Makefile and Makefile.inc so they support this current PR, and push for separating DBCSR's compilation from CP2K after that.

dev-zero:
Wrt DBCSR in CP2K:

CP2K and DBCSR are still in a co-development state (see timings between rc-releases of DBCSR and updates of the submodule), except for the accelerator part obviously. At the moment people will get an automatic rebuild of DBCSR whenever the submodule is updated, while the other way round they would simply get build failures in CP2K (best case scenario, worst case it builds and they get wrong results) to remind them to manually update the toolchain as well. Given that there are enough (mostly barely tested as it seems) changes to the toolchain at the moment by 3rd-party projects I really would like to avoid throwing DBCSR into this mix for now.

shoshijak:
Ok, I'll fix the make build for this PR, for now. 
@dev-zero copying your reply to the relevant cp2k issue.

src/acc/acc_stream.cpp
```diff
@@ -25,10 +30,8 @@ extern "C" int acc_stream_priority_range(int* least, int* greatest){
   *least = -1;
   *greatest = -1;
 
-#ifndef __HAS_NO_CUDA_STREAM_PRIORITIES
-  cudaError_t cErr = cudaDeviceGetStreamPriorityRange(least, greatest);
-  if (cuda_error_check(cErr)) return -1;
-  if (cuda_error_check(cudaGetLastError())) return -1;
+#ifndef __ACC_HAS_NO_STREAM_PRIORITIES
```
dev-zero: CUDA 5.5 for which this flag was (according to CP2K's `INSTALL.md`) was released in 2013. I'd suggest to drop it and require CUDA >= 5.5, guessing that nobody actually still tests with a CUDA that old.

shoshijak: Agreed, that makes a lot of sense.

alazzaro:
Hello Shoshana, I was just checking the Makefile.inc changes. I don't know if you have completed the changes (apologize if this is the case), but I see that there is now in the top description `NVCC` (which is used in the Makefile.inc) and HIPCC (same for the flags `NVFLAGS` and `HIPFLAGS`). Then I see you did a flag GPU=2, which is still setting the variables NV*. Likely, it would make sense to have ACC* variables for both NV and HIP, but that would require to change CP2K too... 
Anyway, I don't like the idea to have the variable `GPU` in the makefile to check which version we are using... The `GPU` is just a convenient way to use the Makefile.inc.
Assuming that we keep `NVCC`, would it be possible to make a condition based on 

`ifneq (,$(findstring nvcc,$(NVCC)))`

or

`ifneq (,$(findstring hipcc,$(NVCC)))`

shoshijak:
> I see that there is now in the top description NVCC (which is used in the Makefile.inc) and HIPCC (same for the flags NVFLAGS and HIPFLAGS). Then I see you did a flag GPU=2, which is still setting the variables NV*. 

Thank you for catching that, sorry, I started out thinking that I would use two separate variables `NVCC` and `HIPCC`, then I realised exactly what you wrote: 

> Likely, it would make sense to have ACC* variables for both NV and HIP, but that would require to change CP2K too...

And it seemed to me that changing the interface to CP2K was too much overhead now.

So I guess we now need to decide how to deal with 2 GPU backend and the Makefile, I see following possibilities:

#### 1. Support CUDA and HIP in the Makefile

> Anyway, I don't like the idea to have the variable GPU in the makefile to check which version we are using... The GPU is just a convenient way to use the Makefile.inc.
Assuming that we keep NVCC, would it be possible to make a condition based on
>
> ifneq (,$(findstring nvcc,$(NVCC)))
>
> or
>
> ifneq (,$(findstring hipcc,$(NVCC)))

Making the backend choice via the `NVCC` variable means users will have to write things like `make NVCC=hipcc` which is not impossible, but a bit weird. It seems to me that `make GPU=1`, `make GPU=2` or `make GPU=CUDA`, `make GPU=HIP` is more intuitive... 

Advantage: supports both compilation with CUDA and HIP in the Makefile
Disadvantage: Makefile gets a little uglier, and it's quite a bit of work, but I've already done it. Changes to CP2K/DBCSR interface are required to add the AMD acceleration option to CP2K.

#### 2. Don't support HIP compilation in the Makefile

Advantages: Minimal work and cleaner Makefile, no change to CP2K/DBCSR interface
Disadvantages: People who want to run DBCSR on AMD have to use cmake.

#### 3. Drop the Makefile completely, and let CP2K compile DBCSR with cmake

Advantages: only 1 build system, clean building syntax
Disadvantage: requires (rather minor) changes to CP2K in its building of DBCSR

alazzaro:
So, for the time being, we don't need CP2K to use the AMD backend (we have to wait for the release to be out), we just need to be consistent within DBCSR...

I propose the option 1, however I would change the NV* to ACC* variables within Makefile and Makefile.inc and make use of condition with findstrings to switch between solutions within the Makefile. Really, GPU= is only for the Makefile.inc, we cannot have it in CP2K...
I think this is the best solution to make DBCSR consistent.

Then, for the CP2K part, currently we have these lines:

https://github.com/cp2k/cp2k/blob/836046e5d55ab12c7fa8ee337a7303aaf45794b7/exts/Makefile.inc#L15-L20

I think, it would be enough to acc ACC=$(NVCC) and ACCFLAGS=$(NVFLAGS) in the command line. Then, whenever CP2K wants to use HIP, we can improve the call and document in CP2K the two backends...

In any case, I propose to proceed with the merge and then we can discuss within an issue how to change the documentation and the inclusion in CP2K...

I see a problem in the Jenkins for the Intel: time limit... I triggered to rerun it...

alazzaro:
The cmake command for the Intel compiler takes too much time on Daint and the job gets canceled due to time limit (30m!). I'm not sure why...
@shoshijak could you give a try directly on Daint and see if it works following the steps at:

https://github.com/cp2k/dbcsr/blob/develop/.ci/daint.cscs.ch/intel.build.sh

?

shoshijak:
> @shoshijak could you give a try directly on Daint and see if it works following the steps at:
>
> https://github.com/cp2k/dbcsr/blob/develop/.ci/daint.cscs.ch/intel.build.sh

I just tried these steps out on Daint and they complete in ~5 minutes. I'm not sure why they take a lot longer in Jenkins.

alazzaro:
Thanks for the test!
Yes, the problem was on the CSCS said, very slow access tot the Intel compiler.
I've restarted the test.

Besides the Makefile.inc documentation, that we can fix in a different PR, is there any other issue for this PR? 
@dev-zero any other comment?

shoshijak:
> Besides the Makefile.inc documentation

I'm on it ...

alazzaro:
For the record, a new jenkins run passed OK...

shoshijak:
> For the record, a new jenkins run passed OK...

Great!

alazzaro:
I must admit, I got confused by the changes in Makefile/Makefile.inc :)
Since there is not really functional change in the code that prevents this PR to be merged, I propose to merge and discuss elsewhere the makefile (see #263)... As far as cmake is working and @dev-zero is happy, I'm happy to merge...

shoshijak:
> I must admit, I got confused by the changes in Makefile/Makefile.inc :)

I'm sorry, what confused you? I did the two changes you mentioned: 

1. Remove use of the variable `GPU` in the Makefile
`ifneq ($(GPU),)	` -> `ifneq ($(NVCC),)`
`ifeq ($(GPU),1)	` -> `ifneq (,$(findstring nvcc,$(NVCC)))`
`ifeq ($(GPU),2)	`-> ``ifneq (,$(findstring hipcc,$(NVCC)))`
2. Adapt docs at the top of Makefile.inc
3. tiny minor fixes

Great, if @dev-zero is happy, I'm happy to merge too

dev-zero:
ACK from me :)

Issue #242, opened 2019-10-24
Inefficiencies with large memory pools


pseewald:
I ran into this issue when testing cp2k/cp2k#581 which is mainly based on DBCSR tensors. After contraction of large tensors, DBCSR multiplication of much smaller matrices takes much more time than with the reference methods that does not use tensors. This must be connected to the memory pools because this inefficiency disappears when I call `dbcsr_clear_mempools()` after all tensor contractions are done.

Here some timings output:

Timings without calling `dbcsr_clear_mempools()`:
```
 SUBROUTINE                       CALLS  ASD         SELF TIME        TOTAL TIME
                                MAXIMUM       AVERAGE  MAXIMUM  AVERAGE  MAXIMUM
multiply_cannon_multrec           3051 13.6   15.324   21.590   26.355   31.443
dbcsr_mm_multrec_finalize          280 14.9    0.003    0.004   11.031   15.310
dbcsr_mm_sched_finalize            280 15.9   10.981   15.263   10.981   15.263

```
Timings with calling `dbcsr_clear_mempools()`:
```
 SUBROUTINE                       CALLS  ASD         SELF TIME        TOTAL TIME
                                MAXIMUM       AVERAGE  MAXIMUM  AVERAGE  MAXIMUM
multiply_cannon_multrec           3051 13.6   11.731   12.871   11.781   12.919
dbcsr_mm_multrec_finalize          280 14.9    0.002    0.003    0.049    0.125
dbcsr_mm_sched_finalize            280 15.9    0.001    0.101    0.001    0.101
```

I'm testing the code with pure MPI (no GPUs and 1 thread per MPI rank) on mc partition of Piz Daint. I can share CP2K input once cp2k/cp2k#581 is merged.

Another obvious issue with memory pools is that memory stays reserved for DBCSR mm but may be needed for other operations. For instance, the memory bottleneck of the DBCSR tensor implementation is copying/redistribution operations after matrix multiplication, but only if memory pools are not cleared after matrix multiplication.

vondele:
probably because some operations use the full size of the memory buffer, whereas they should be using the size of the data? Might be a good testcase to figure out.

PR #244, opened 2019-10-29
Disable mempools on CPUs by default

pseewald:
This circumvents the following issues #242:
1) memory may be needed otherwise
2) large memory pools cause inefficiencies for small matrix
multiplications

Issue 2) should be properly fixed but the presence of this configuration
option will still be justified by 1).

### DIFF
--- a/src/core/dbcsr_config.F
+++ b/src/core/dbcsr_config.F
@@ -305,6 +305,7 @@ CONTAINS
 !> \param accdrv_do_inhomogenous ...
 !> \param accdrv_binning_nbins ...
 !> \param accdrv_binning_binsize ...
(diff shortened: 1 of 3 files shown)


alazzaro:
LGTM. Any reason why the test is failing?

pseewald:
All tests pass, except for a formatting issue with a Python file I did not touch.

alazzaro:
OK, likely @dev-zero has an idea?

dev-zero:
Since running the `pre-commit` hooks only on changed files turned out to be unreliable and it's really fast to run them on all files in DBCSR, conventions are run on all files.
One of the tools got updated and seems to complain now about that one file. I'll merge #245 and then you can rebase.

PR #249, opened 2019-11-07
Support GNU toolchain and Intel MPI

hfp:
Ensure CMake attempts to determine the version/vendor string of the MPI component (MPI_DETERMINE_LIBRARY_VERSION=true). Define __HAS_NO_MPI_MOD preprocessor variable if non-Intel (Fortran-)compiler but Intel MPI is used. Apply -D__HAS_NO_MPI_MOD on a global basis. Support __HAS_NO_MPI_MOD (examples).

Note: attempts to apply settings more locally did not work out for various reasons (`get_source_file_property` and friends). In the end, the HAS_NO_MPI_MOD must be global and even apply to example code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,6 +106,13 @@ foreach(line ${VERSION_INFO})
   endif ()
 endforeach()
 
+# __HAS_NO_MPI_MOD controls if mpi.f is included which may not accept strict f2008 flag during compilation
+if (MPI_FOUND AND (MPI_Fortran_LIBRARY_VERSION_STRING MATCHES "Intel") AND (NOT ${CMAKE_Fortran_COMPILER_ID} MATCHES "Intel"))
+  # get_target_property for COMPILE_OPTIONS is apparently not working even with own/separate target (NOTFOUND)
+  # see also: https://stackoverflow.com/questions/28344564/cmake-remove-a-compile-flag-for-a-single-translation-unit
+  STRING(REGEX REPLACE "-std=.[^ ]*" "" CMAKE_Fortran_FLAGS ${CMAKE_Fortran_FLAGS})  # mpi.f is not strict f2008
+endif ()
+
 add_subdirectory(src)
 add_subdirectory(tests)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,7 @@ find_package(PkgConfig)
 if (USE_MPI)
   get_property(REQUIRED_MPI_COMPONENTS GLOBAL PROPERTY ENABLED_LANGUAGES)
   list(REMOVE_ITEM REQUIRED_MPI_COMPONENTS CUDA)  # CUDA does not have a MPI component
+  set(MPI_DETERMINE_LIBRARY_VERSION true)  # prior to find_package
   find_package(MPI COMPONENTS ${REQUIRED_MPI_COMPONENTS} REQUIRED)
 endif ()
 
@@ -104,6 +105,12 @@ foreach(line ${VERSION_INFO})
     continue ()
   endif ()
 endforeach()
+
+# must be global since modified FLAGS apply to other subdirectories as well, e.g., examples
+# must be after include(CompilerConfiguration) due to modifying CMAKE_Fortran_FLAGS
+if (MPI_FOUND AND (MPI_Fortran_LIBRARY_VERSION_STRING MATCHES "Intel") AND (NOT ${CMAKE_Fortran_COMPILER_ID} MATCHES "Intel"))
+  STRING(REGEX REPLACE "-std=.[^ ]*" "-D__HAS_NO_MPI_MOD" CMAKE_Fortran_FLAGS ${CMAKE_Fortran_FLAGS})
+endif ()
 
 add_subdirectory(src)
 add_subdirectory(tests)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,12 +106,6 @@ foreach(line ${VERSION_INFO})
   endif ()
 endforeach()
 
-# must be global since modified FLAGS apply to other subdirectories as well, e.g., examples
-# must be after include(CompilerConfiguration) due to modifying CMAKE_Fortran_FLAGS
-if (MPI_FOUND AND (MPI_Fortran_LIBRARY_VERSION_STRING MATCHES "Intel") AND (NOT ${CMAKE_Fortran_COMPILER_ID} MATCHES "Intel"))
-  STRING(REGEX REPLACE "-std=.[^ ]*" "-D__HAS_NO_MPI_MOD" CMAKE_Fortran_FLAGS ${CMAKE_Fortran_FLAGS})
-endif ()
-
 add_subdirectory(src)
 add_subdirectory(tests)
 
--- a/.ci/daint.cscs.ch/gnu.build.sh
+++ b/.ci/daint.cscs.ch/gnu.build.sh
@@ -27,6 +27,7 @@ cd "${SCRATCH}/${BUILD_TAG}.gnu"
 
 cmake \
     -DCMAKE_SYSTEM_NAME=CrayLinuxEnvironment \
+    -DCMAKE_CROSSCOMPILING_EMULATOR="" \
     -DUSE_CUDA=ON \
     -DUSE_CUBLAS=ON \
     -DWITH_GPU=P100 \
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -16,6 +16,12 @@ foreach (dbcsr_program_src ${DBCSR_PROGRAM_SRCS})
   set_target_properties(${dbcsr_program_name} PROPERTIES LINKER_LANGUAGE Fortran)
 endforeach ()
 
+# __HAS_NO_MPI_MOD controls whether mpi.mod is used or if mpi.f is included
+if (MPI_FOUND AND (MPI_Fortran_LIBRARY_VERSION_STRING MATCHES "Intel") AND (NOT ${CMAKE_Fortran_COMPILER_ID} MATCHES "Intel"))
+  STRING(REGEX REPLACE "-std=.[^ ]*" "" CMAKE_Fortran_FLAGS ${CMAKE_Fortran_FLAGS})  # mpi.f is not strict f2008
+  target_compile_definitions(${dbcsr_program_name} PRIVATE __HAS_NO_MPI_MOD)
+endif ()
+
 if (WITH_C_API)
   add_executable(dbcsr_example_3_cpp dbcsr_example_3.cpp)
   target_link_libraries(dbcsr_example_3_cpp dbcsr_c MPI::MPI_CXX)
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -15,7 +15,7 @@ foreach (dbcsr_program_src ${DBCSR_PROGRAM_SRCS})
   # and needs to be told explicitly:
   set_target_properties(${dbcsr_program_name} PROPERTIES LINKER_LANGUAGE Fortran)
 
-  # __HAS_NO_MPI_MOD controls whether mpi.mod is used or if mpi.f is included
+  # __HAS_NO_MPI_MOD controls whether mpi.mod is used or if mpi.f is included; mpi.f may not accept flags about strict Fortran standard
   if (MPI_FOUND AND (MPI_Fortran_LIBRARY_VERSION_STRING MATCHES "Intel") AND (NOT ${CMAKE_Fortran_COMPILER_ID} MATCHES "Intel"))
     target_compile_definitions(${dbcsr_program_name} PRIVATE __HAS_NO_MPI_MOD)
   endif ()
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -17,7 +17,7 @@ foreach (dbcsr_program_src ${DBCSR_PROGRAM_SRCS})
 
   # __HAS_NO_MPI_MOD controls whether mpi.mod is used or if mpi.f is included
   if (MPI_FOUND AND (MPI_Fortran_LIBRARY_VERSION_STRING MATCHES "Intel") AND (NOT ${CMAKE_Fortran_COMPILER_ID} MATCHES "Intel"))
-    target_compile_definitions(dbcsr PRIVATE __HAS_NO_MPI_MOD)
+    target_compile_definitions(${dbcsr_program_name} PRIVATE __HAS_NO_MPI_MOD)
   endif ()
 endforeach ()
 
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -14,13 +14,12 @@ foreach (dbcsr_program_src ${DBCSR_PROGRAM_SRCS})
   # with the Intel compiler CMake 3.12 seems to forget that the source is actually Fortran
   # and needs to be told explicitly:
   set_target_properties(${dbcsr_program_name} PROPERTIES LINKER_LANGUAGE Fortran)
-endforeach ()
 
-# __HAS_NO_MPI_MOD controls whether mpi.mod is used or if mpi.f is included
-if (MPI_FOUND AND (MPI_Fortran_LIBRARY_VERSION_STRING MATCHES "Intel") AND (NOT ${CMAKE_Fortran_COMPILER_ID} MATCHES "Intel"))
-  STRING(REGEX REPLACE "-std=.[^ ]*" "" CMAKE_Fortran_FLAGS ${CMAKE_Fortran_FLAGS})  # mpi.f is not strict f2008
-  target_compile_definitions(${dbcsr_program_name} PRIVATE __HAS_NO_MPI_MOD)
-endif ()
+  # __HAS_NO_MPI_MOD controls whether mpi.mod is used or if mpi.f is included
+  if (MPI_FOUND AND (MPI_Fortran_LIBRARY_VERSION_STRING MATCHES "Intel") AND (NOT ${CMAKE_Fortran_COMPILER_ID} MATCHES "Intel"))
+    target_compile_definitions(dbcsr PRIVATE __HAS_NO_MPI_MOD)
+  endif ()
+endforeach ()
 
 if (WITH_C_API)
   add_executable(dbcsr_example_3_cpp dbcsr_example_3.cpp)
--- a/examples/dbcsr_example_1.F
+++ b/examples/dbcsr_example_1.F
@@ -11,7 +11,9 @@ PROGRAM dbcsr_example_1
    !! DBCSR example 1
    !! This example shows how to create a dbcsr matrix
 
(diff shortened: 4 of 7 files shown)


hfp:
No idea what failed for continuous-integration/CSCS/Daint/PR.

CMakeLists.txt
```diff
@@ -101,6 +102,12 @@ foreach(line ${VERSION_INFO})
   endif ()
 endforeach()
 
+# must be global since modified FLAGS apply to other subdirectories as well, e.g., examples
+# must be after include(CompilerConfiguration) due to modifying CMAKE_Fortran_FLAGS
```
dev-zero: I would still prefer this in `src/CMakeLists.txt` as a `target_compile_definitions`.
The examples currently don't seem to need it and if they do then we should rather duplicate it to give the user an example on how to do it themselves (also the example's CMakeLists.txt is an example)

dev-zero:
> No idea what failed for continuous-integration/CSCS/Daint/PR.

CSCS updated the PrgEnv, have to investigate, you can ignore it for now

alazzaro:
So, the new PrgEnv (6.0.5) uses GCC 8.3 as default, so it cannot compile with other libraries (LIBXSMM). As a fast solution, I'm using 7.3.0.
 I've just updated the dashboard and found the same issue...

CMakeLists.txt
hfp: ACK. Will do.

dev-zero:
ok, there are errors coming from this PR:

```
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.2") 

CMake Error: TRY_RUN() invoked in cross-compiling mode, please set the following cache variables appropriately:

   MPI_RUN_RESULT_C_libver_mpi_normal (advanced)

   MPI_RUN_RESULT_C_libver_mpi_normal__TRYRUN_OUTPUT (advanced)

For details see /scratch/snx3000/jenkg90/jenkins-g90-DBCSR-228.gnu/TryRunResults.cmake

-- Found MPI_C: /opt/cray/pe/craype/2.6.1/bin/cc (found version "3.1") 

CMake Error: TRY_RUN() invoked in cross-compiling mode, please set the following cache variables appropriately:

   MPI_RUN_RESULT_CXX_libver_mpi_normal (advanced)

   MPI_RUN_RESULT_CXX_libver_mpi_normal__TRYRUN_OUTPUT (advanced)

For details see /scratch/snx3000/jenkg90/jenkins-g90-DBCSR-228.gnu/TryRunResults.cmake

-- Found MPI_CXX: /opt/cray/pe/craype/2.6.1/bin/CC (found version "3.1") 

CMake Error: TRY_RUN() invoked in cross-compiling mode, please set the following cache variables appropriately:

   MPI_RUN_RESULT_Fortran_libver_mpi_F90_MODULE (advanced)

   MPI_RUN_RESULT_Fortran_libver_mpi_F90_MODULE__TRYRUN_OUTPUT (advanced)

For details see /scratch/snx3000/jenkg90/jenkins-g90-DBCSR-228.gnu/TryRunResults.cmake
```

the way we run CMake: https://github.com/cp2k/dbcsr/blob/develop/.ci/daint.cscs.ch/gnu.build.sh

... and now after the PrgEnv it is CMake 3.10.x (previously they provided a module for CMake 3.12 which is now unavailable, but the system-cmake was updated)

alazzaro:
Should we install our cmake somewhere on the system?

dev-zero:
> Should we install our cmake somewhere on the system?

That's not the issue, we're officially still supporting 3.10 and should work.

dev-zero:
I think the issue with the PR here is that for pulling the MPI version CMake builds and RUNs a test, but on Cray you're cross-compiling. So we have to either tell CMake somehow that it is fine to also run the executable or we have to provide the result of the run.

alazzaro:
I've never seen this problem before, I would rather open a ticket with CSCS...

alazzaro:
BTW, just noticed that cmake is now CMake:

```
module avai CMake

----- /apps/dom/UES/jenkins/7.0.UP01/gpu/easybuild/modules/all -----
CMake/3.14.5
```

You need to load daint-gpu or -mc first.

dev-zero:
@hfp sorry for the need to rebase
But I've looked into the build failure on Daint some more: current `develop` works fine, so it is really this PR.
From what I've gathered is that since you're requesting the MPI version, the corresponding CMake module uses `try_run` to build and run an executable to get the version from the library via a function call.
Building for Cray (using the `-DCMAKE_SYSTEM_NAME=CrayLinuxEnvironment`) automatically turns on cross-compilation, so https://cmake.org/cmake/help/latest/command/try_run.html#behavior-when-cross-compiling applies. The easiest solution since we're building on a full Daint node (but a login node works just for determining MPI as well) is adding the following to the cmake call:

    -DCMAKE_CROSSCOMPILING_EMULATOR=""

(setting the emulator to passthrough).

dev-zero:
for the flag filtering: you should be able to use `set_source_files_properties` to set flags on a per-flag basis, see the `__SHORT_FILE__` "hack". If it is F2003 compatible, there's the option to simply add `-std=f2003` only for that file then.

A completely different way would be to build the `mpi.mod` from the `mpi.f` in our build system instead (with whatever flags necessary).

hfp:
> for the flag filtering: you should be able to use `set_source_files_properties` to set flags on a per-flag basis, see the `__SHORT_FILE__` "hack". If it is F2003 compatible, there's the option to simply add `-std=f2003` only for that file then.

I need to remove the flag, which should probably receive the currently applied flags. I tried to do this with properties, like receiving properties on a per-file basis (COMPILE_OPTIONS), removing the flags, and reapplying it. That already fails with the first step (receiving the COMPILE_OPTIONS-property). Though, other properties work just fine, e.g., LOCATION. Anyhow, I burned too much time with CMake already.

> A completely different way would be to build the `mpi.mod` from the `mpi.f` in our build system instead (with whatever flags necessary).

I will report Intel `mpi.f` as non-compliant with strict language settings. I have to verify first that this `mpi.f` is referring to Intel MPI. The other option, i.e., compiling the interface is possible but then we still support __HAS_NO_MPI_MOD without implementing it in all necessary places...

hfp:
Still failing...

hfp:
@dev-zero Thank you!  I will resolve the conflict when the test passes.

dev-zero:
I've pushed a commit which should make it build on Daint, let's see.

Anyway, about the `mpif.h`: this seems to be legacy-only nowadays, which means: F77 (maybe F2003) compatible. Dropping the `-std=f2008` from everything is definitely not the way to go here (resulting in `gnu`, which is `f95` plus extensions).

But when checking my `2019.5.281` installation of Intel's Parallel Studio I have this:

```
$ find impi/2019.5.281/ -iname "*.mod"
impi/2019.5.281/intel64/include/mpi_constants.mod
impi/2019.5.281/intel64/include/gfortran/8.2.0/mpi_constants.mod
impi/2019.5.281/intel64/include/gfortran/8.2.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/gfortran/8.2.0/mpi_base.mod
impi/2019.5.281/intel64/include/gfortran/8.2.0/mpi.mod
impi/2019.5.281/intel64/include/gfortran/6.1.0/mpi_constants.mod
impi/2019.5.281/intel64/include/gfortran/6.1.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/gfortran/6.1.0/mpi_base.mod
impi/2019.5.281/intel64/include/gfortran/6.1.0/mpi.mod
impi/2019.5.281/intel64/include/gfortran/4.8.0/mpi_constants.mod
impi/2019.5.281/intel64/include/gfortran/4.8.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/gfortran/4.8.0/mpi_base.mod
impi/2019.5.281/intel64/include/gfortran/4.8.0/mpi.mod
impi/2019.5.281/intel64/include/gfortran/5.1.0/mpi_constants.mod
impi/2019.5.281/intel64/include/gfortran/5.1.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/gfortran/5.1.0/mpi_base.mod
impi/2019.5.281/intel64/include/gfortran/5.1.0/mpi.mod
impi/2019.5.281/intel64/include/gfortran/4.9.0/mpi_constants.mod
impi/2019.5.281/intel64/include/gfortran/4.9.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/gfortran/4.9.0/mpi_base.mod
impi/2019.5.281/intel64/include/gfortran/4.9.0/mpi.mod
impi/2019.5.281/intel64/include/gfortran/9.1.0/mpi_constants.mod
impi/2019.5.281/intel64/include/gfortran/9.1.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/gfortran/9.1.0/mpi_base.mod
impi/2019.5.281/intel64/include/gfortran/9.1.0/mpi.mod
impi/2019.5.281/intel64/include/gfortran/7.1.0/mpi_constants.mod
impi/2019.5.281/intel64/include/gfortran/7.1.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/gfortran/7.1.0/mpi_base.mod
impi/2019.5.281/intel64/include/gfortran/7.1.0/mpi.mod
impi/2019.5.281/intel64/include/mpi_f08_types.mod
impi/2019.5.281/intel64/include/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/mpi_base.mod
impi/2019.5.281/intel64/include/pmpi_f08.mod
impi/2019.5.281/intel64/include/mpi_f08.mod
impi/2019.5.281/intel64/include/mpi_f08_link_constants.mod
impi/2019.5.281/intel64/include/ilp64/mpi_constants.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/8.2.0/mpi_constants.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/8.2.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/8.2.0/mpi_base.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/8.2.0/mpi.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/6.1.0/mpi_constants.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/6.1.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/6.1.0/mpi_base.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/6.1.0/mpi.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/4.8.0/mpi_constants.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/4.8.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/4.8.0/mpi_base.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/4.8.0/mpi.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/5.1.0/mpi_constants.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/5.1.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/5.1.0/mpi_base.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/5.1.0/mpi.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/4.9.0/mpi_constants.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/4.9.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/4.9.0/mpi_base.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/4.9.0/mpi.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/9.1.0/mpi_constants.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/9.1.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/9.1.0/mpi_base.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/9.1.0/mpi.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/7.1.0/mpi_constants.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/7.1.0/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/7.1.0/mpi_base.mod
impi/2019.5.281/intel64/include/ilp64/gfortran/7.1.0/mpi.mod
impi/2019.5.281/intel64/include/ilp64/mpi_sizeofs.mod
impi/2019.5.281/intel64/include/ilp64/mpi_base.mod
impi/2019.5.281/intel64/include/ilp64/mpi.mod
impi/2019.5.281/intel64/include/mpi_f08_compile_constants.mod
impi/2019.5.281/intel64/include/mpi.mod
impi/2019.5.281/intel64/include/mpi_f08_callbacks.mod
```

So, I'd rather give users pointers on setting up their environment correctly and only use `mpi.mod` from now on. From what I have on my system, the following is sufficient to build with Intel MPI and GNU when the Intel environment was loaded with `source bin/compilervars.sh intel64`:

```
CC=mpicc FC=mpifc CXX=mpicxx cmake ~/work/dbcsr/dbcsr/
```

see
```console
$ ldd tests/dbcsr_unittest1 
	linux-vdso.so.1 (0x00007ffc6d8ec000)
	libgomp.so.1 => /usr/lib64/libgomp.so.1 (0x00007f6f04fcd000)
	libopenblas_openmp.so.0 => /usr/lib64/libopenblas_openmp.so.0 (0x00007f6f02a26000)
	libmpifort.so.12 => /data/tiziano/intel-bin/compilers_and_libraries_2019.5.281/linux/mpi/intel64/lib/libmpifort.so.12 (0x00007f6f02668000)
	libmpi.so.12 => /data/tiziano/intel-bin/compilers_and_libraries_2019.5.281/linux/mpi/intel64/lib/release/libmpi.so.12 (0x00007f6f01663000)
	librt.so.1 => /lib64/librt.so.1 (0x00007f6f0145b000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f6f0123d000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f6f01039000)
	libgfortran.so.4 => /usr/lib64/libgfortran.so.4 (0x00007f6f00c65000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f6f0092d000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f6f00715000)
	libquadmath.so.0 => /usr/lib64/libquadmath.so.0 (0x00007f6f004d5000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f6f0011b000)
	libfabric.so.1 => /data/tiziano/intel-bin/compilers_and_libraries_2019.5.281/linux/mpi/intel64/libfabric/lib/libfabric.so.1 (0x00007f6effee3000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f6f051fd000)
```

hfp:
I could be also something wrong with the paths setup by the FindMPI.cmake module. I will check asap with `make VERBOSE=1` to see which paths are actually referred wrt module search.

hfp:
(Tests passed with your changes, hence I've updated the PR)

hfp:
Updated PR fails again; not clear why.

alazzaro:
About this PR, now that v2.0.0 is out...

My understanding is that 

#include "mpif.h"

was legacy-related to <f2008. 
Actually, the standard (MPI 3.1) says:

> The use of the mpif.h include file is strongly discouraged and may be deprecated in a future version of MPI.

https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node411.htm
 
Actually, the macro `__HAS_NO_MPI_MOD` is there just because of CP2K compatibility, otherwise it should be removed (and indeed it is not documented in Makefile.inc). CP2K Install.md reports:

> -D__HAS_NO_MPI_MOD` - workaround if mpi has been built for a different (version of the) Fortran compiler, rendering the MPI module unreadable (reverts to f77 style mpif.h includes)

Now, I understand the possibility to use cross-compiling between different compilers (e.g. non-Intel) and MPI distribution (Intel MPI). Is this really necessary? I mean, does Intel MPI provides already a module GNU-compatible? It seems CINECA had a similar problem:

hfp:
I think the problem is that MOD-files are compiler-specific and may vary on a per-version basis. Given this mess, a solution seems to be generating the MOD-file for the compiler used (on the fly). However, I believe this is asking again to rely on mpif.h...

Another option might be to use an implicit interface i.e., no interface. This could seal everything in CP2K's or DBCSR's MPI-wrapper. On the other hand, this does not solve the problem for user's code or even the examples...

dev-zero:
> I think the problem is that MOD-files are compiler-specific and may vary on a per-version basis. Given this mess, a solution seems to be generating the MOD-file for the compiler used (on the fly). However, I believe this is asking again to rely on mpif.h...

As listed above, Intel Parallel Studio provides `.mod` files for various versions of `gfortran`, and guessing from the version selection, someone looked into the question which versions of `gfortran` generate compatible versions of the MOD file.
Furthermore, as I've illustrated above do the `mpi*` wrappers provided in the Intel Parallel Studio automatically use the appropriate include directory. 
As I've shown above, I was able to successfully build with the current version of our CMake scripts DBCSR with gfortran but linking against Intel MPI.

So, what exactly is missing here?
Dropping `mpif.h` is also possible in CP2K according to Jürg.

alazzaro:
> I think the problem is that MOD-files are compiler-specific and may vary on a per-version basis. Given this mess, a solution seems to be generating the MOD-file for the compiler used (on the fly). However, I believe this is asking again to rely on mpif.h...

Yes, I think so. However, I don't think the DBCSR library should take care of this problem...

> 
> Another option might be to use an implicit interface i.e., no interface. This could seal everything in CP2K's or DBCSR's MPI-wrapper. On the other hand, this does not solve the problem for user's code or even the examples...

I don't like this idea either... Actually, my idea is to move to f08 API (in a very long long timescale)... Still, I don't think we should care about this problem from the library's point of view...

Now, if we really want to find a solution, what about providing a series of wrappers from the C MPI API? In this way, we don't need to rely on the MPI module and we will build our own from the C header... I think OpenMPI does already something similar...

hfp:
@alazzaro I think anything that solves this issue would be very useful. I believe Intel MPI and GNU Compiler is a very common installation.

alazzaro:
> @alazzaro I think anything that solves this issue would be very useful. I believe Intel MPI and GNU Compiler is a very common installation.

Agree.
I think a way to solve the problem is really to use a C interface, so that we don't rely on any Fortran module, actually we build our own based on C functions... I don't like the idea to use a F77 interface (as we are doing it now), but we cannot use the F08 either, unless we want to support 2 interface (our own + F08)... 

In any case, this is not a solution we can solve soon. I propose to turn this PR in an issue and leave it there as a placeholder (MPI wrappers refactoring)

hfp:
Ok, sounds good! I will turn this PR somehow into an issue.

I will also not reintroduce it from upcoming PRs on my end. Hint, my OpenMP based offload/acc backend is nearly ready to be PR'd...

PR #257, opened 2019-11-13
drop support for mpif.h

dev-zero:


alazzaro:
Fantastic, I like the Install file...

dev-zero:
thanks, did that testing anyway, could as well put it in shape and document it somewhere properly

Issue #214, opened 2019-08-02
OpenMP/offload backend


hfp:
OpenMP 4.0 introduced offloading to attached devices and made extensions in OpenMP 4.5 (future versions of OpenMP will likely only evolve the offload capability with more advanced features on top of a mature baseline). It can be considered to implement an OpenMP-based backend which is able to offload to GPUs (and potentially other PCI-attached devices). This would be applicable to Fortran and C/C++ (even for the code within the offload-region). However since the ACC-interface is already C-based, an even wider set of working toolchains is already working: LLVM/Clang w/ NV offload plugin can be built from source (tested on x86 w/ NVidia GPU) and IBM XL compilers support OpenMP 4.5 and work right away for NVidia GPUs.

hfp:
https://github.com/hfp/dbcsr/tree/develop/src/acc/openmp (work in progress)

hfp:
PR #260 may be reviewed. Limitations are left in the PR's decription. Limitations are worked out preferably using subsequent PRs.

alazzaro:
Closing here, I don't think we will ever use OpenMP-offload at this point...

PR #258, opened 2019-11-17
Adjusted makedep to drop self-dependencies

hfp:
A "circular dependency" such as "module -> module" is dropped. The latter can appear for PACKAGE files that contribute to the same archive (libdbcsr), which is just fine.

The change will be also proposed for CP2K's copy of makedep.

### DIFF
--- a/tools/build_utils/makedep.py
+++ b/tools/build_utils/makedep.py
@@ -353,8 +353,14 @@ def find_cycles(parsed_files, mod2fn, fn, src_dir, S=None):
     for m in pf["module"]:
         if m in S:
             i = S.index(m)
-            error("Circular dependency: " + " -> ".join(S[i:] + [m]))
-        S.append(m)
+            n = S[i:]
+        else:
+            n = []
+
+        if 0 == len(n) or (1 == len(n) and m in n):
+            S.append(m)
+        else:
+            error("Circular dependency: " + " -> ".join(n + [m]))
 
     for m in collect_use_deps(parsed_files, fn, src_dir):
         if m in mod2fn.keys():
@@ -380,8 +386,14 @@ def collect_pkg_deps(packages, p, archives=None, S=None):
 
     if a in S:
         i = S.index(a)
-        error("Circular package dependency: " + " -> ".join(S[i:] + [a]))
-    S.append(a)
+        b = S[i:]
+    else:
+        b = []
+
+    if 0 == len(b) or (1 == len(b) and a in b):
+        S.append(a)
+    else:
+        error("Circular package dependency: " + " -> ".join(b + [a]))
 
     for r in packages[p]["requires"]:
         d = normpath(path.join(p, r))


oschuett:
This seems to be a special requirement for DBCSR. I'd propose we instead wait for https://github.com/cp2k/cp2k/issues/351.

hfp:
Sure, I will hold-back any related PR for CP2K!

This PR enables upcoming PRs to be tested as early as possible (upcoming draft PR for #214). I will probably merge this one asap (in oder to get upcoming/depending PRs into a green state).

oschuett:
The one-to-one correspondence between PACKAGEs and archives is basic assumption of the CP2K build system. We rely on this e.g. in [check_archives.py](https://github.com/cp2k/dbcsr/blob/develop/tools/build_utils/check_archives.py) and there might be other places.

If this constraint is holding you back then we should rather revisit our decision to wait for Spack. Moving DBCSR into the toolchain is actually quite simple and I could probably do it within day.

hfp:
Perhaps it's a good idea to settle the build system (not just for this PR)!

I have nothing against GNU make ("wysiwyg") or the old way things are currently done based on install scripts (but I am happy to adhere with whatever is decided). I am also fine with Spack if it is what people want and use. Though, I have some "friendship" with CMake (I think it is a blackbox [if not magic]; sort of a productivity-hog which can solve problems that are rare on mostly Unix-like targets).

hfp:
> The one-to-one correspondence between PACKAGEs and archives is a basic assumption of the CP2K build system.

I think this is not touched by this change. It only seems to solve an issue that was not reproduced so far. Though, it may somehow depend on the way of traversing the project (`src` directory). The problem solved was a "circular dependency" printed as `"libdbcsr -> libdbcsr"`. However, I did nothing new compared to other subfolders listing `"archive": "libdbcsr"`. Honestly, the issue did not reproduce under native Linux, but in my Windows Subsystem for Linux box (WSL). I wonder if the issue is of general nature but not exposed (yet).

This change does not allow circular dependencies, but only drops a (false/self) dependency discovered for a PACKAGE if this dependency points out the same archive as the PACKAGE contributes to.

alazzaro:
@hfp 
I did change the makedep script to make a single module. @oschuett is right, the original (which is the current in CP2K) does a module per each directory.
In the current DBCSR design, we don't need this check, I agree. However, DBCSR will be modularized too. For instance, we will have a Tensor module, with a library

`libdcbsr_tensor`

Therefore, the check has to stay there for across "modules".
There is an entry to specify the archive in the PACKAGE files:

`archive": "libdbcsr",`

Finally, cmake will remove all this stuff...

alazzaro:
> This change does not allow circular dependencies, but only drops a (false/self) dependency discovered for a PACKAGE if this dependency points out the same archive as the PACKAGE contributes to.

Just saw your comment, this is exactly what I meant in my comment...

hfp:
@alazzaro shall I leave this PR open?

No worries, I think the problem does not reproduce on systems we use/test on (so it does not block me). I only discovered the issue with WSL for whatever reason (and found the change somewhat reasonable).

alazzaro:
If it does what you said (i.e. checking for circular dependencies only when archive is different), then it is fine for me to merge... Sorry, I don't have time to check it...

PR #260, opened 2019-11-17
OpenMP/offload backend

hfp:
* Updated ACC and related interfaces for improved type-safety.
* Existing code (CUDA) may be updated to rely on new type definitions.

Issue #214
* Introduced OpenMP/offload backend, which implements the stream programming model.
* Though, the stream programming is not strictly necessary in order to use accelerator devices.
* However, streams are used and baked by DBCSR's ACC interface.
* OpenMP/offload interface brings few functions on top of current ACC interface.
* OpenMP/offload could be used without ACC interface and allows mixed C/Fortran impl.
* Introduced tests/dbcsr_test_acc.c testing any of the existing backends (cuda, hip, openmp).
* Certain generic preconditions, post-conditions, and thread-safety are checked.


hfp:
This draft has mostly everything in it already (no large code additions needed). However, the following is known to be missed (and will be contributed asap):

* Well performing SMMs and batch transpose kernels. For Intel, this is planned to be based on Intel MKL library calls. However, separate (naive) kernels may be given. In fact, the latter is already true for batch-transpose (naive C based implementation). If the MKL road does not workout (it should), there will be plenty of flexibility to get an implementation. Even things like leveraging MATMUL or TRANSPOSE intrinics (Fortran) can be used to get at least a working implementation.
* The CMake based build integration is already functioning (XLC). There some flexibility `WITH_OFFLOAD_FLAGS` and `WITH_OPENMP_FLAGS` on top of `USE_OFFLOAD`, which allows to accomodate a moving target (Intel toolchain for GPU). The Makefile based build system may be adjusted accordingly, but it is already good for building all bits plus new ACC-test.

hfp:
This work may depend on #258.

hfp:
A few words about the implementation of the OpenMP/offload backend: [https://github.com/hfp/dbcsr/tree/openmp_acc/src/acc/openmp#openmpoffload-backend](https://github.com/hfp/dbcsr/tree/openmp_acc/src/acc/openmp#openmpoffload-backend).

hfp:
I need some help regarding the Daint log. The build fails on Daint-CI but passes on Travis (it works for me on Archer wrt CCE).

alazzaro:
It seems that problem is due to the fact that we are building the CCE version with CUDA:

```
d /scratch/snx3000/jenkg90/jenkins-g90-DBCSR-264.cray/src/acc/libsmm_acc/libcusmm && /opt/nvidia/cudatoolkit10/10.1.105_3.27-7.0.1.1_4.1__ga311ce7/bin/nvcc  -DARCH_NUMBER=60 -I/scratch/snx3000/jenkg90/jenkins-g90-DBCSR-264.cray/src/acc/libsmm_acc/libcusmm -I/users/jenkg90/workspace/g90/DBCSR/src/acc/libsmm_acc/libcusmm  -arch=sm_60 --cudart static -O3 -DNDEBUG   -std=c++11 -x cu -c /users/jenkg90/workspace/g90/DBCSR/src/acc/libsmm_acc/libcusmm/libcusmm_benchmark.cu -o CMakeFiles/libcusmm.dir/libcusmm_benchmark.cu.o

/users/jenkg90/workspace/g90/DBCSR/src/acc/libsmm_acc/libcusmm/libcusmm.cpp:245:16: error: conflicting types for 'libsmm_acc_process'

extern "C" int libsmm_acc_process (void *param_stack, int stack_size, int nparams, int datatype, void *a_data, void *b_data, void *c_data, int m, int n, int k, int def_mnk, void *stream){

               ^

/users/jenkg90/workspace/g90/DBCSR/src/acc/libsmm_acc/libcusmm/../include/libsmm_acc.h:29:5: note: previous declaration is here

int libsmm_acc_process(const libsmm_acc_stack_descriptor_type* dev_param_stack, int stack_size,

    ^

/users/jenkg90/workspace/g90/DBCSR/src/acc/libsmm_acc/libcusmm/libcusmm.cpp:340:16: error: conflicting types for 'libsmm_acc_transpose'

extern "C" int libsmm_acc_transpose (void *trs_stack, int offset, int nblks, void *buffer,int datatype, int m, int n, void* stream) {

               ^

/users/jenkg90/workspace/g90/DBCSR/src/acc/libsmm_acc/libcusmm/../include/libsmm_acc.h:26:5: note: previous declaration is here

int libsmm_acc_transpose(const int* dev_trs_stack, int offset, int nblks,

    ^

2 errors generated.

make[2]: *** [src/acc/libsmm_acc/libcusmm/CMakeFiles/libcusmm.dir/build.make:79: src/acc/libsmm_acc/libcusmm/CMakeFiles/libcusmm.dir/libcusmm.cpp.o] Error 1

make[2]: *** Waiting for unfinished jobs....

make[2]: Leaving directory '/scratch/snx3000/jenkg90/jenkins-g90-DBCSR-264.cray'

make[1]: *** [CMakeFiles/Makefile2:351: src/acc/libsmm_acc/libcusmm/CMakeFiles/libcusmm.dir/all] Error 2

make[1]: Leaving directory '/scratch/snx3000/jenkg90/jenkins-g90-DBCSR-264.cray'

make: *** [Makefile:141: all] Error 2
```

alazzaro:
For the record, Intel and GNU are failing with the same error...

hfp:
@alazzaro Thank you, that already helps and I know what to do. I will give it another try likely by EOD today.

CMakeLists.txt
```diff
@@ -63,8 +63,9 @@ find_package(LAPACK REQUIRED)
 find_package(PkgConfig)
 
 if (USE_MPI)
-  get_property(REQUIRED_MPI_COMPONENTS GLOBAL PROPERTY ENABLED_LANGUAGES)
-  list(REMOVE_ITEM REQUIRED_MPI_COMPONENTS CUDA)  # CUDA does not have a MPI component
+  # MPI-parallelization only happens with Fortran rather than ENABLED_LANGUAGES
```
dev-zero: negative, for C-API/CXX-API examples we also need the C MPI, resp. the CXX (latter I'm not sure, might be we test always only use the C).
But we could easily depend on Fortran and C statically since all implementations seem to provide C anyway.

hfp: Fortunately the code above is outdated. Sorry for this!  I found MPI is needed for both C and C++ in addition to Fortran.

CMakeLists.txt
```diff
@@ -9,6 +9,9 @@ include(CMakeDependentOption)
 
 option(USE_MPI "Build with MPI support" ON)
 option(USE_OPENMP "Build with OpenMP support" ON)
+cmake_dependent_option(USE_OFFLOAD "Offloading to devices if OpenMP is enabled" OFF "USE_OPENMP" OFF)
+set(WITH_OFFLOAD_FLAGS "" CACHE STRING "Additional flag(s) if OpenMP/offload is enabled")
+set(WITH_OPENMP_FLAGS "" CACHE STRING "To replace OpenMP default flag, e.g., to replace -fopenmp")
```
dev-zero: how does this differ from the possibility to set `OpenMP_Fortran_FLAG`, `OpenMP_C_FLAG`, etc. except that it sets the flag for all languages? See https://github.com/Kitware/CMake/blob/5f630a934aa4b8616aa59988ea70c14065d3a08d/Modules/FindOpenMP.cmake#L94

CMakeLists.txt
```diff
@@ -88,6 +92,39 @@ endif ()
 
 if (USE_OPENMP)
   find_package(OpenMP REQUIRED)
+
+  # necessarily a starting point; ultimately the OpenMP module/package must to be extended
+  if (USE_OFFLOAD AND WITH_OPENMP_FLAGS)
```
dev-zero: If I get the logic correctly, you're bypassing CMake's checks of the given flags here: The user can get a "OpenMP found - everything's fine" from CMake, but if he passes invalid flags the compilation fails later on.

At least for the `WITH_OPENMP_FLAGS` it should be possible to set `OpenMP_${LANG}_FLAG` prior to the `find_package`, if we can't tell the user to directly set `OpenMP_${LANG}_FLAG` (which I'd prefer rather than introducing another custom flag).

hfp: Both flags serve a temporary purpose:

* WITH_OFFLOAD_FLAGS: is meant to add an additional flag to the list of OpenMP flags, e.g., `-qoffload`. This is not supported yet by the CMake OpenMP module.
* WITH_OPENMP_FLAGS: is meant to replace OpenMP flags regardless of what was figured out by the OpenMP CMake-module. This serves for instance Intel's OneAPI compiler for the time being. At the moment there special flags needed to trigger the OpenMP based GPU code. 

Anyhow, both of these helper options are nothing to worry about. This can be easily cleaned up, and should not interfere.

CMakeLists.txt
hfp: Thank you, I will lookup this reference!  I expect the CMake part to cleanup along with our evolving compiler, and more/alternative toolchains. So far, the CMake portion of this PR is small and may get even smaller. Even with some of the (obviously) temporary additions, I expect no interference with the way things are working and being tested.

hfp:
@dev-zero Thank you for the review so far!  I have to incorporate the hip changes now before any next step.

hfp:
DBCSR/ACC now refactored common code for CUDA and HIP backends under `src/acc`. This flaws the idea of having separated backends. I need to find a solution for this.

alazzaro:
> DBCSR/ACC now refactored common code for CUDA and HIP backends under `src/acc`. This flaws the idea of having separated backends. I need to find a solution for this.

This is what is worrying me too...
Acc will have 3 backends, and probably we will add a new one in the future (oneAPI, Sycl?)...
We need some refactoring there. I will try to take a look in the next days...

hfp:
> 
> 
> > DBCSR/ACC now refactored common code for CUDA and HIP backends under `src/acc`. This flaws the idea of having separated backends. I need to find a solution for this.
> 
> This is what is worrying me too...
> Acc will have 3 backends, and probably we will add a new one in the future (oneAPI, Sycl?)...
> We need some refactoring there. I will try to take a look in the next days...

Let's not worry. I will make an intermediate step and I am half-way through. Although, the code under `src/acc` is currently sort of backend-dependent, it's furtunately still separate at build-level.

What I am doing:

* Moved libsmm_acc/include/* -> acc/include. Now, `acc.h` and `acc_libsmm.h` (renamed from libsmm_acc.h) are side-by-side under `acc/include`.
* Created `libsmm_omp` side-by-side with `libsmm_acc`.

Both changes cleaned up some cumbersome/deep nesting (minor). At the same time (more important), `libsmm_acc` (which is was made effectively specific for CUDA and HIP because of sharing common at that level), is now again separate. With this, we can again cleanly add "SMM backends" and ACC backend".

hfp:
> What I am doing:
> * Moved libsmm_acc/include/* -> acc/include. Now, `acc.h` and `acc_libsmm.h` (renamed from libsmm_acc.h) are side-by-side under `acc/include`.
> * Created `libsmm_omp` side-by-side with `libsmm_acc`.

I will send a separate PR (if I can can weed it out ;-).

hfp:
PR #267.

hfp:
This PR will be rebased on top of PR #267.

hfp:
This PR passes now in CI on both Travis and Daint. For the latter, dbcsr_test_acc (the new interface test applicable to all backends) did not pass. I simply disabled it for now on Cray systems (not CCE, but entire Cray). Note, it passes with CUDA on my test system but it failed on Daint with some weird MPI-related system message. It needs some debug to effectively enable it on Daint, I will get my access refreshed. I think the PR can be taken (or reviewed a bit more). As described ("limitations"), "only" the actual kernels are missing. I will send PRs on top to bring at least naive code and later add performance for at least the Intel part.

alazzaro:
Thanks Hans!
Gimme some time to review what you did...

hfp:
@alazzaro No rush the whole subject remains WIP anyway and additional PRs are needed on top to actually get things running/performing. I will involve people as things get ready on Intel's side wrt to HW and SW.

There are interesting options possible: mixed language offload, offload without ACC backend (i.e., no stream programming), etc. Regarding stream programming and OpenMP based ACC-backend, I wrote up some intro/motivation: [https://github.com/hfp/dbcsr/tree/openmp_acc/src/acc/omp#openmpoffload-backend](https://github.com/hfp/dbcsr/tree/openmp_acc/src/acc/omp#openmpoffload-backend). Further, I noticed Cray Compiler supports OpenMP/offload (or at least it attempts to do something); can be interesting to try-out. I have [disabled the Cray offload](https://github.com/hfp/dbcsr/blob/openmp_acc/src/acc/omp/dbcsr_omp.h#L116), it is triggered by sole existence of respective OpenMP directives. However, I did not want to bother with it at this time.

hfp:
Reintroduced workaround: [https://github.com/hfp/dbcsr/blob/openmp_acc/tests/CMakeLists.txt#L84](https://github.com/hfp/dbcsr/blob/openmp_acc/tests/CMakeLists.txt#L84)

Output from Daint-CI is:

```text
Tue Nov 26 13:06:54 2019: [unset]:_pmi_alps_init:alps_get_placement_info returned with error -1
Tue Nov 26 13:06:54 2019: [unset]:_pmi_init:_pmi_alps_init returned -1
```

Not sure if the error message appears because of the test is failing or if the test fails because of an issue.

hfp:
One may review this PR and comment. However, please do not merge yet. There is a change needed regarding the barriers used in the code.

hfp:
I think this PR needs more work! The prerequisites for this to work are just too specific: see my [README.md](https://github.com/hfp/dbcsr/tree/openmp_acc/src/acc/omp#openmpoffload-backend). For instance, a parallel region is needed to operate tasks; here it's even harder - we want tasks to keep running (i.e., the same parallel region must stay) but not even all of the relevant ACC code is hosted inside of a parallel region. Moreover if the implementation is based on workshare, all threads of the parallel region must take the same branches as otherwise synchronization with any sort of barrier may be impossible.

I can separate the new ACC test along with some other improvements (independent of any backend). I would take one more turn trying plain threads to implement the streaming model. If this gets too complicated, I may just rely on DPC++/SYCL to implement an offload backend. As I understood, this is desired anyway? Shall we sync on this subject upfront/f2f somehow?

hfp:
Parts of the work will be recycled and next attempt made starting by begin of next month.

Issue #261, opened 2019-11-18
rename acc_init function


hfp:
Rename `acc_init` (ACC interface) as this name conflicts when using Cray Compiler (CCE). It appears OpenACC is enabled by default but has a similar prototype/symbol. The hint about this issue is a relatively unspecific warning message:

```text
A declaration for library function acc_init is invalid.
```

Perhaps `dbcsr_` should used instead or in addition of/to `acc_`. The solution may also impact CP2K.

hfp:
I guess this should be resolved for 2.0?

alazzaro:
Adding `dbcsr_` namespace it is always a good idea (we enforce this policy for the API functions).
Concerning CCE, I think we don't see the problem in the dashboard because we don't use it in combination with the ACC. As far I can tell, we don't use it on the CI with the ACC either.

So, at this point, I think you want to have ACC enabled with CCE? Personally, I would not support it (CCE is used for testing, no for production). The current ACC-GPU support is GNU or Intel+CUDA.

Moving ahead, the rename of the function can be a good idea. Probably it can be related to what @shoshijak is doing with #241 ?

alazzaro:
> I guess this should be resolved for 2.0?

Well, CCE+GPU is not supported, so I would not consider the fix for 2.0... let's keep it for 2.1

hfp:
Makes sense. Alternatively (or in addition), we could figure out the `-h` setting to disable OpenACC when using CCE/Cray. It is definitely a concern wrt GPU support to have a symbol clash like this.

alazzaro:
Well, the flag is

`-hnoacc`

I agree, that's the easiest way, let's keep the issue open for future discussion.
For the time being, could you update the CCE flag in the CI?
I will update the dashboard...

hfp:
ACK, will do!

hfp:
I will update `cmake/CompilerConfiguration.cmake` to include `-hnoacc`. Btw, I am testing this on Archer.

alazzaro:
OK, perfect!
Actually, Archer has a pretty old CCE version (up to 8.7.7). With the CCE 9.x (based on LLVM) some of the flags, such as OpenMP, are off by default... Do we consider such a case, @dev-zero ?

hfp:
Indeed, I use the "latest" 8.7.7.

hfp:
[...]

PR #262, opened 2019-11-18
Issue #261: apply -hnoacc flag

hfp:
I am not completely certain about this PR, but perhaps good enough.

### DIFF
--- a/cmake/CompilerConfiguration.cmake
+++ b/cmake/CompilerConfiguration.cmake
@@ -68,10 +68,9 @@ elseif (CMAKE_CXX_COMPILER_ID STREQUAL "Cray")
   endif ()
   # OpenACC support with CCE is EOL: causes https://github.com/cp2k/dbcsr/issues/261
   # eventually check compiler version (similar to -h system_alloc)
-  # compiler warning remains (-M1234 is not accepted)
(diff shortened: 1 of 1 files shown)


alazzaro:
I'm not sure, but likely it requires a space between -M and the number?
However, I'm not sure the CC-1234 is a warning, so we cannot suppress it...

hfp:
I tried with and without space: "invalid characters after -M". Anyhow, the `cc` binary on Archer (as figured/used by CMake) did not want any such option not even the deployed suppression (`-M3105`). However, I did not check with `ftn`. I am certain that CMake was using v8.8.7 as the path to the binary also included the version number.

alazzaro:
So, the conclusion is: the `-hnoacc` is good to have, but it doesn't solve the problem. Right?
The only solution is to rename the function... This PR can be merged anyway...

hfp:
I can confirm, `-M` flag is Fortran specific (`ftn`). I think, this can be worked out for C/C++ as well: [https://pubs.cray.com/content/S-2179/8.7/cray-c-and-c++-reference-manual/message-options](https://pubs.cray.com/content/S-2179/8.7/cray-c-and-c++-reference-manual/message-options). Let me try this and updated this PR.

PR #267, opened 2019-11-22
Separate implementations for ACC and ACC_LIBSMM

hfp:
Regain separate implementations for ACC ("cuda", "hip", later "openmp") and ACC_LIBSMM ("libsmm_acc", later "libsmm_omp"). This change also includes the updated interfaces (header files) for ACC and ACC_LIBSMM both implementing improved type-safety. Both, CUDA and HIP backends are updated accordingly.

In addition, the CUDA/HIP backend got some minor adjustment to pass with the new/upcoming ACC/interface test (`acc_stream_destroy`, `acc_event_destroy`). Also, some issue compiling with CUDA on POWER received an update (disabled "issue seems to be present even with CMake 3.14.5 in some environments" as it is present with newer CMake).

PR #268, opened 2019-11-22
ci/daint: add build-only step for GNU with CMake 3.10

dev-zero:


shoshijak:
Thanks a lot for this!

hfp:
Ok, Travis fails in the same way as for #267  (I hope it's not the end of Python 2.x ;-)

dev-zero:
@hfp yes. I just cleared all the caches for the builds in case we see an issue that Travis updated the macOS image and now we have a discrepancy between the cached homebrew and the base image.

shoshijak:
@dev-zero thank you for the fix! Sorry I didn't get to it yet ...

dev-zero:
@shoshijak np, just thought I'd give it a shot, obviously it's not exactly the same as the other fix ;-)

shoshijak:
@dev-zero this is annoying. I've looked into it, it seems your solution doesn't work because `CUBLAS` is a variable (set through `find_library(CUBLAS ...)`), not a target. 

I've come up with these two workarounds, both working, but not entirely satisfactory: 

1. Set `-L ... libclublas.so` as a compile flag for the target `acc`. Has the disadvantage of adding `libcublas` at all compilation stages of `acc`, not only for linking.

```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 8f71b2fe..7fde584c 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -247,7 +247,9 @@ if (USE_CUDA)
   target_sources(dbcsr PRIVATE $<TARGET_OBJECTS:acc>)
   if (USE_CUBLAS)
     target_compile_definitions(acc PRIVATE __DBCSR_ACC=2)
-    target_link_libraries(acc PRIVATE ${CUBLAS})
+    # with CMake 3.12+ the following can be replaced by:
+    #   target_link_libraries(acc PRIVATE ${CUBLAS})
+    target_compile_options(acc PRIVATE -L ${CUBLAS})
   else ()
     target_compile_definitions(acc PRIVATE __DBCSR_ACC)
   endif ()
```

2. A little better than (1), since it uses CUBLAS only for linking, not compiling, as it should. But uses cmake's archaic `set_target_properties`.

```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 8f71b2fe..50e132cd 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -247,7 +247,9 @@ if (USE_CUDA)
   target_sources(dbcsr PRIVATE $<TARGET_OBJECTS:acc>)
   if (USE_CUBLAS)
     target_compile_definitions(acc PRIVATE __DBCSR_ACC=2)
-    target_link_libraries(acc PRIVATE ${CUBLAS})
+    # with CMake 3.12+ the following can be replaced by:
+    #   target_link_libraries(acc PRIVATE ${CUBLAS})
+    set_target_properties(acc PROPERTIES INTERFACE_LINK_LIBRARIES CUBLAS)
   else ()
     target_compile_definitions(acc PRIVATE __DBCSR_ACC)
   endif ()
```

dev-zero:
@shoshijak thanks, I went with the latter, using archaic stuff to fix archaic versions seems like a good compromise ;-)

shoshijak:
makes sense :)

Issue #95, opened 2018-11-19
Include new prettify


alazzaro:
New prettify from https://github.com/pseewald/fprettify as submodule
Unfortunately, the new prettify "forces" make to always run the target since it always updates the file, even there were no changes!

alazzaro:
Close with [972f09e](https://github.com/cp2k/dbcsr/commit/972f09eb17c021c4970ddbd08596869b43e88a33)

alazzaro:
Sure, of course. My first step was to include it...

alazzaro:
@pseewald @dev-zero What's the status of this issue?
My trick was to do this in the Makefile:

```
# Pretty function, check if the file requires update
define pretty_func
        @mkdir -p $(PRETTYOBJDIR)
        @rm -f $2
        $(TOOLSRC)/fprettify/fprettify.py --disable-whitespace -s $1 > $2
        @cmp -s $1 $2; \
        RETVAL=$$?; \
        if [ $$RETVAL -ne 0 ]; then \
            cp $2 $1; \
        fi
endef
```

but I think we can now remove it with the new prettifier, right?

pseewald:
I still need to fix pseewald/fprettify#34

alazzaro:
OK, then I will move this issue to v3. No rush.

pseewald:
> I still need to fix [pseewald/fprettify#34](https://github.com/pseewald/fprettify/issues/34)

This is fixed

alazzaro:
closed via #270

PR #270, opened 2019-11-25
Update fprettify to current master

pseewald:
This will fix formatting issue with FORD documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -365,16 +365,10 @@ prettyclean:
 	-rm -rf $(PRETTYOBJDIR)
 TOOL_HELP += "prettyclean : Remove prettify marker files"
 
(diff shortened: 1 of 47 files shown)


alazzaro:
I think, this PR also includes the #95 issue, right?

pseewald:
in fprettify this was fixed, but the DBCSR makefile still has the workaround - thanks for mentioning, I'll include it in this PR.

alazzaro:
LGTM, I wonder if we can have pretty as part of the pre-commit stuff and eventually cmake... 
As far I can tell, for the moment it is only part of the "deprecated" makefile...

@dev-zero any ^^

dev-zero:
@alazzaro cmake no, pre-commit yes

Issue #248, opened 2019-11-07
c interface for tensors


ambmax00:
Hi again,
Are there any plans for making a c interface for the tensor subroutines in DBCSR?
If not, I would be glad to help out on that.

alazzaro:
Thanks for your collaboration!

Currently, there are no plans to improve the C interface. 
The current C interface for the main DBCSR functions is still a prototype. The reason is that, before any C API, we have to improve the Fortran API. Then, ideally, the C API should be automatically generated from Fortran...

We have a quite strict schedule for DBCSR, soon we will release v2.0 (requested by CP2K). Therefore, any change to the API should wait for v3.0, reasonable before mid next year.

However, however, if you want to can start to prototype with the current C design.
As you can see, there are two files to consider:
1) [dbcsr_api_c.F](https://github.com/cp2k/dbcsr/blob/develop/src/dbcsr_api_c.F)
2) [dbcsr.h](https://github.com/cp2k/dbcsr/blob/develop/src/dbcsr.h)

Note that the entry point is [dbcsr_api.F](https://github.com/cp2k/dbcsr/blob/develop/src/dbcsr_api.F).

I propose to add two new files for the tensors (with _tensors). The entry point is https://github.com/cp2k/dbcsr/blob/develop/src/tensors/dbcsr_tensor_api.F

ambmax00:
Great, I will start working on it then. I will let you know when I have a first prototype.
Thanks!

pseewald:
A C interface to tensors is very welcome from my side. Since many of the API routines are written using [Fypp](https://github.com/aradi/fypp)-preprocessor macros, I suggest an automatic translation from Fortran source to C, or to alternatively use a C preprocessor that is equally powerful as Fypp. Otherwise writing and maintaining C prototypes gets rather complicated, for instance [dbcsr_t_put_block](https://github.com/cp2k/dbcsr/blob/develop/src/tensors/dbcsr_tensor_block.F#L685) generates 13 different routines (and even more if the tensor library is generated with a maximum tensor rank > 4).

alazzaro:
Let's keep this issue open as a placeholder...
I agree with Patrick, that's why I wrote "ideally, the C API should be automatically generated from Fortran..."

However, as a fast start, I propose to write easy prototyping of the API by hand...

ambmax00:
Any recommendations for a C preprocessor? I planned on using fypp for C, too.

pseewald:
[...]

PR #274, opened 2019-11-25
Tensor C API

ambmax00:


ambmax00:
How should I proceed to fix the coverage issue? I'm not very familiar with it.

pseewald:
The coverage is not an issue and we can ignore it. We have switched from Doxygen to FORD for documentation but I suggest to remove the documentation boxes in `dbcsr_tensor_api_c.F` since they were just copied over anyway (of course, if there is any important Fortran documentation, please convert it to FORD).

ambmax00:
The build failed for OSX...

pseewald:
Thanks for keeping this up to date!

Can we merge this @dev-zero @alazzaro?

alazzaro:
Thanks for the work!

No, we can't merge for v2.1. The first deadline I see is after January.

ambmax00:
You're welcome.
I have started writing a program with this interface, so there will be a few changes here and there in the future, if I encounter any bugs. 
By the way, is there a function to do dot products with tensors?

pseewald:
> By the way, is there a function to do dot products with tensors?

Not at present, I'm doing workarounds like this:
https://github.com/cp2k/cp2k/blob/master/src/rpa_gw.F#L5374
That is straightforward but only for the simple case that the result is a vector that can be stored locally on each MPI rank. A general implementation of dot products is under way.

If you write a program based on DBCSR tensor API, there are a few things to know for performance:
- you should choose block sizes not too large since the resulting number of elements in one block can grow large for tensors of rank 3 or more, resulting in bad load balancing. In CP2K I set a maximum block size of 10.
- load balancing depends on the chosen distribution. For sparse data with heterogeneous block sizes, I heuristically choose the distribution vectors like this:
https://github.com/cp2k/cp2k/blob/master/src/qs_tensors_types.F#L236-L258

Please contact me if you have questions. Documentation is currently not the best, I hope I can provide better documentation and a commented example beginning of next year.

ambmax00:
Ok, I will keep those points in mind. My block sizes will also be around 10 to 20. 

Another question: is there a reason why we are doing the C interface with void pointers for non-trivial types like tensor, pgrid etc? Wouldn't it be more beneficial to also bind derived types and use struct types? This would allow to access the members of the types in the C code, too.

pseewald:
Sorry for the late reply, I did not see your questions at first sight.

> And concerning tensor contraction: As far as I can understand, you should make sure that the indices to be contracted are on the same "side" of the underlying matrix, e.g. for a tensor contraction (i,k,j) x (i,j,l,m) = (k,l,m), you should distribute them like (ij | k) x (ij | lm) = (k|lm). If the tensors are not compatible, redistribution happens automatically, right?

Yes, that is correct. The tensor API is designed in a way that redistributions that are not strictly necessary can be avoided. For tensors of very different sizes/occupancies, redistribution may be more expensive than performing the matrix multiplication. This is why the API lets the user choose the matrix representation of the tensor (via the `map1_2d` and `map2_2d` arguments). But redistribution to a consistent layout happens automatically if needed in `dbcsr_t_contract`.

> 
> Now, if I have a lot of tensors which are always contracted in different ways, is it better to keep copies around with different distributions, or should I let dbcsr redistribute it by itself?

I recommend to write a first version of the code as simple as possible, especially if you have a lot of tensors. Performance is difficult to predict, especially with tensors of very different sizes and sparsity. In my code, I prefer to do all redistributions manually with `dbcsr_t_copy` just to be able to profile redistributions independently of contractions. A further advantage of manual redistribution with `dbcsr_t_copy` is that the tensors don't need to have the same block sizes, whereas identical block sizes is a requirement for `dbcsr_t_contract`. The internal optimizations depend on the chosen process grid dimensions - I recommend to choose the process grid dimensions proportionally to the tensor dimensions.

> About distributions: Both in creating a dist and a tensor, we need to give it two maps, map1_2d and map2_2d. Do they need to be the same in both cases?

For the time being they need to be the same. But this is definitely a shortcoming of the current API and to be improved. I opened an issue #280 to remind myself about this (and other) issues.

> 
> Another question: is there a reason why we are doing the C interface with void pointers for non-trivial types like tensor, pgrid etc? Wouldn't it be more beneficial to also bind derived types and use struct types? This would allow to access the members of the types in the C code, too.

You should not try to access members of the types if you use the API. If you need to extend the API, this should be done in the Fortran implementation. Please open an issue if something is missing in the API, and I can add this.

ambmax00:
Thanks for the detailed explanation! I am making good progress.

> You should not try to access members of the types if you use the API.

Ok, I have written my c++ wrapper class in a way I won't need any of the members.

Another question I got:
If I want to do a tensor contraction (1|2) * (12|34) = (3|4) , I guess I have to introduce a dummy tensor dimension such that it's possible to redistribute the indices (12|X) * (12|34) = (34|X) ? i get an error if I try it straight forward.

pseewald:
> Thanks for the detailed explanation! I am making good progress.
> 
> > You should not try to access members of the types if you use the API.
> 
> Ok, I have written my c++ wrapper class in a way I won't need any of the members.

I'm aware that some API routines are missing to access info about process grids among other things. Good that you could work around this.

> 
> Another question I got:
> If I want to do a tensor contraction (1|2) * (12|34) = (3|4) , I guess I have to introduce a dummy tensor dimension such that it's possible to redistribute the indices (12|X) * (12|34) = (34|X) ? i get an error if I try it straight forward.

Operations like this are not yet implemented but will be in the future. For the time being I would choose the same workaround as you did, and this should work. If this triggers an error and needs debugging, could you maybe send me a simple example that I can try myself?

ambmax00:
Yes, if I get any error, I will send you an example.

ambmax00:
I will also add the "dbcsr_t_get_mapping_info" subroutine to the C API, I think I might need that. So there will be some more changes before merging.

pseewald:
> About distributions: Both in creating a dist and a tensor, we need to give it two maps, map1_2d and map2_2d. Do they need to be the same in both cases?

In a first round of API improvements (7b2e9d0fc040fc4234d533b2ed139ae8dafa555c), I removed the `map1_2d` and `map2_2d` arguments from the routines that create grids or distributions (the mappings are internally handled).

ambmax00:
Tensor contractions with dummy dimensions are working without any problems.

> In a first round of API improvements (7b2e9d0), I removed the map1_2d and map2_2d arguments from the routines that create grids or distributions (the mappings are internally handled).

Ok, I will update the interface accordingly.

ambmax00:
For reserving blocks in large tensors: I do not know what blocks are zero before computing them (electron integrals). Is it better to keep a reserve list, and then reserve them all at once, or in batches, or can I do it individually, e.g. in a for-loop after generating the respective block? What threshold are you using for screening blocks in cp2k?

pseewald:
Reserving blocks is not strictly necessary but causes a large overhead in `dbcsr_put_block` calls (probably all existing data needs to be reshuffled). My experience was that reserving blocks in batches also causes noticeable overhead. So I recommend to reserve all blocks at once.

In CP2K, we use distance screening for atomic basis functions to know in advance which integrals need to be calculated, and we reserve DBCSR blocks accordingly. The integrals are then calculated in a second step. After calculating all integrals, we filter the tensors (`dbcsr_t_filter`) to refine the sparsity even more. Filtering manually is a good idea before calling `dbcsr_t_contract` because the internal filtering is applied to the result tensor (`tensor_3`) but does not reduce the amount of data in `tensor_1` and `tensor_2` being communicated.

In CP2K, we typically consider 10^(-9) a safe default for filtering (`eps_filter` parameter).

ambmax00:
Hi again, some questions came up concerning openmp parallelization.
What's a good way to iterate over local non-zero blocks with openmp? The while(iterator_blocks_left) doesn't really allow to do that.
For a N-dimensional tensor, maybe I just construct a  N-nested loop, loop over all blocks, and skip over the zero-blocks?
Are iterator_next_block, put_block and get_block thread-safe?

pseewald:
Sorry for the late response. We have no experience with using the tensor API with more than one thread per MPI rank (scaling is good with flat MPI with our applications). But the DBCSR matrix library distributes blocks over threads and, if I'm not wrong, that feature should work without issues from the tensor API (since the get/put and iterator routines are simple wrappers around DBCSR functions).

DBCSR internally distributes blocks over threads and the `iterator` routines return different blocks on each thread if called within a parallel region. DBCSR uses a different work matrix for each thread to ensure thread-safety of the `put_block/get_block` routines. At the end of the parallel region `dbcsr_t_finalize` must be called to merge all work matrices.

For further documentation for iterators please have a look [here](https://github.com/cp2k/dbcsr/blob/develop/src/block/dbcsr_iterator_operations.F#L99-L133). Please note that only the default behaviour can be used with the DBCSR API.

I hope this works as expected - could you let me know if you encounter any difficulties?

alazzaro:
@ambmax00 Sorry for the late reply, now it is the right time to work on this PR.
Could you rebase with the current develop?

pseewald:
There have been some more API changes, could you update the C API accordingly?

ambmax00:
Ok, I'm on it.

alazzaro:
I don't have any objection to merge this PR as far as @pseewald (who is the tensor expert) is happy for that... 
Please update and we will merge ;)

ambmax00:
Do you know what causes the pre-commit error?

pseewald:
It says:
```
src/dbcsr_tensor_api_c.F ... required header not found
examples/dbcsr_tensor_example_1.cpp ... required header not found
```
@dev-zero could you have a look?

alazzaro:
@ambmax00 
OK, sorry, I didn't look at Patrick's comment... forget my previous comment...
The problem is easy to fix:
You need to add the default headers on the top of your files:

src/dbcsr_tensor_api_c.F
examples/dbcsr_tensor_example_1.cpp 

The templates are in:

https://github.com/cp2k/dbcsr/tree/develop/.pre-commit/headers

tests/CMakeLists.txt
```diff
@@ -67,6 +71,7 @@ set(dbcsr_unittest3_SRCS dbcsr_unittest3.F)
 set(dbcsr_tensor_unittest_SRCS dbcsr_tensor_unittest.F)
 set(dbcsr_tas_unittest_SRCS dbcsr_tas_unittest.F)
 set(dbcsr_test_csr_conversions_SRCS dbcsr_test_csr_conversions.F)
+set(c_dbcsr_tensor_unittest_SRCS ../examples/dbcsr_tensor_example_1.cpp)
```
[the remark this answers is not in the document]
ambmax00: Ok, I will split them. The test is just there to check that the c code doesn't segfault due to errors in the interface or any nullpointer stuff, so checking if the example can run was my test (for now).

ambmax00:
Ok, this should be it.
I decided to leave the c interface pretty bare bones. You could make it a bit more user-friendly and cpp-like, by introducing template classes to avoid null pointers (e.g. tensor<3,double>), but in my opinion, that's the responsibility of the user. An example of how something like that might look like, [here's the one I'm using.](https://github.com/ambmax00/chem/blob/master/math/tensor/dbcsr.hpp).

src/PACKAGE
```diff
@@ -2,6 +2,6 @@
 "description": "Distributed Block Compressed Sparse Row, A sparse matrix library",
 "archive": "libdbcsr",
 "requires": ["acc/cublaswrap", "acc/hipblaswrap", "base", "mpi", "data", "dist",
-"block", "ops", "mm", "core", "utils", "work"],
+"block", "ops", "mm", "core", "utils", "work", "tensors"],
```
alazzaro: At this point, I think you can remove "tensors" here...

alazzaro:
OK, thanks, I did few other minor remarks. I'm going to test it on Daint too...

alazzaro:
retest this please

alazzaro:
BTW, add yourself tot he list of authors ;) 

https://github.com/cp2k/dbcsr/blob/develop/AUTHORS

ambmax00:
It *should* work now.

alazzaro:
OK, thanks for all fixes!

Daint takes a while to run, I will report to you (you can follow it on: https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-435/ )

At this point,  the only open "suggestion" is to rename the test...
Likely today we can (finally) merge...

alazzaro:
retest this please

ambmax00:
What would you like to call the test?

alazzaro:
@ambmax00 it appears that both Intel and Cray compilers don't like the C++ tensor test.
See

https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-436/

Do you have any possibility to test with one of these compilers somewhere?
It seems a memory problem (double free), so it can be that you can use valgrind with GNU to diagnostic it.. 

If you cannot get more insight, I will try to add some debugging info to the Daint test...

ambmax00:
I can use the intel compiler. I'll have a look at it.

ambmax00:
The problem is that a pointer to an array allocated in Fortran cannot be freed in C.
The Solution is to use CFI_CDESC ([example](https://www.ibm.com/support/knowledgecenter/SSGH4D_16.1.0/com.ibm.xlf161.aix.doc/language_ref/allocandpointerargu_expl.html)). This is only necessary for certain arrays in get_info and get_mapping_info, which are allocated on the Fortran side.
Alternatively, it might also be a good idea to use CFI_DESC for the whole interface instead of naked pointers? It could make the interface a bit easier and safer.

pseewald:
The sizes of the arrays you mentioned are known:

For `dbcsr_t_get_info`:
- `blks_local` should not be needed, it exists only because the tensor API should support all features as the DBCSR matrix library (it is the generalization of `local_rows` and `local_cols`). I suggest to remove this from the C API.
- `blk_size_i`, `blk_offset_i` and `proc_dist_i` have size `nblks_total(i)`.
For `dbcsr_t_get_mapping_info`:
- `dims1_2d` and `map1_2d` have size `map%ndim1_2d`
- `dims2_2d` and `map2_2d` have size `map%ndim2_2d`

So I think the easiest solution would be to copy these arrays to static arrays in `dbcsr_tensor_api_c.F`.

I have never seen or used `CFI_CDESC` before but I guess we should only use it if absolutely necessary (also w.r.t. supporting older compiler versions).

alazzaro:
> The problem is that a pointer to an array allocated in Fortran cannot be freed in C.

Can we avoid that? I would prefer to keep the two allocations separate, so the API should be designed aronud that. Would it be an option to have Fortran routines for deallocation, such that you call them from C?

I agree with Patrick, we have to design the library such that you don't use "advanced" features...

pseewald:
> Would it be an option to have Fortran routines for deallocation, such that you call them from C?

There is no need to do this (see my comment above).

ambmax00:
Ok I will follow the advise of @pseewald. But I think blks_local should be kept. I actually use it a few times in my code. 

Another question for dbcsr_t_contract: When I have `retain_sparsity = true`, does it actually enforce the sparsity during the calculation, i.e. does it only calculate the necessary blocks? Or does it calculate all, and then remove the unwanted blocks?

pseewald:
> Ok I will follow the advise of @pseewald. But I think blks_local should be kept. I actually use it a few times in my code.

Ok good, then let's keep it. Of course `blks_local_i` has size `nblks_local(i)`, I think you figured that out by now.

I see that it's not very user-friendly to require that all arrays are allocated with the correct size on the C side, especially if sizes are not well-documented. How should this ideally be solved? My idea is to have very light-weight functions returning often used sizes (such as `nblks_local(tensor, i)`) and document the required array sizes in the C wrappers.

pseewald:
> Another question for dbcsr_t_contract: When I have `retain_sparsity = true`, does it actually enforce the sparsity during the calculation, i.e. does it only calculate the necessary blocks? Or does it calculate all, and then remove the unwanted blocks?

It selectively calculates only those blocks that exist in `tensor_3`.

ambmax00:
> My idea is to have very light-weight functions returning often used sizes (such as nblks_local(tensor, i)) and document the required array sizes in the C wrappers.

Sounds good to me. Then we can also use wrapper functions, to make it easier. I will adapt the tensor interface and document the array sizes.

>It selectively calculates only those blocks that exist in tensor_3.

That's great. This will make things a lot easier for me. :)

pseewald:
> Sounds good to me. Then we can also use wrapper functions, to make it easier. I will adapt the tensor interface and document the array sizes.

Ok then I create utility functions returning often-used array sizes in the Fortran code.

src/tensors/dbcsr_tensor_api_c.F
```diff
@@ -0,0 +1,1231 @@
...
+      TYPE(c_ptr), INTENT(IN), VALUE :: c_tensor
+      TYPE(dbcsr_t_type), POINTER :: tensor
+      
+      CALL c_f_pointer(c_tensor, tensor)
+      
+      CALL dbcsr_t_finalize(tensor)
+
+   END SUBROUTINE
+
+   SUBROUTINE c_dbcsr_t_pgrid_create_expert(fcomm, c_dims, dims_size, c_pgrid, &
```
pseewald: I created the `expert` routines for internal use and they should not be exposed to the API. Do you use this? It's part of the Fortran API only for testing with CP2K. I may remove this at a later time.

src/tensors/dbcsr_tensor_api_c.F
```diff
@@ -0,0 +1,1231 @@
...
+
+      CALL c_f_pointer(c_tensor, tensor)
+
+      c_ndims_tensor = tensor%nd_index%ndim_nd
+   END FUNCTION
+
+#:for dparam, dtype, dsuffix, basetype, cftype in cf_dtype_float_list
+#:for ndim in ndims
+
+   SUBROUTINE c_dbcsr_t_allocate_and_get_${ndim}$d_block_${dsuffix}$ (c_tensor, tensor_dim, c_ind, c_block, c_found) &
```
pseewald: If all allocations should be controlled in C, we should not expose this routine.

I'll write a function to more easily get block sizes.

src/tensors/dbcsr_tensor_api_c.F
```diff
@@ -0,0 +1,1231 @@
...
+      TYPE(dbcsr_t_iterator_type), POINTER               :: iterator
+      LOGICAL(kind=c_bool)                               :: c_dbcsr_t_iterator_blocks_left
+
+      CALL c_f_pointer(c_iterator, iterator); 
+      c_dbcsr_t_iterator_blocks_left = dbcsr_t_iterator_blocks_left(iterator)
+
+   END FUNCTION
+
+
+   SUBROUTINE c_dbcsr_t_split_blocks(c_tensor_in, tensor_dim, c_tensor_out, c_block_sizes, c_nodata) &
```
pseewald: This routine should not be in the API (my bad, I see it's in the Fortran API and used within CP2K), `dbcsr_t_copy` should do all the tricks. We can leave it for now but I may remove it later.

src/tensors/dbcsr_tensor_api_c.F
```diff
@@ -0,0 +1,1231 @@
...
+      ENDIF
+      
+#:set list = ['nblks_total', 'nfull_total', 'nblks_local', 'nfull_local', 'my_ploc', 'pdims' ]
+#:for var in list
+      !IF (PRESENT(c_${var}$)) NULLIFY(${var}$)
+#:endfor
+      
+   END SUBROUTINE
+   
+ !  FUNCTION c_ndims_mapping(map) BIND(C, name="ndims_mapping")
```
pseewald: This should probably be removed

src/tensors/dbcsr_tensor_api_c.F
ambmax00: I don't use it, so you can remove it later.

src/tensors/dbcsr_tensor_api_c.F
ambmax00: Ok, I will adapt it later then.

src/tensors/dbcsr_tensor_api_c.F
ambmax00: Ok.

src/tensors/dbcsr_tensor_api_c.F
ambmax00: Will do.

pseewald:
I decided to update the Fortran API in #320 such that there are no arrays allocated in API routines. The C wrappers need to be updated but they can be written exactly like the Fortran routines. I hope this makes things simpler.

The only routine that returns an allocatable is `dbcsr_t_contract_index`. I suggest to write a function that deallocates the array in `dbcsr_tensor_api_c.F`.

ambmax00:
Thanks for your work. I'm currently adapting the interface.

> I suggest to write a function that deallocates the array in dbcsr_tensor_api_c.F

There is still a problem with that when using the intel compiler. Ideally, we would allocate an array in Fortran, pass its address to the c code. Later, we call a function to deallocate it in Fortran, so we pass the array pointer from c back to Fortran. But by having made the detour into c, by passing the address, the Fortran Array has lost some of its hidden parameters, and it is impossible to deallocate it later in Fortran, or in C if using intel compilers.

The only way to circumvent this, without using CFI, is to have a global array of allocatable arrays or pointers in the fortran module e.g.

```` 
   TYPE intarray
      INTEGER, DIMENSION(:), POINTER :: p
   END TYPE intarray
  
   TYPE(intarray_1), DIMENSION(256) :: intarray_all
   ! 256 or any other number that indicates how many allocations are allowed
 ````
Fortran can then look for allocatable arrays, and use them to store data. Then the pointer is passed to C, but no information is lost. Then you can call a deallocate function in C which frees the memory at the correct place in intarray_all. Even if I would use something like this, result_index in `dbcsr_t_contract_index` does not accept pointers as argument. And you cannot do the trick above with allocatable arrays instead of pointers...

That's the only way at the moment, but it comes with a whole set of new problems. It's a known problem for intel compilers, and a reason why the c descriptors have been developed. See the discussion [here](https://software.intel.com/en-us/forums/intel-fortran-compiler/topic/807005)

**Comment by pseewald** [comment]
I'm not sure whether I understand the workaround. Can we not do this:

```fortran
type c_array_2d
   integer, dimension(:, :), allocatable :: array
end type

subroutine c_dbcsr_t_contract_index(..., result_index)
...
   type(c_array_2d), intent(out) :: result_index
   call dbcsr_t_contract_index(..., result_index%array)
end subroutine

subroutine deallocate_c_array_2d(array)
   type(c_array_2d), intent(inout) :: array
   deallocate(array%array)
end subroutine
```

ambmax00:
Self declared types are not c interoperable. You can do something like:
````
type, bind(C) :: c_array_2d 
        integer, dimension(:, :), allocatable :: array
end type
````
and then in the header:
````
struct c_array_2d {
		int* array;
}
````
But then the compiler complains that array is not allowed to be allocatable.

**Comment by pseewald** [comment]
Ok I get it. For `dbcsr_t_contract_index` we know an upper bound to the size of `result_index` which is the number of local blocks if the tensor was dense. So I can make this array static and get rid of intent(out) allocatables for good.

**Comment by ambmax00** [comment]
Ok, I think that would be the best way. 
But I still advocate the usage of c_desc, maybe later down the road. It's also practical on the c side for managing arrays. :)

**Comment by alazzaro** [comment]
@ambmax00 just to let you know that University of Zurich is organizing an event on Fortran (see https://tcevents.chem.uzh.ch/event/12/) You might want to submit an abstract for a presentation with what you are doing...

**Comment by ambmax00** [comment]
@alazzaro Would a presentation focus more on the theory part (so computational chemistry, and why sparse block tensors are useful), or on the coding part (I'm writing in C++, so I would talk about the C-Fortran interface?)

**Comment by alazzaro** [comment]
> @alazzaro Would a presentation focus more on the theory part (so computational chemistry, and why sparse block tensors are useful), or on the coding part (I'm writing in C++, so I would talk about the C-Fortran interface?)

Well, it is a Fortran event, so it is about coding... my proposal was that you can share your experience on the Fortran/C++ interface (in particular related to the memory management)... It is definitely a good topic to present!

**Comment by pseewald** [comment]
Is this ready to be merged? Can I help with anything?

**Comment by alazzaro** [comment]
I see that you are hitting my last check on the number of subcommunicators:

`ABORT in dbcsr_lib.F:322 Number of DBCSR sub-communicators is not zero!`

I've added this check to make sure that the destroy all created subcomminicators when we finalize the library. Unfortunately, this is not easy to debug...
I do suggest to:

1. add some print statements just after all calls that create subcommunicators (check for the line `debug_comm_count` in the file `src/mpi/dbcsr_mpiwrap.F`
2. Add a watchpoint with gdb to this variable and see the stack trace of the calls

Usually, the problem is that a matrix is not destroyed or you are creating/destroy communicators outside DBCSR calls...

**Comment by alazzaro** [comment]
retest this please

**Comment by alazzaro** [comment]
So, Intel compiler still shows problem with the communicators numbers. 
Cray compiler is crashing. I will run the cray compiler job and get a stacktrace for you.

**Comment by alazzaro** [comment]
This is the Cray stacktrace:
```
#0  dbcsr_put_block_d$dbcsr_block_access_ () at block/dbcsr_block_access.F:985
#1  0x000000000043d1eb in dbcsr_put_block2d_d$dbcsr_block_access_ () at block/dbcsr_block_access.F:942
#2  0x000000000080542c in dbcsr_tas_put_block_r_dp$dbcsr_tas_base_ () at tas/dbcsr_tas_base.F:1158
#3  0x000000000065fbb6 in dbcsr_t_put_3d_block_r_dp$dbcsr_tensor_block_ () at tensors/dbcsr_tensor_block.F:974
#4  0x0000000000422d5e in c_dbcsr_t_put_3d_block_r_dp () at tensors/dbcsr_tensor_api_c.F:1343
#5  0x0000000000410e3a in void fill_random<double>(void*, std::vector<std::vector<int, std::allocator<int> >, std::allocator<std::vector<int, std::allocator<int> > > >, std::function<double ()>&) ()
```
I'm investigating more...

**Comment by alazzaro** [comment]
@ambmax00 After a bit of more investigation, we think we found the problem with the Cray compiler crash. In short, the problem is in the file dbcsr_tensor_api_c.F with the OPTIONAL parameters, more specifically for the LOGICAL parameters. In particular, there is a crash if the parameter is not present. It seems that this case is not protected, for example:

```
     LOGICAL(kind=c_bool), INTENT(IN), OPTIONAL         :: c_summation
      ${basetype}$ (kind=${cftype}$), INTENT(IN), &
         OPTIONAL                                          :: c_scale

      TYPE(dbcsr_t_type), POINTER                        :: tensor
      LOGICAL, ALLOCATABLE                               :: summation
      ${dtype}$, ALLOCATABLE                             :: scale

      IF (PRESENT(c_summation)) summation = c_summation
      IF (PRESENT(c_scale)) scale = c_scale

      CALL c_f_pointer(c_tensor, tensor)

      CALL dbcsr_t_put_block(tensor, c_ind + 1, c_sizes, &
                             c_block, summation, scale)

```

Here, you always call the internal Fortran function always with the OPTIONAL parameter, so any `IF (PRESENT(parameter))` will be true. The problem is when summation or scale are not present... In this case, the internal function will try to use an aLLOCATABLE without being allocated and it will raise an error. No idea why GNU and Intel are not giving error...
Even if you remove the ALLOCATABLE, this is still wrong since you are not setting any value for these variables...

So, a possible solution would be to make a set of `IF (PRESENT(parameter))` and then make the proper call to the internal Fortran subroutine. In the above case, there is a fortunate situation: intrinsic types (c_int, c_double...) maps directly to Fortran types. This is not the case fo the LOGICAL. So, you can change the code in:

```
     LOGICAL(kind=c_bool), INTENT(IN), OPTIONAL         :: c_summation
      ${basetype}$ (kind=${cftype}$), INTENT(IN), &
         OPTIONAL                                          :: c_scale

      TYPE(dbcsr_t_type), POINTER                        :: tensor
      LOGICAL                                                           :: summation

      CALL c_f_pointer(c_tensor, tensor)

      IF (PRESENT(c_summation)) THEN
         summation = c_summation
         CALL dbcsr_t_put_block(tensor, c_ind + 1, c_sizes, &
                               c_block, summation, c_scale)
      ELSE
         CALL dbcsr_t_put_block(tensor, c_ind + 1, c_sizes, &
                               c_block, scale=c_scale)
      ENDIF

```
Unfortunately, if you have more optional no-interoperable arguments, that means you have to do a chain of `IF (PRESENT(this) and .NOT. PRESENT(that) ...)`. I hope we don't have so many cases like that (I can see c_ptr, c_bool, and strings). @pseewald has a way to use fypp to generated such cases...

**Comment by alazzaro** [comment]
One more error. In the function:

```
  SUBROUTINE c_dbcsr_t_iterator_stop(c_iterator) BIND(C, name="c_dbcsr_t_iterator_stop")                                                                                                                                                                      
                                                                                                                                                                                                                                                               
      TYPE(c_ptr), INTENT(INOUT)                         :: c_iterator                                                                                                                                                                                         
      TYPE(dbcsr_t_iterator_type), POINTER               :: iterator                                                                                                                                                                                           
                                                                                                                                                                                                                                                               
      CALL c_f_pointer(c_iterator, iterator)                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                               
      CALL dbcsr_t_iterator_stop(iterator)                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                               
      DEALLOCATE (iterator)                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                               
      c_iterator = c_null_ptr                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                               
   END SUBROUTINE                       
```

The cray compiler complains with the following message:

 An argument in the DEALLOCATE statement is a disassociated pointer, an
 unallocated array, or a pointer not allocated as a pointer.

Does the `dbcsr_t_iterator_stop` already deallocates the iterator?
For this kind of problem, we have to set memory leaks and valgrind tests...

**Comment by ambmax00** [comment]
@alazzaro Normally, variables that are unallocated, and are passed as optional argument to a function, the function treats it as not present (if the variable does not have the allocatable attribute). Maybe the cray compiler doesn't work that way. Alternatively, I can make it a pointer, which is either associated or null, and then it should work.

I will look at the other problems too.

**Comment by ambmax00** [comment]
Another question. In a recent commit, I removed a few pointer/target lines, such as:
````
subroutine c_function(c_array) BIND(C,name="c_function")
INTEGER(kind=c_int), INTENT(IN), TARGET, DIMENSION(dim) :: c_array
INTEGER, POINTER, DIMENSION(:) :: f_pointer

f_pointer => c_array

CALL f_function(f_pointer)
END subroutine
````
And replaced it directly by:
````
subroutine c_function(c_array) BIND(C,name="c_function")
INTEGER(kind=c_int), INTENT(IN), DIMENSION(dim) :: c_array

CALL f_function(c_array)
END subroutine
````
I did it like in the first example, because that's how it was done in the other file `dbcsr_api_c.F`. But I don't really see any problem with just passing it directly? Why would we want to do it with pointers?

**Comment by alazzaro** [comment]
> I did it like in the first example, because that's how it was done in the other file `dbcsr_api_c.F`. But I don't really see any problem with just passing it directly? Why would we want to do it with pointers?

I simply don't know... In my previous post, I wrote that you should go for the second solution, i.e. don't need to use temporary arrays with pointers. Let me try this change and see if it works.

Concerning the ALLOCATABLE consider as not-present OPTIONAL, I don't know either... I'm not sure that the use of pointers will solve the problem in any case, I mean PRESENT will still give true and now you have a null pointer... I will ask around if this is a missing feature in the Cray compiler...

**Comment by alazzaro** [comment]
I asked around, the only way to make an OPTIONAL PRESENT to be false is that it has not to be present during the function call. So, neither ALLOCATABLE or null POINTER would work. You really need to go for a set of protected calls with `IF (PRESENT)`... The F2018 standard says:
```
Fortran 2018 draft says (of PRESENT(A))
Argument. A shall be the name of an optional dummy argument that is accessible in the subprogram in which
the PRESENT function reference appears. There are no other requirements on A.
```

Concerning the question on the dbcsr_api_c.F, you are right, the direct calls of the pointers are perfectly fine, I will make the change.

Concerning the deallocates (similar to what I mentioned here https://github.com/cp2k/dbcsr/pull/274#issuecomment-614560525), the same problem is in the dbcsr_api_c.F. I'm investigating more...

**Comment by ambmax00** [comment]
@alazzaro Ok, thanks.
The subcommunicator problem might actually be related to the the whole OPTIONAL problem. The dbcsr_t_contract function creates more subcommunicators than it destroys. I will first address the optional stuff, and see if that fixes it.

**Comment by alazzaro** [comment]
OK, me again: digging more in the standard, you are actually right:
```
A dummy argument or an entity that is host associated with a dummy argument is not present if the dummy argument does not correspond to an actual argument, corresponds to an actual argument that is not present, or does not have the ALLOCATABLE or POINTER attribute, and corresponds to an actual argument that { has the ALLOCATABLE attribute and is not allocated, or { has the POINTER attribute and is disassociated. }
```

So, simply the cray compiler doesn't have this feature yet...

alazzaro:
@ambmax00 You are hitting a problem in the tensors... We have to wait for @pseewald to fix it...

On the other side, I'm considering to drop CCE compiler tests for the C API if we don't get it to work (it seems the compiler is missing some F2018 feature)

ambmax00:
@alazzaro  Ok. 
As you proposed, I used fypp to generate chains of `IF(PRESENT(var1) .AND. .NOT. PRESENT(var2) ...)` to take care of problematic types. In order to not have hundreds of IF/ELSE barnches,  I set default values for the `LOGICAL`s (e.g. in `c_dbcsr_t_contract`). I hope it looks okay.

alazzaro:
> @alazzaro Ok.
> As you proposed, I used fypp to generate chains of `IF(PRESENT(var1) .AND. .NOT. PRESENT(var2) ...)` to take care of problematic types. In order to not have hundreds of IF/ELSE branches, I set default values for the `LOGICAL`s (e.g. in `c_dbcsr_t_contract`). I hope it looks okay.

Yes, I saw that. The problem is that CCE doesn't like the Pointer interoperability, so the `DEALLOCATE` statement will fail (even if you add `IF (ASSOCIATED())...`. This is hopeless...
At this point, I would like to ask you if you can remove the useless `IF` (sorry for "wasting" your time on that).

Then, I think you can rebase to fix the conflicts and we are ready to merge. I will add the change to drop the CCE compiler for the C API.

ambmax00:
Ok, will do.

ambmax00:
By the way, would you like me to do an overhaul for the C interface for matrices, too? There are a few functions missing that might be useful, and I could extend it to different types, too, like for the tensor interface.

alazzaro:
> By the way, would you like me to do an overhaul for the C interface for matrices, too? There are a few functions missing that might be useful, and I could extend it to different types, too, like for the tensor interface.

Let's do in a different PR...

alazzaro:
retest this please

ambmax00:
@pseewald Could you maybe give me an example how to use iterators with OpenMp?
In Fortran code, I would do it like this:
````
TYPE(dbcsr_t_type) :: tensor
TYPE(dbcsr_t_iterator_type) :: iter
...
$!OMP parallel threadprivate(iter)

dbcsr_t_iterator_start(iter,tensor)

DO WHILE(dbcsr_t_iterator_blocks_left(iter))
   ... DO STUFF ....
END DO

dbcsr_t_iterator_stop(iter)
dbcsr_t_finalize(tensor)

$!OMP end parallel
````

Do you do it like this? I get errors when using a similar way in C, so I wonder if there may be some error in Fortran or in the way I do it.

ambmax00:
Continuing the discussion for omp directives on #248 ...

PR #284, opened 2020-01-09
libsmm_acc: bugfix

shoshijak:
- **libsmm_acc**: fix a concurrency bug

    The static variable `cpu_fallback`, once set to `false` by one
    thread, was causing other threads to fall back to the CPU although
    they could launch a kernel on the GPU.
    Solve this issue by determining whether to fall back to the CPU
    using the state of the "kernel handle", and getting rid of the static
    variable `cpu_fallback`.
    Also introduce a typedef and some simplifications for readability.

- **libsmm_acc**: remove superfluous ifdefs around openmp pragmas
- **tests**: add very small and very large block sizes

### DIFF
--- a/src/acc/libsmm_acc/libsmm_acc.cpp
+++ b/src/acc/libsmm_acc/libsmm_acc.cpp
@@ -186,7 +186,9 @@ inline void jit_kernel(ACC_DRV(function)& kern_func, libsmm_acc_algo algo, int t
 }
 
 
-void add_kernel_handle_to_jitted_kernels(ACC_DRV(function) kern_func, ACC_DRV(stream) stream, Triplet h_mnk, int& threads, int& grouping, bool& cpu_fallback){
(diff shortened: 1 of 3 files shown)


dev-zero:
LGTM, but Daint doesn't seem to like it

shoshijak:
The issue on daint is not with my changes to the code base, but with the additional test I introduced ([this one](https://github.com/cp2k/dbcsr/pull/284/commits/29fcd310bdb43f0040eab9b070f196c31754344d#diff-680398c34a0dea43588b0ec42bb1321fR115-R120)). 

More specifically, the transpose kernel of size 79x79 fails on GPU. I'm trying to figure out from what sizes onwards this kernel fails so I can deal with this more cleanly than this crash... I'll commit here again soon.

Alternatively, I can remove this test from this PR, so the fix can be added, and work on the transposition with large sizes separately.

shoshijak:
I changed the new test that was causing the CI to fail.
If this passes the CI, I suggest we merge this, and I take care of the issue with transpose's big block sizes in a separate PR.

PR #288, opened 2020-01-22
make: add -D__CUDA if compiling with nvcc

shoshijak:
If compiling with nvcc, automatically add `-D__CUDA` to `NVFLAGS` in the `Makefile`.

This removes the need to add `-D__CUDA` to CP2K arch files when compiling CP2K with DBCSR > v2.0.1. 

Solves the issue in https://github.com/cp2k/cp2k/pull/744 .


shoshijak:
CI fails because of https://github.com/cp2k/dbcsr/issues/289, I'll merge this.

Issue #291, opened 2020-01-31
Installation Guide: page not found


pseewald:
The link to the installation guide in README.md does not work.

alazzaro:
Fixed via #307

Issue #289, opened 2020-01-24
Building with CMake 3.10 fails since  #285


shoshijak:
Building with CMake 3.10 fails since commit [34bfa005e1aba2982707917a499839aea220c764](https://github.com/cp2k/dbcsr/commit/34bfa005e1aba2982707917a499839aea220c764) (PR #285), with a segfault between configuration and generation:

```
-- FYPP preprocessor found.
Tests will run with 4 MPI ranks and 3 OpenMP threads each
-- Configuring done
Segmentation fault (core dumped)
```

All PRs subsequent to #285 thus fail the CI test `gnu-cmake-3.10.build`. 

Building with CMake >= 3.14 doesn't present any issues.

dev-zero:
Can reproduce manually on Daint, but not locally, might be [this bug](https://gitlab.kitware.com/cmake/cmake/commit/ab3b549ee0dcd79607f8347a9b69c9b029590fbd).\
Reported to CSCS to forward it to SUSE.

dev-zero:
OK, it's not only SUSEs CMake-3.10.2, but also the one in Ubuntu Bionic (and therefore in the `nvidia/cuda:devel`. Theo from CSCS sent me the following `Dockerfile` as a reproducer:

```
FROM nvidia/cuda:10.1-devel

RUN apt-get update && \
    apt-get install -y cmake git \
    libblas-dev liblapack-dev mpich python3-dev

RUN git clone --recursive https://github.com/cp2k/dbcsr.git && \
    mkdir build && cd build && \
    cmake -DCMAKE_CROSSCOMPILING_EMULATOR="" -DUSE_CUDA=ON -DUSE_CUBLAS=ON -DBLAS_FOUND=ON -DBLAS_LIBRARIES="-lsci_gnu_mpi_mp" -DLAPACK_FOUND=ON -DLAPACK_LIBRARIES="-lsci_gnu_mpi_mp" ../dbcsr/ --debug-output
```

So, we have 3 options:

1. Revert (parts of) the commit
2. Motivate distros to pull a backported patch for the issue
3. Drop support for CMake 3.10

Opinions?

dev-zero:
oh, and removing `-DCMAKE_CROSSCOMPILING_EMULATOR=""` makes the segfault go away, so it is the cited CMake issue, meaning that before #285 we were not calling the emulator if needed, so the change in #285 is actually correct.

shoshijak:
@dev-zero, thank you for doing all this research!

[...]

PR #287, opened 2020-01-21
Make FORD documentation & README more user friendly

shoshijak:
## Add documentation pages to FORD in tab called "Guide"

Screenshot 2020-01-21 at 13 39 13

## Make DBCSR's README lighter

Screenshot 2020-01-21 at 13 40 51

Issue #263, opened 2019-11-20
Remove makefile from the official documentation


alazzaro:
Since CMake will be the official way to compile the library, and makefile will be only considered for CP2K compilation, I propose to drop any Makefile mention in the documentation and move the Makefile and Makefile.inc within a `.make` directory, leaving the requirement to update Makefile.inc for the developers who want any new feature to be used in CP2K. 

Furthermore, we will still use the makefile for coverage and some testings...

dev-zero:
wrt coverage: CMake accepted [my tiny patch](https://gitlab.kitware.com/cmake/cmake/merge_requests/4048#note_653427) which fixes building with `fypp -n` together with ninja and is now in their nightly release for 3.17.0 :)
So, my plan is to switch to cmake-nightly for the Travis-CI and ninja for the coverage (keeping make for the rest to ensure we don't break any of the two).

alazzaro:
Great! We will keep make because of CP2K and we need to test it. I will flag this issue as v2.1 then...

alazzaro:
https://github.com/cp2k/dbcsr/pull/241#issuecomment-557037969

alazzaro:
Current Makefile serves for
- [ ] Pretty. We can move it to pre-commit (see https://github.com/cp2k/dbcsr/pull/270#issuecomment-558139568 )

- [ ] Aggressive debugging flags

- [ ] Coverage (see #273)

This is the `make help` output:
```
=================== Default ====================
libdbcsr                     Build DBCSR library

=================== Binaries ====================
all                          Builds all executables
dbcsr_performance_driver     
dbcsr_tas_unittest           
dbcsr_tensor_unittest        
dbcsr_test_csr_conversions   
dbcsr_unittest1              
dbcsr_unittest2              
dbcsr_unittest3              

===================== Tools =====================
pretty                       Reformat all source files in a pretty way
prettyclean                  Remove prettify marker files

================= Other Targets =================
version                      Print DBCSR version
toolversions                 Print versions of build tools
toolflags                    Print flags used with build tools
install                      Install the library and modules under PREFIX=<directory> (default /Users/alazzaro/Work/dbcsr/v2/dbcsr/install)
test                         Run the unittests available in tests/
clean                        Remove intermediate object and mod files, but not the libraries and executables
realclean                    Remove all files
help                         Print this help text
================= Variables =====================
For convenience, some variables can be set during compilation,
e.g. make VARIABLE=value (multiple variables are possible):
[...]

PR #307, opened 2020-03-05
Makefile limited to CP2K only, MISC improvements

alazzaro:
Originally, this PR was meant to remove the Makefile as a supported DBCSR build system, but then I've used to address several other issues.

- Include FPE, compiler warnings and checks (#194 )
- Remove Makefile (#263 )
- Upgrade to CMake 3.12 (#289 ). Removed Cmake test 3.10 on Daint
- Fix missing documentation link, README should use local repo links (#291 )
- Fixed travis compilation (including the update to new Ubuntu)
- Extend ban for OpenMPI versions for RMA tests (taken from 5dbe04d0810fbc25b7af016d761dba90f5fd4dd7) (#141 and #238 )
- Improve documentation in several places (Thanks to @shoshijak for the #287 !)


alazzaro:
@shoshijak Could you check if the MACROs reported at https://github.com/cp2k/dbcsr/blob/makefile_drop/docs/guide/3-developer-guide/3-programming/1-overview/index.md#list-of-macros-used-in-the-code are all for the GPU part?

shoshijak:
Do you mean whether all GPU macros are correctly reported? 
As far as I can tell, yes.

alazzaro:
thanks Shoshana, what about `__HIP_PLATFORM_NVCC__`? It is something we want to add to the table? (if so, please give me a short description)

shoshijak:
`__HIP_PLATFORM_NVCC__` is a macro provided by HIP, that DBCSR does not set itself. 
So I think we should not add it to the table.

.cp2k/Makefile
```diff
@@ -107,27 +175,22 @@ OBJ_SRC_FILES  = $(shell cd $(SRCDIR); find . ! -name "dbcsr_api_c.F" -name "*.F
...
+  ifneq (,$(findstring nvcc,$(ACC)))
     OBJ_SRC_FILES += $(LIBSMM_ACC_DIR)/../cuda/acc_cuda.cpp
     # Exclude autotuning files
     OBJ_SRC_FILES += $(shell cd $(SRCDIR);  find . ! -name "tune_*_exe*_part*.cu" ! -name "tune_*_exe*_main*.cu"  -name "*.cu")
   # if compiling with hipcc
-  else ifneq (,$(findstring hipcc,$(NVCC)))
+  else ifneq (,$(findstring hipcc,$(ACC)))
     OBJ_SRC_FILES += $(LIBSMM_ACC_DIR)/../hip/acc_hip.cpp
     # Exclude autotuning files
     OBJ_SRC_FILES += $(shell cd $(SRCDIR);  find . ! -name "tune_*_exe*_part*.cpp" ! -name "tune_*_exe*_main*.cpp"  -name "*.cpp")
```
shoshijak: I've just tested compilation of CP2K with HIP-accelerated DBCSR and noticed we should delete these 2 lines.
Indeed, this adds some `.cpp` files to `OBJ_SRC_FILES` that were already added in line 180. This leads to errors in `makedep` when compiling, since source files appead twice in the list.

.cp2k/Makefile
```diff
@@ -452,38 +340,38 @@ FYPPFLAGS ?= -n
 	$(CC) -c $(CFLAGS) $<
 
 # Compile the CUDA/HIP files
-ifneq ($(NVCC),)
```
shoshijak: This means that the name of this variable will have to be changed in CP2K's Makefile and arch files accordingly ...

CMakeLists.txt
```diff
@@ -111,18 +111,16 @@ find_package(PkgConfig)
...
-      find_package(Python COMPONENTS Interpreter REQUIRED)
-    endif ()
+    find_package(Python COMPONENTS Interpreter REQUIRED)
 endif ()
 
 # =================================== MPI
 if (USE_MPI)
   get_property(REQUIRED_MPI_COMPONENTS GLOBAL PROPERTY ENABLED_LANGUAGES)
   list(REMOVE_ITEM REQUIRED_MPI_COMPONENTS CUDA)  # CUDA does not have an MPI component
+  if (NOT CMAKE_CROSSCOMPILING)  # when cross compiling, assume the user know what he/she's doing
```
shoshijak: suggestion: use gender-neutral pronoun "they" for the user instead of he/she.
(Reference: https://en.wikipedia.org/wiki/Singular_they)

.cp2k/Makefile
alazzaro: So, take a look at my PR https://github.com/cp2k/cp2k/pull/798
The idea is to use:

https://github.com/cp2k/cp2k/pull/798/files#diff-1f791d555349bad7410338a8fad208ee

I don't really want to change CP2K documentation, so we can keep NVCC in CP2K and ACC in DBCSR. For people they what to try HIP in CP2K (note, it is not documented) they can still use it by setting `NVCC` and `NVFLAGS` (or setting in the arch file and use `override`).

alazzaro: Please let me know if my solution is fine with you...

shoshijak: Yes, this looks good to me!

PR #311, opened 2020-03-10
gha: workflow for testing on macos

dev-zero:


dev-zero:
putting this on hold until our tcopt1 is back in service and I can reproduce this there

hfp:
There are conflicts now and it appears to be stale work. If still desired, one can start over from the closed PR.

PR #312, opened 2020-03-11
removed OpenMP critical section around mp_allocate

hfp:
Removed OpenMP critical section around mp_allocate since it calls MPI_TYPE_SIZE() and MPI_Mem_alloc() both of which are thread-safe (contribution suggested by Dmitri M.).

### DIFF
--- a/src/data/dbcsr_ptr_util.f90
+++ b/src/data/dbcsr_ptr_util.f90
@@ -207,9 +207,7 @@
      IF (mem_type%acc_hostalloc .AND. n > 1) THEN
         CALL acc_hostmem_allocate(mem, n, mem_type%acc_stream)
      ELSE IF (mem_type%mpi .AND. dbcsr_data_allocation%use_mpi_allocator) THEN
-!$OMP critical(allocate)
         CALL mp_allocate(mem, n)
-!$OMP end critical(allocate)
      ELSE
         ALLOCATE (mem(n))
      ENDIF


alazzaro:
Unfortunately, just yesterday I made the MPI allocation off by default, so this change is not tested anymore (I will keep open this PR to add a test with MPI allocation).
Then, I see a problem, but I'm not completely sure: inside the function mp_allocate there is a call to add_perf (https://github.com/cp2k/dbcsr/blob/a6811710a137f5997cacc88593bbd6141007e790/src/mpi/dbcsr_mpiwrap.F#L5769 )

which doesn't seem to be thread-safe... 
I'm actually in favor of removing this call, but it has been a very long time since last time we checked thread-safety in DBCSR... 

Any suggestion? I would keep in hold this PR...

hfp:
Yes, let's keep an eye on it since the critical section was discovered due to performance impact (Dmitry M.). No need to merge it for the ongoing release. There are a number of things to look at:

* MPI allocation may (finally) help with performance, i.e., `I_MPI_SHM_HEAP=1` (disabled by default, I believe) can finally save intra-node traffic when enabled. Either MPI mallocs have to be used or a malloc-proxy must be LD_PRELOADed.
* Unrelated: `use_mempools_cpu` is now `false` by default (as opposed to CP2K 6.1), it may have up to 10% adverse impact (at least for large LS DFTs).

alazzaro:
As a result fo the #316, which removes useless add_perf call, I'm going to merge this PR...

PR #316, opened 2020-03-19
MPI wrapper: remove time measurement where it causes significant overhead

pseewald:


src/mpi/dbcsr_mpiwrap.F
```diff
@@ -962,15 +960,12 @@ SUBROUTINE add_perf(perf_id, count, msg_size)
       IF (.NOT. ASSOCIATED(mp_perf_stack(stack_pointer)%mp_perf_env)) return
 
       mp_perf => mp_perf_stack(stack_pointer)%mp_perf_env%mp_perfs(perf_id)
-      IF (PRESENT(count)) THEN
-         mp_perf%count = mp_perf%count + count
-      END IF
+      mp_perf%count = mp_perf%count + 1
       IF (PRESENT(msg_size)) THEN
```
alazzaro: you don't need this IF anymore, msg_size is not optional

PR #321, opened 2020-03-25
libxsmm_diff to implement array_eq

hfp:
More compact version-checks: revised TO_VERSION macro and provided variants (TO_VERSION2, TO_VERSION3). Introduced libxsmm_diff to implement family of array_eq functions. Prettify.

The reproducer is here [https://github.com/hfp/libxsmm/blob/master/samples/utilities/diff/diff.f](https://github.com/hfp/libxsmm/blob/master/samples/utilities/diff/diff.f). To try:

```bash
cd libxsmm && make
cd samples/utilities/diff && make
./difff
```

Note: the advantage of libxsmm_diff compared to ALL/ANY depends on the Fortran compiler, OS distribution, execution target, and type-size. However, it is consistently better than ALL/ANY.


alazzaro:
I like a lot this function `array_eq` implementation from LIBXSMM.
I wonder if we should replace it everywhere... 

Currently, the main array functions in DBCSR are in `src/core/dbcsr_array_types.F`

hfp:
Honestly, compilers should do a better job for these primitives. I normally would not replace Fortran intrinsics just used in the intended way (use of ALL is absolutely fine). However, it is just that we have it and know how to use it (all ingredients for using libxsmm are there anyway).

hfp:
I can take a look at `src/core/dbcsr_array_types.F`.

alazzaro:
Well, we need a bit of refactoring here....
I think we can merge this PR and probably @pseewald can have a look on how to move the TAS routines in core?

hfp:
Hmm, `dbcsr_tas_unittest` fails with Cray Compiler...

pseewald:
Unfortunately there is no output in case of an abort with cray compiler due to buffering. Using `srun` with the option `-u` fixes this. I guess [this line](https://github.com/cp2k/dbcsr/blob/develop/.ci/daint.cscs.ch/cray.build.sh#L38) should be updated.

alazzaro:
> Unfortunately there is no output in case of an abort with cray compiler due to buffering. Using `srun` with the option `-u` fixes this. I guess [this line](https://github.com/cp2k/dbcsr/blob/develop/.ci/daint.cscs.ch/cray.build.sh#L38) should be updated.

@hfp could you make this change suggested by Patrick? I'm actually rerunning the test...

pseewald:
`dbcsr_tas_unittest` also fails for #320, so it's not your fault, I'll fix this.

alazzaro:
Then I propose to merge here and fix the problem in the appropriate #320 ...
Thanks Hans!

hfp:
The suggested change may look like `-DMPIEXEC_EXECUTABLE="$(if [ "" != "$(command -v srun)" ]; then echo "srun -u"; fi)"`.

PR #322, opened 2020-03-25
avoid array-temporary

hfp:
The temporary array is issued here: [https://github.com/cp2k/dbcsr/blob/develop/src/mm/dbcsr_mm_cannon.F#L1066](https://github.com/cp2k/dbcsr/blob/develop/src/mm/dbcsr_mm_cannon.F#L1066).

### DIFF
--- a/src/dist/dbcsr_dist_util.F
+++ b/src/dist/dbcsr_dist_util.F
@@ -754,15 +754,19 @@ CONTAINS
 
    PURE SUBROUTINE count_bins(nelements, bins, nbins, bin_counts)
       INTEGER, INTENT(IN)                                :: nelements
-      INTEGER, DIMENSION(1:nelements), INTENT(IN)        :: bins
+      INTEGER, DIMENSION(:), INTENT(IN)                  :: bins
       INTEGER, INTENT(IN)                                :: nbins
       INTEGER, DIMENSION(1:nbins), INTENT(OUT)           :: bin_counts
 
-      INTEGER                                            :: el
+      INTEGER                                            :: bin, i, i0, i1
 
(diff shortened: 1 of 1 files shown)


alazzaro:
Good spot. Are you using some code analysis to check such cases?
I'm thinking to add more checks...

hfp:
I believe there are many more such cases. I find them using GNU Fortran debug build. The flags are adopted/according to CP2K/DBCSR like:

```text
-fimplicit-none -fsanitize=leak -fcheck=all -finit-derived -finit-real=snan -finit-integer=-42 -finline-matmul-limit=0 -Werror=aliasing -Werror=ampersand -Werror=c-binding-type -Werror=intrinsics-std -Werror=line-truncation -Werror=tabs -Werror=target-lifetime -Werror=underflow -Werror=unused-but-set-variable -Werror=conversion -Werror=unused-variable -Werror=unused-dummy-argument -Werror=zerotrip -Werror=uninitialized -Wno-maybe-uninitialized -Wuse-without-only -Wintrinsic-shadow
```

Then, running CP2K with some input (GNU Fortran causes some additional/debug console output about array temporaries). Perhaps different workloads produce different temps depending on what code is exercised.

hfp:
This is PR is potentially ready (failure is related to Cray/RPA).

alazzaro:
It is also true that I do use quite the same flags in

https://github.com/cp2k/dbcsr/blob/196ecd057453b8c476a7c5440986435495780dfd/cmake/CompilerConfiguration.cmake#L2

We do run the -O0 coverage version on git CI, is this a possible explanation?

PR #324, opened 2020-03-26
revised calc_norms

hfp:
In calc_norms, ABS was removed and a simd-loop was introduced. In addition, the `schedule(dynamic)` was specified (which is default with GNU Compiler), however the loop in particular is potentially unbalanced due to data-dependent `DATA(bp:bpe)`.

### DIFF
--- a/src/mm/dbcsr_mm_cannon.F
+++ b/src/mm/dbcsr_mm_cannon.F
@@ -1586,7 +1586,7 @@ CONTAINS
             CALL dbcsr_print(right_buffer_calc%mats(v_ki_right, 1), nodata=.TRUE.)
          ENDIF
          !
-         ! form here the code for dbcsr_mm_driver_inner_init was taken
+         ! from here the code for dbcsr_mm_driver_inner_init was taken
          !
          IF (.FALSE.) WRITE (*, *) routineN//" TICK", metronome
          ! Since the right matrix is shifted vertically, the
--- a/src/mm/dbcsr_mm_common.f90
+++ b/src/mm/dbcsr_mm_common.f90
@@ -8,7 +8,7 @@
 !--------------------------------------------------------------------------------------------------!
 
 #:include '../data/dbcsr.fypp'
-#:for n, nametype1, base1, prec1, kind1, type1, dkind1 in inst_params_float
+#:for n, nametype1, base1, prec1, kind1, type1, typesize1, dkind1 in inst_params_float
   SUBROUTINE calc_norms_${nametype1}$ (norms, nblks, &
                                        blki, rbs, cbs, DATA)
      !! Calculates norms of the entire matrix with minimal overhead.
@@ -19,7 +19,7 @@
      ${type1}$, DIMENSION(:), &
         INTENT(IN)                            :: DATA
 
-     INTEGER, PARAMETER                       :: simd = 64 / ${typesize1}$
+     INTEGER, PARAMETER                       :: simd = 64/${typesize1}$
      INTEGER                                  :: i, n, blk, bp, bpe, row, col
      REAL(kind=sp)                            :: val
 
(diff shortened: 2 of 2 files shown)


alazzaro:
Hans, you are making me very happy with this PR...
Could you quantify the performance improvement? 
This part was on my to-do list for many years...

hfp:
Very minor, unfortunately. Starting point was DFT-LS with larger NREP, where calc_norms is up to 6% of the wall time. Anyhow, I tried several implementations incl. OpenMP tasks and different ways to go for SIMD. However, it's (slightly) better than before now. I may also revisit this topic at a later point again.

alazzaro:
I think it is memory-bound (and we use single-precision to store results)... That's why I was surprised about your changes...
The approach I had in mind was a bit more radical: move the pre-calculation of the norms before the multrec and move this part within the multrec to reduce memory traffic... 
It is also true that we do calculate the norms for each rank for each tick of the multiplication...

hfp:
Well, I was a bit attracted by SQRT and wanted to rely on vectorized primitives (SVML). Anyhow, I was actually thinking whether SQRT is necessary at all if we just compare the sum of squares on both sides (I have not looked where the "other" side is).

pseewald:
@hfp I tested the newest DBCSR with CP2K (in #842) and there is an issue with `calc_norms`. Many tests fail with the backtrace below (sdbg build). Could you have a look?

```
Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x7f5af95dbf1f in ???
#1  0x28c54ba in __dbcsr_mm_common_MOD_calc_norms_z._omp_fn.0
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_mm_common.f90:54
#2  0x7f5af9dff481 in GOMP_parallel
	at /opt/cp2k-toolchain/build/gcc-9.2.0/libgomp/parallel.c:171
#3  0x28c6262 in calc_norms_z
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_mm_common.f90:30
#4  0x28c6262 in __dbcsr_mm_common_MOD_calculate_norms
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_mm_common.F:543
#5  0x27becad in __dbcsr_mm_cannon_MOD_multiply_cannon
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_mm_cannon.F:1617
#6  0x27825d8 in __dbcsr_mm_MOD_dbcsr_multiply_generic
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_mm.F:871
#7  0x28c3916 in __dbcsr_multiply_api_MOD_dbcsr_multiply_z
	at /workspace/cp2k/exts/dbcsr/src/mm/dbcsr_multiply_api.F:152
#8  0x281a2ef in __dbcsr_api_MOD_dbcsr_multiply_z
	at /workspace/cp2k/exts/dbcsr/src/dbcsr_api.F:1280
```

alazzaro:
@hfp I can assume your optimizations are valid for single and double real numbers, no for the complex case. I think it would be OK to have a FYPP if statement and provide a fallback for the complex case...

alazzaro:
> Well, I was a bit attracted by SQRT and wanted to rely on vectorized primitives (SVML). Anyhow, I was actually thinking whether SQRT is necessary at all if we just compare the sum of squares on both sides (I have not looked where the "other" side is).

@hfp 
I did an example where I drop the SQRT and do exactly what you are proposing. However, I don't see any gain in the evaluation of the norms, I can assume we are bound by the memory-bandwidth anyway... Any clue?

alazzaro:
OK, I did a bit more investigation and found that we can optimize the code by removing the IF and the SQRT. I've also found that it is better to specify SIMD. See my commit f06fad9ce552003581f870af684b9de8f7ec681e

PR #337, opened 2020-04-16
gha: implement stale issue handling

dev-zero:
needs another feature from the action to be useful:
https://github.com/actions/stale/issues/21


dev-zero:
not going to fly without the bot also removing the `needinfo` label, upstream doesn't want to implement it and I don't have time for it

Issue #212, opened 2019-08-01
C-API improvements


hfp:
See suggestions made on top of PR #209.

PR #348, opened 2020-05-04
Matrix C API

ambmax00:
This PR greatly expands the C API for matrix operations.


ambmax00:
I have added two new functions to dbcsr_api: `dbcsr_nblkrows_local` and `dbcsr_nblkcols_local`, which were necessary for the C interface. I added them temporarily, as you probably want to add them yourself in a separate commit, and then I will rebase.

alazzaro:
retest this please

alazzaro:
@ambmax00 There is a problem with the CCE compilation.

alazzaro:
@ambmax00 I gave a fast look at your changes, you did a tremendous work! Thanks! Definitely you are the C-API DBCSR expert ;)
Since you are there, I would like to ask you if you can include some of the suggestions proposed at https://github.com/cp2k/dbcsr/issues/212 (notify @hfp ) Mainly the idea it to replace the void* with some typedef...

ambmax00:
Thanks! :)

> Mainly the idea it to replace the void* with some typedef...

Ok, I will take care of introducing some typedefs for it, @hfp.
If possible, I would like to keep this PR open for a bit, so I can start using it in my program, and fix issues that may come up.

I have not introduced any c bindings for the csr functions. Are they used often?
I'm also not so sure about what exactly the 'work' functions do, like `dbcsr_work_create`, `dbcsr_verify_matrix`, `dbcsr_add_work_coordinate`, `dbcsr_get_wms_data_p`, or `dbcsr_work_size`, so I have not yet included them.

alazzaro:
> Thanks! :)
> 
> > Mainly the idea it to replace the void* with some typedef...
> 
> Ok, I will take care of introducing some typedefs for it, @hfp.
> If possible, I would like to keep this PR open for a bit, so I can start using it in my program, and fix issues that may come up.

No problem (next time you can name the PR with WIP). This PR will go >=v2.2 anyway...

> 
> I have not introduced any c bindings for the csr functions. Are they used often?
> I'm also not so sure about what exactly the 'work' functions do, like `dbcsr_work_create`, `dbcsr_verify_matrix`, `dbcsr_add_work_coordinate`, `dbcsr_get_wms_data_p`, or `dbcsr_work_size`, so I have not yet included them.

CSR functions are not that used. The idea is to send data to other libraries with CSR format (It think @pseewald did them). For the workspace, these are internal functions used in DBCSR to handle threads. In practice, whenever you want to use threads, you can ask DBCSR to create a workspace per each thread (so that you don't get any conflict by simultaneous changes to data). Then we do finalize the data and merge all the workspaces. I think dbcsr_work_create is fine to add, no sure if the other functions are useful in general...

One note, at this point we should consider that there was years ago a PR on the C-API (see https://github.com/cp2k/dbcsr/pull/64). For some reason of priority, that PR never got merged...
I'm asking you as the expert here, do you think it makes sense to resume it? (I mean, merging with your current changes)

ambmax00:
Ok, I will leave out the csr procedures for now.
Another question I have: what is a good way to compute linear algebra problems with dbcsr, like eigenvalue problems or cholesky decomposition? Use ScaLAPACK? At the moment, I'm just collecting the matrix on one node, and then I use the C++ Eigen library... But that won't scale well later. 
> I'm asking you as the expert here, do you think it makes sense to resume it?

I have seen the PR. I would personally do a lot differently with the C++ classes. In my opinion, we should only give a low-level C interface to the user, who can then incorporate it into classes if they want. I'm currently also using my own C++ wrapper classes. At any rate, it's be better to merge it in a different PR.

alazzaro:
> Ok, I will leave out the csr procedures for now.
> Another question I have: what is a good way to compute linear algebra problems with dbcsr, like eigenvalue problems or cholesky decomposition? Use ScaLAPACK? At the moment, I'm just collecting the matrix on one node, and then I use the C++ Eigen library... But that won't scale well later.

Well, I assume you want to do dense algebra. Now, you can set in DBCSR the same block-cycling distribution as it is in SCALAPACK, then you don't need to collect on a single node and you can directly call SCALAPACK (or whatever it is compatible). In principle, we have functions for doing DBCSR to SCALAPACK format. We can discuss this in a different PR.


> 
> > I'm asking you as the expert here, do you think it makes sense to resume it?
> 
> I have seen the PR. I would personally do a lot differently with the C++ classes. In my opinion, we should only give a low-level C interface to the user, who can then incorporate it into classes if they want. I'm currently also using my own C++ wrapper classes. At any rate, it's be better to merge it in a different PR.

Yes, sure. The point is if it is still convenient to work on that PR (it is quite old now). I understand your comments on the C++ API, and I would say I agree with you... Anyway, yes, this is a topic for a different PR...

ambmax00:
> Ok I see. I'd prefer to enable disk I/O as a feature on the Fortran side and to keep the C API in sync with the Fortran API. 

I can gladly help with setting up disk i/o in dbcsr later down the road, but I cannot do it right now

> However I leave it up to you whether this is a priority - if not I think direct data access is a reasonable compromise.

Direct data access would be the best for the moment. I will also remove the routine for accessing tensor blocks by pointers - it actually does not give any considerable speed-up.

pseewald:
> Direct data access would be the best for the moment. I will also remove the routine for accessing tensor blocks by pointers - it actually does not give any considerable speed-up.

I can confirm this - for less sparse or dense tensors this was a problem in the past but I optimized this recently and the default way of accessing blocks should not add any overhead now.

alazzaro:
@ambmax00 
We are going to make an official DBCSR release in two or three weeks for the new CP2K release.
Do you what to have your PR merged?

ambmax00:
I would prefer to wait for the next release - until then I will have a working version of my program.

ambmax00:
Ready for merging. I originally wanted to wait for #435 to be resolved, but I will finish up the matrix API anyway.

hfp:
> Ready for merging. I originally wanted to wait for #435 to be resolved, but I will finish up the matrix API anyway.

ACK, but let's wait for @alazzaro. The interface can a useful starting point for @oschuett's Proof Of Concept (POC) tensor library.

hfp:
Btw, Daint is perhaps under reconstruction or at least CI is not operable.

PR #358, opened 2020-07-20
C code/header format and cleanup

hfp:
Removed tabs and trailing white spaces. Rely on C comments in case of header files. Indentation and other format properties are still inconsistent wrt C/C++ code depending on author/maintainer/editor preferences/settings. Note it might be an option to incorporate Clang format into "local hooks" running for every commit. A starting point with Clang format might be https://github.com/hfp/libxsmm/blob/master/.clang-format and https://github.com/hfp/libxsmm/blob/master/.github/pre-commit).


oschuett:
Yes, since we keep adding C code, we should definitely start using a formatter and clang-format would have been my choice too.

Do think we could just use the default style (llvm)?

hfp:
> Do think we could just use the default style (llvm)?

It is a matter of taste which format we want to adopt. Everybody must adjust the own taste or what´s accepted because clang-format does not allow to control every aspect. I would personally setup a format and tune it over some time and finally deploy it. A good starting point might be [https://zed0.co.uk/clang-format-configurator/](https://zed0.co.uk/clang-format-configurator/) (it is not fully bug-free as it does not load all .clang-format files, etc). Regarding llvm style, I believe it already starts with indentation level and ends with more specific format which might be or may not be acceptable for one or the other person.

I think one may apply for instance [https://github.com/hfp/libxsmm/blob/master/.clang-format](https://github.com/hfp/libxsmm/blob/master/.clang-format) for a set of C/C++ files and tune the format until acceptable. Of course, the reformatted code might be compared with the default format-policy (llvm).

oschuett:
> It is a matter of taste which format we want to adopt.

Indeed and I was hoping that we could avoid this discussion by simply not having a `.clang-format` config. For Python we're already following this approach by using `black`.

dev-zero:
> > It is a matter of taste which format we want to adopt.
> 
> Indeed and I was hoping that we could avoid this discussion by simply not having a `.clang-format` config. For Python we're already following this approach by using `black`.

Python has an official style guide (`PEP 8`) which we follow by using `black`.

Btw, about formatting changes: if someone has some spare time, then it would be nice to have a `.git-blame-ignore-revs` file listing all the commits with only formatting changes such that one can ignore them with `git blame --ignore-revs-file .git-blame-ignore-revs`

hfp:
> simply not having a .clang-format config
> Python has an official style guide

I can concatenate the existing C/C++ code and reformat it with default settings and perhaps with an initial/tuned setup. Perhaps @shoshijak and others can "vote" (I do not believe this is an exercise in democracy but maybe it is less artificial to decide).

Using defaults has an upfront bonus since nothing must be configured. Using `.clang-format` file is actually by applying `--style=file` as command line flag (location of the dot-file is derived automatically if such file is part of the file-tree). On the other hand, calling clang-format as part of a commit, checking sufficient version of that tool etc. requires a "non-zero" infrastructure regardless of whether a `.clang-format` file is used on top.

Do we have an infrastructure in DBCSR to reformat changes when committed? Though, reformatting code must happen on the client (like `make pretty`) or be rejected on the server-side (during CI/tests). In LIBXSMM, I am using the former (reformat during commit based on hooks). What´s the preferred way?

dev-zero:
> Do we have an infrastructure in DBCSR to reformat changes when committed? Though, reformatting code must happen on the client (like `make pretty`) or be rejected on the server-side (during CI/tests). In LIBXSMM, I am using the former (reformat during commit based on hooks). What´s the preferred way?

We have it actually for both DBCSR and CP2K via [pre-commit](https://pre-commit.com/), [instructions](https://github.com/cp2k/cp2k/wiki/Git-Tips-&-Tricks#git-pre-commit-hooks).
The CIs run them also via pre-commit to ensure we have the same result everywhere.

PR #360, opened 2020-08-07
Minor fixes to libsmm_acc/transpose

shoshijak:


oschuett:
The change in https://github.com/cp2k/dbcsr/pull/360/commits/9bdf8f0a702d9519fd273fb56d13aef334e0007b seems like it's fixing an actual bug. Could that be the root cause for https://github.com/cp2k/cp2k/issues/893 or https://github.com/cp2k/cp2k/issues/911?

shoshijak:
> The change in 9bdf8f0 seems like it's fixing an actual bug. Could that be the root cause for cp2k/cp2k#893 or cp2k/cp2k#911?

No, because the changes in 9bdf8f0 are just fixing a previous commit of this same PR, namely f6b9b4174fda5e528aadf2981c00ff24a81378c7. 
Upon reflection, I should have squashed these 2 commits.

alazzaro:
@oschuett
The two problems are very hidden... I did several testings to try to make a small reproducer. So far I found that there is a problem with threading communicating to the GPU only for a specific GPU kernel. I hope it is not a problem on the CUDA code itself...
Next step is to use thread-checker and see if we are lucky... The problem is that there are several false-positive... I hope to have time in September to finalize it.

oschuett:
> The two problems are very hidden... 

@alazzaro thanks for keeping on it! Yes, race conditions as a real pain. Could you narrow down the commit range?

shoshijak:
Merging this, since the one check that does not pass is due to a failed `Upload coverage info to codecov`.

PR #381, opened 2020-10-01
Refactor acc blas

shoshijak:
Until now, cu/hipblas was handled by DBCSR, while libsmm_acc (DBCSR's GPU backend) was completely separate from cu/hipblas. 

This PR makes cu/hipblas the responsibility of libsmm_acc.

- This is an improvement from the point of view of separation of concerns: libsmm_acc now handles all of the GPU acceleration, including cu/hipblas. This means that libsmm_acc can handle deciding whether to multiply a block using its custom kernels or cu/hipblas, instead of leaving this decision up to DBCSR's Fortran code as is currently the case.
- Removes the additional round trip to Fortran in dbcsr_acc_operations.F. Now, libsmm_acc simply calls cu/hipblas directly.
- LOC is reduced, because we removed `dbcsr/src/acc/cublaswrap/` and `dbcsr/src/acc/hipblaswrap/`


shoshijak:
The ROCm test that doesn't pass here in the CI, does pass when I use cmake with the Makefile generator rather than the Ninja generator. Looks like this is due to Ninja handling relative paths differently than Make does. 

@dev-zero is it crucial that we can build with Ninja, or could Make be enough?

src/acc/libsmm_acc/CMakeLists.txt
```diff
@@ -62,8 +62,10 @@ else () # i.e. USE_HIP
   set_source_files_properties(${LIBSMM_ACC_SRC_FILES} PROPERTIES HIP_SOURCE_PROPERTY_FORMAT 1)
   # hip_add_library does not support OBJECT libraries. For this reason, we are forced to use a static library
   hip_add_library(libsmm_acc STATIC ${LIBSMM_ACC_FILES})
+  target_include_directories(libsmm_acc PUBLIC ${HIP_PATH}/../include)
```
dev-zero, line 65: does this have to be `PUBLIC`?

src/mm/dbcsr_acc_operations.F
```diff
@@ -226,9 +170,9 @@ SUBROUTINE dbcsr_acc_transpose(trs_stack, offset, nblks, datatype, buffer, m, n,
       IF (careful_mod) CALL timeset(routineN, error_handle)
       istat = 0
 
-      ! Limits the call to small kernels
-      IF (m .LE. max_kernel_dim .AND. &
-          n .LE. max_kernel_dim) THEN
+      ! Call batched in-place transpose in libsmm_acc
+      IF (m .LE. 80 .AND. &
+          n .LE. 80) THEN
```
dev-zero, line 175: here we are losing documentation and introduce a magic constant, maybe introduce a `parameter` for that and document it?

dev-zero:
only some minor comments, otherwise LGTM, thanks!

src/acc/libsmm_acc/CMakeLists.txt
shoshijak: I just tested with `PRIVATE` and `INTERFACE`, both of which fail. So, yes.

src/mm/dbcsr_acc_operations.F
shoshijak: Good catch, I'll fix this ...

src/acc/libsmm_acc/CMakeLists.txt
dev-zero: Interesting, was it the examples/tests or dbcsr itself?

shoshijak: the compilation fails in the tests, yes

dev-zero: I see. For the tests we leak the CUDA/HIP headers through `libsmm_acc.h`, but as far as I can see that tests anyway bypass DBCSR and directly accesses `libsmm_acc`, hence the requirement for the CUDA/HIP headers.
For CUDA it seems we used the following in `tests/CMakeLists.txt` instead of exposing the include directory via a `PUBLIC` attribute:

```
      target_include_directories(
        ${libsmm_acc_test} PRIVATE ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
```

In the end, I am not sure whether it is worthwhile to follow this scheme to reduce the number of external include directories/libraries a consumer of the DBCSR library sees (what happens when we add a dependency with the `PUBLIC` attribute) or whether I am overly conscious about this stuff.

src/mm/dbcsr_acc_operations.F
shoshijak: fixed in commit 7a90508

src/acc/libsmm_acc/CMakeLists.txt
shoshijak: No, I don't think you're being overly conscious, what you say makes a lot of sense. 
I just haven't been putting as much time and effort in the HIP backend's cmake, because HIP changes quite a bit with each new ROCm release, and still misses some cmake features important to us (e.g. support for object libraries). But I'll go ahead and fix what you pointed out, let's not let this get too messy ;)

PR #385, opened 2020-10-15
Feature: transpose blocks of dim>80 with cu/hipblas

shoshijak:
Up to now, DBCSR's GPU backend did not support the transposition of matrix blocks with dimension >80. This PR adds support for this via cu/hipblas.

### DIFF
--- a/src/acc/acc_libsmm.h
+++ b/src/acc/acc_libsmm.h
@@ -31,12 +31,13 @@ typedef enum libsmm_acc_data_t {
 int libsmm_acc_init(void);
 acc_bool_t libsmm_acc_is_thread_safe(void);
 
-int libsmm_acc_transpose(const int* dev_trs_stack, int offset, int stack_size,
-  void* dev_data, libsmm_acc_data_t datatype, int m, int n, int max_kernel_dim, void* stream);
+int libsmm_acc_transpose(const int* host_trs_stack, const int* dev_trs_stack, int offset,
+  int stack_size, void* dev_data, libsmm_acc_data_t datatype, int m, int n, int max_kernel_dim,
+  void* stream);
 
 int libsmm_acc_process(const int* host_param_stack, const int* dev_param_stack, int stack_size,
   int nparams, libsmm_acc_data_t datatype, const void* dev_a_data, const void* dev_b_data, void* dev_c_data,
-  int m_max, int n_max, int k_max, int max_kernel_dim, acc_bool_t def_mnk, void* stack_stream, void* c_stream);
+  int m_max, int n_max, int k_max, acc_bool_t def_mnk, void* stack_stream, void* c_stream);
 
 #if defined(__cplusplus)
 }
--- a/src/acc/cuda/acc_blas.h
+++ b/src/acc/cuda/acc_blas.h
@@ -32,4 +32,8 @@ int acc_blas_dgemm(ACC_BLAS(Handle_t) *handle, char transa, char transb,
                    const double *a_data, const double *b_data, double *c_data,
                    double alpha, double beta, ACC(Stream_t) *stream);
 
+/****************************************************************************/
+int acc_blas_transpose(ACC_BLAS(Handle_t) *handle, int m, int n, int offset,
+                       double *data, ACC(Stream_t) *stream);
+
 #endif /* ACC_BLAS_H */
--- a/src/acc/cuda/acc_cublas.cu
+++ b/src/acc/cuda/acc_cublas.cu
@@ -56,3 +56,31 @@ int acc_blas_dgemm(ACC_BLAS(Handle_t) *handle, char transa, char transb,
 
   return(0);
 }
+
+/****************************************************************************/
+int acc_blas_transpose(ACC_BLAS(Handle_t) *handle, int m, int n, int offset,
+                       double *data, ACC(Stream_t) *stream)
+{
+
+  ACC_BLAS_CALL(SetStream, (*handle, *stream));
+
+  const double alpha = 1.;
+  const double beta = 0.;
+
+  // cu/hip blas does not have in-place transpose:
+  // we duplicate the matrix into a temporary buffer before transposing
+  double *data_temp;
+  ACC_API_CALL(Malloc, (&data_temp, m * n * sizeof(double)));
+  ACC_API_CALL(Memcpy, (data, data_temp, m * n * sizeof(double), ACC(MemcpyDeviceToDevice)));
+
+  ACC_BLAS_CALL(Dgeam, (*handle, ACC_BLAS_OP_T, ACC_BLAS_OP_N,
+				m, n,
+				&alpha, data_temp, n,
+				&beta, data_temp, m,
+                &data[offset], m));
+
+  // free temporary buffer
+  ACC_API_CALL(Free, (data_temp));
+
+  return(0);
+}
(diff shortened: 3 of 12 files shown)


alazzaro:
Does it assume that we can run multiplications of blocks of dim>80?

shoshijak:
We can run multiplications of blocks of dim > 80 via hip/cublas.

alazzaro:
> We can run multiplications of blocks of dim > 80 via hip/cublas.

OK, I see. I know, I can look at the code, but I'm lazy and I will ask you ;) : is the transposition done in the cUBLAS call or we run our transpose kernel? Originally, the code was running the former case...

shoshijak:
> OK, I see. I know, I can look at the code, but I'm lazy and I will ask you ;) : is the transposition done in the cUBLAS call or we run our transpose kernel? Originally, the code was running the former case...

I'm sorry, I'm not sure I understand the question 🙈
- DBCSR's custom [transpose kernel](https://github.com/cp2k/dbcsr/blob/develop/src/acc/libsmm_acc/kernels/smm_acc_transpose.h) loads the entire matrix into shared memory, which is why it cannot transpose large matrices
- the new way of transposing large matrices is done via a cublas call

alazzaro:
Yeap, the second bullet gave me the answer. Thanks. The original code was already doing it.

shoshijak:
> The original code was already doing it.

The code before this PR was already transposing matrix blocks > 80? Where?

alazzaro:
Sorry, this was a while back ago. This is the line:

https://github.com/cp2k/dbcsr/blob/d2dfe6e25efde255f72ac27f542124b69c7af1c2/src/mm/dbcsr_acc_operations.F#L150-L186

Your solution is better...

hfp:
Is transposing larger blocks necessary for practical inputs or can we enable this to be useful? Btw, I found batch-transposing matrices with small shapes is not beneficial on GPU when compared to reasonable multicore CPUs (even when omitting PCIe transfer). I am not comparing energy to solution but rather execution time (raw performance). I played with batches aka "stacks" of 30k matrices. Let me rephrase, it does not seem to be able to leverage the memory bw speedup when comparing kernel execution time (even if data transfer is not included). In turn, if the pre-transposed data is needed by a subsequent kernel on the GPU (let us say SMMs), it means that transposes are eating into the speedup leveraged by such kernel. Maybe I am doing something wrong...

alazzaro:
@shoshijak I went through this PR. I have some comments:
1. I see that the cublas kernel is called only when there are homogenous stacks, i.e.

https://github.com/cp2k/dbcsr/blob/8e76ca107a595916811d0260f7956d3f32f5c33d/src/acc/libsmm_acc/libsmm_acc.cpp#L303-L304

In the previous version of the code, no uniform stacks were sent to the Cublas execution. Should we consider that also now?

2. I don't think we need a transpose function for the big-blocks. Cublas-dgemm can already handle the T case, so we can skip a specific call for that. This was used in the previous code I linked above. Of course, assuming we do run inhomogenous stacks, the check if it is N or T for the B blocks has to be considered for each block of the stack.

shoshijak:
> I see that the cublas kernel is called only when there are homogenous stacks, i.e.

Ah, that's a good point ... should we send to cublas in the case of inhomogeneous stacks, in your opinion? 

> In the previous version of the code, no uniform stacks were sent to the Cublas execution. Should we consider that also now?

By "uniform", do you mean "homogeneous" ? I thought homogeneous stacks >80 (big blocks) were also sent to cublas, in the earlier version of the code? 

> I don't think we need a transpose function for the big-blocks. Cublas-dgemm can already handle the T case, so we can skip a specific call for that. This was used in the previous code I linked above. Of course, assuming we do run inhomogenous stacks, the check if it is N or T for the B blocks has to be considered for each block of the stack.

Ah, I see ... so are you saying that as long as the cublas multiply handles both N and T cases (like [here](https://github.com/cp2k/dbcsr/blob/8e76ca107a595916811d0260f7956d3f32f5c33d/src/acc/libsmm_acc/libsmm_acc.cpp#L235-L238)), and as long as we do not process inhomogeneous stacks with cublas, we do not need a cublas transpose?

alazzaro:
> > I see that the cublas kernel is called only when there are homogenous stacks, i.e.
> 
> Ah, that's a good point ... should we send to cublas in the case of inhomogeneous stacks, in your opinion?
> 
> > In the previous version of the code, no uniform stacks were sent to the Cublas execution. Should we consider that also now?
> 
> By "uniform", do you mean "homogeneous" ? I thought homogeneous stacks >80 (big blocks) were also sent to cublas, in the earlier version of the code?

Well, at this point I'm not sure about the old code anymore 😄  ... Anyway, likely the inhomogeneous are not worth running with cublas, so let's keep them out. So I agree that only the homogeneous stacks with big size should be sent to the GPU.

> 
> > I don't think we need a transpose function for the big-blocks. Cublas-dgemm can already handle the T case, so we can skip a specific call for that. This was used in the previous code I linked above. Of course, assuming we do run inhomogenous stacks, the check if it is N or T for the B blocks has to be considered for each block of the stack.
> 
> Ah, I see ... so are you saying that as long as the cublas multiply handles both N and T cases (like [here](https://github.com/cp2k/dbcsr/blob/8e76ca107a595916811d0260f7956d3f32f5c33d/src/acc/libsmm_acc/libsmm_acc.cpp#L235-L238)), and as long as we do not process inhomogeneous stacks with cublas, we do not need a cublas transpose?

Exactly, then this PR can be simplified a lot to a single IF for checking the size of the B blocks...

shoshijak:
On another hand, we might come to want to multiply bigger blocks with libsmm_acc rather than with CUBLAS in the near future. 
On architectures such as NVIDIA A100, DP tensor cores might be advantageous for running libsmm_acc with larger block sizes. 
And in Patrick's work on tensor contractions has made block sizes of, for example, 100, relevant.

Do you see an issue in keeping this possibility of transposing larger blocks via cublas? If not, I'd be in favor of merging this (fixes pending) ..

alazzaro:
OK, I see your point... This is for future extension of libsmm_acc...

So, in principle we can "increase" the max size of the blocks (currently 80) for libsmm_acc and gain there. However, cusmm transpose kernel cannot be bigger than 80? In this respect, we will have three max sizes then:
1. for multiplications <80 use the current cusmm
2. >=80 and < of a new given threshold, use special new kernels, otherwise fallback to cublas
3. for transpose, tight threshold at 80, fallback to cublas

Is this what you are suggesting?

Personally, as Hans pointed out, transposing for large kernels probably will not be fast unless we can use a batched version. Have you timed it? I mean, we can make a test where we call transpose and then cublas NN, or we directly call cublas NT...

At this point, I will keep this PR for further investigation after v2.1 and consider the easy change of enable T in cublas for small B kernels...

alazzaro:
@shoshijak just in the case you are waiting for the CSCS test, today Daint is offline...

shoshijak:
@alazzaro thanks, I know 👍
I want to give this PR a little more thought and am trying to finish up my docs PR in the meantime ...

alazzaro:
I'm putting this PR for v2.2, but let me know if you have time to merge for the v2.1 (I would like to have a stable release for CP2K 8.1, expected by the end of the year).

hfp:
Let us close this PR. It has conflicts now and it is stale work.

Issue #384, opened 2020-10-14
dbcsr_tensor_unittest fails with CCE 10


shoshijak:
Since about around PR #370 , the CI test `dbcsr_tensor_unittest` fails with a segfault, when compiled with the Cray compiler. 

I've run tests and could determine that the fail is caused by CSCS upgrading the Cray compiler (in terms of module names, from `cce/9.0.2` to `cce/10.0.2`), and not by a change in DBCSR's code base.

alazzaro:
Good spot, I will fix it. Thanks!

PR #390, opened 2020-11-23
Revised ACC interface

hfp:
Introduced enumeration libsmm_acc_data_t collecting macro symbols like dbcsr_type_real_8 (LIBSMM/ACC interface); this is for reusing the same definitions in every backend implementation (centralized type definition). Cleanup ACC/main interface. Prevent compilation warning about non-const string literal (CUDA/HIP header). Introduced NO_DBCSR_TIMESET to get rid of circular dependency (simplifies Makefile for upcoming changes/sample/driver code).

Note: CUDA/HIP build may need revisit; currently the ACC/CUDA backend depends on LIBSMM/ACC due to initialization code (previously the dependency edge was supposedly limited to the other way around: LIBSMM uses backend). Further, CUDA/HIP backend now not only depends on LIBSMM/ACC but also on DBCSR main library due to timeset functionality.

### DIFF
--- a/src/acc/acc.h
+++ b/src/acc/acc.h
@@ -16,17 +16,7 @@ extern "C" {
 #endif
 
 /** types */
-typedef void acc_stream_t;
-typedef void acc_event_t;
 typedef int acc_bool_t;
-
-typedef enum acc_data_t {
-  ACC_DATA_F64 = 3,
-  ACC_DATA_F32 = 1,
-  ACC_DATA_C64 = 7,
-  ACC_DATA_C32 = 5,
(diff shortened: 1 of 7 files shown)


hfp:
GPU implementation seems to be [broken](https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-579/cray.test.out) (TAS/tensor test). Note, this PR/code does not change the implementation and is unrelated.

alazzaro:
Problem related to https://github.com/cp2k/dbcsr/issues/384
Please ignore it, for the time being...

PR #391, opened 2020-11-23
Stand-alone benchmarks and test for ACC/LIBSMM interface

hfp:
Added some test code (carry-forward from OpenMP/backend experiments), which exercises the ACC interface in an abstract fashion (backend agnostic); building this test may be integrated later with CMake infrastructure. Added two benchmark drivers (matrix transpose and multiplication). Added Makefile to build test and driver in a stand-alone fashion using the CUDA based backend (please note, the Makefile is not as sophisticated as the CMake based tool chain, however, it is meant to be simple or easy to adjust for the local environment).

### DIFF
--- a/src/acc/README.md
+++ b/src/acc/README.md
@@ -1,0 +1,22 @@
+# ACCerator Interfaces
+
+## Overview
+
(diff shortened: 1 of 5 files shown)


hfp:
Documentation: [https://github.com/cp2k/dbcsr/tree/develop/src/acc](https://github.com/cp2k/dbcsr/tree/develop/src/acc#accelerator-interfaces).

dev-zero:
@hfp would it make sense to move the docs here: https://github.com/cp2k/dbcsr/tree/develop/docs/guide ?

hfp:
If you think it is worth. So far, it "documents" the content of the directory (src/acc). Let me prove it more useful, and add another backend (I can move/extend this documentation as part of such PR). This documentation is so far "developer focused" as it uses the ACC interfaces (as if they provide an own library). With the current status, the documentation contributes very little "need to know" wrt DBCSR's usage.

Issue #361, opened 2020-08-14
Add -Wconversion-extra to the warning flags


mkrack:
Conversion warning:
```
/home/krack/github/mkrack/cp2k/exts/dbcsr/src/base/dbcsr_machine.F:149:11:

  149 |       wt = 0.0 ! fallback default
      |           1
Warning: Conversion from ‘REAL(4)’ to ‘REAL(8)’ at (1) [-Wconversion-extra]

/home/krack/github/mkrack/cp2k/exts/dbcsr/src/base/dbcsr_machine_posix.f90:130:41:

  130 |         mem_local = mem_local*getpagesize()
      |                                         1
Warning: Conversion from INTEGER(4) to INTEGER(8) at (1) [-Wconversion-extra]
```
Unused function?:
```
/home/krack/github/mkrack/cp2k/exts/dbcsr/src/acc/dbcsr_acc_devmem.F:330:0:

  330 |    SUBROUTINE acc_devmem_info(free, avail)
      | 
Warning: ‘acc_devmem_info’ defined but not used [-Wunused-function]
```
Getting rid of unused parameter routineP
```
/home/krack/github/mkrack/cp2k/exts/dbcsr/src/acc/dbcsr_acc_hostmem.F:217:45:

  217 |                                      routineP = moduleN//':'//routineN
      |                                             1
Warning: Unused parameter ‘routinep’ declared at (1) [-Wunused-parameter]
/home/krack/github/mkrack/cp2k/exts/dbcsr/src/acc/dbcsr_acc_hostmem.F:198:45:

  198 |                                      routineP = moduleN//':'//routineN
      |                                             1
Warning: Unused parameter ‘routinep’ declared at (1) [-Wunused-parameter]
/home/krack/github/mkrack/cp2k/exts/dbcsr/src/acc/dbcsr_acc_hostmem.F:217:45:

  217 |                                      routineP = moduleN//':'//routineN
      |                                             1
Warning: Unused parameter ‘routinep’ declared at (1) [-Wunused-parameter]
/home/krack/github/mkrack/cp2k/exts/dbcsr/src/acc/dbcsr_acc_hostmem.F:198:45:

  198 |                                      routineP = moduleN//':'//routineN
      |                                             1
Warning: Unused parameter ‘routinep’ declared at (1) [-Wunused-parameter]
/home/krack/github/mkrack/cp2k/exts/dbcsr/src/acc/dbcsr_acc_hostmem.F:217:45:

[...]

PR #392, opened 2020-11-24
Cleanup unused parameters, vars, procedures, interfaces

dev-zero:
- mpi: generate interface decls based on nametype1 whenever possible
- mpi: drop unused/unexported functions, respectively export some of them (for consistency)

intended to solve parts (or maybe all) of #361


dev-zero:
commit d82360c could maybe be cherry-picked into CP2K's mpi wrapper

dev-zero:
@pseewald the commit you might interested in is `tas & tensors: drop unused (and private) functions`
@alazzaro for you starting from the `mpi: ...` commits

I took now the rather extreme approach to drop subroutines as much as possible instead of making them public. The problem of keeping dead code is that it rots (even simple functions like an iterator reset could break if people add more state to the iterator and forget to update the reset method because it is not tested), and people who start using it may not realize that they build on a possible source of error. On the other hand are we dropping code someone wrote intentionally and people will likely not grep the git history to check whether a function they need was there at some point in the past and that they could simply revive it.
The best would probably be to add some tests (especially for the utility classes), but that takes more time.
For the latter one could instead comment out the code or only leave the commented out function signature with a note, pointing people to use the VCS to recover the lost code.

pseewald (approved):

cmake/CompilerConfiguration.cmake
```diff
@@ -5,7 +5,7 @@ if (CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
   endif ()
   set(CMAKE_Fortran_FLAGS_RELEASE  "-O3 -g -funroll-loops")
   set(CMAKE_Fortran_FLAGS_COVERAGE "-O0 -g --coverage -fno-omit-frame-pointer -fcheck=all -ffpe-trap=invalid,zero,overflow -fbacktrace -finit-real=snan -finit-integer=-42 -finit-derived -Werror=realloc-lhs -finline-matmul-limit=0 -Werror")
-  set(CMAKE_Fortran_FLAGS_DEBUG    "-O2 -ggdb -fno-omit-frame-pointer -fcheck=all -ffpe-trap=invalid,zero,overflow -fbacktrace -finit-real=snan -finit-integer=-42 -finit-derived -Werror=realloc-lhs -finline-matmul-limit=0 -fsanitize=undefined -fsanitize=address -fsanitize-recover=all -Warray-temporaries")
+  set(CMAKE_Fortran_FLAGS_DEBUG    "-O2 -ggdb -fno-omit-frame-pointer -fcheck=all -ffpe-trap=invalid,zero,overflow -fbacktrace -finit-real=snan -finit-integer=-42 -finit-derived -finline-matmul-limit=0 -fsanitize=undefined -fsanitize=address -fsanitize-recover=all -Wall -Wextra -Werror -Werror=realloc-lhs -Wno-error=array-temporaries -Wno-error=compare-reals -Wno-error=function-elimination -Wno-error=surprising")
```
alazzaro, line 8: More warnings, good...

alazzaro:
Nice, very nice, it is an incredible clean-up... Only a single (minor) comment... Well done!

PR #394, opened 2020-11-25
feature/cuda profiling flag

dev-zero:


shoshijak:
I guess we should add the new flag `WITH_CUDA_PROFILING` to the [installation documentation](https://github.com/cp2k/dbcsr/blob/develop/docs/guide/2-user-guide/1-installation/1-install.md), right?

dev-zero:
@shoshijak yes, thanks a lot for the reminder

alazzaro:
This is OK. It is actually an internal "very useful" feature that helps a lot to check GPU and CPU overlaps when providing with nvcc... Probably needs more documentation. I keep an issue open as a reminder https://github.com/cp2k/dbcsr/issues/397

PR #395, opened 2020-11-25
bugfix/explicit fypp python

dev-zero:


alazzaro:
Should we update FYPP at well?

dev-zero:
> Should we update FYPP at well?

we are already on the latest (3.0), and Aradì didn't answer on my PR to change the shebang in fypp itelf. Hence the explicit `$(PYTHON) $(FYPPEXE)` call.

PR #396, opened 2020-11-26
Bugfix/scale by vector

dev-zero:


dev-zero:
* Currently I am testing only square sizes due to my inability to setup the `dbcsr_transposed` in a way that it works (always get `"Given distribution must be compatible with the current distribution"`).
* Despite the fix the test occasionally fails

dev-zero:
ok, the tests verified the issue

alazzaro (approved):

PR #399, opened 2020-12-01
[WIP] bugfix/issue 361

dev-zero:


alazzaro:
Wow!

src/mm/dbcsr_mm.F
```diff
@@ -641,7 +641,7 @@ SUBROUTINE dbcsr_multiply_generic(transa, transb, &
       ! update statistics (we count marketing flops per MPI rank)
       dbcsr_mpi_statistics%last_mpi_ranks_used = numnodes
       marketing_flops = marketing_flops + &
-                        (2.0*(l_row - f_row + 1.0)*(l_col - f_col + 1.0)/numnodes)*(l_k - f_k + 1.0)
+                        REAL(2*(l_row - f_row + 1)*(l_col - f_col + 1))/REAL(numnodes)*REAL(l_k - f_k + 1)
```
dev-zero, line 644: do we have a risk of int overflow here? The previous code did the multiplication using real, while I prefer to use int and then convert

dev-zero:
there are actually quiet a few places where we possibly could have overflown before doing the conversion to a larger int

src/mm/dbcsr_mm.F
alazzaro: There was an overflow here time ago and that's why I changed to real. This is particularly true for long simulations.
This "marketing" flops is there only for statistics, it is not really useful. I like your change...

src/ops/dbcsr_io.F
```diff
@@ -846,19 +848,19 @@ SUBROUTINE dbcsr_binary_read(filepath, distribution, groupid, matrix_new)
...
       ! number of data chunks from sub-header 2 and 3 to be read by every node rounded up
       ! to the next integer to make it even for all the nodes in the specified mpi group
-      share_size = CEILING(REAL(size_of_pgrid, KIND=dp)/group_list_size)
+      share_size = CEILING(REAL(size_of_pgrid, KIND=dp)/REAL(group_list_size, KIND=dp))
 
       ALLOCATE (subh2_offsets(share_size))
       subh2_offsets = BOF
       DO i = 1, share_size
-         offset = subh2_start + mpi_integer_size*2*(worker_id + (i - 1)*group_list_size)
+         offset = subh2_start + mpi_integer_size*2*INT(worker_id + (i - 1)*group_list_size, KIND=file_offset)
```
dev-zero, line 863: not sure what the compiler did with the original version here: first `worker_id + (i - 1)*group_list_size` and then converting to larger size, or converting everything to larger size.

dev-zero:
ok, in Debug with gcc 7.5 tas/tensors is leaking memory

dev-zero:
@pseewald if you have time, https://github.com/cp2k/dbcsr/pull/399/commits/f137a4c271127c90f18159f8c7b9fa05b4d528d1 would be for your ;-)

alazzaro:
> ok, in Debug with gcc 7.5 tas/tensors is leaking memory

Is it something we are already masking via:

https://github.com/cp2k/dbcsr/blob/develop/tools/docker/lsan.supp

?

dev-zero:
> > ok, in Debug with gcc 7.5 tas/tensors is leaking memory
> 
> Is it something we are already masking via:
> 
> https://github.com/cp2k/dbcsr/blob/develop/tools/docker/lsan.supp
> 
> ?

right, sorry. Why haven't we integrated that in CMake yet? And where is the upstream bug tracked again?
It's not a false positive, but a real memleak nonetheless, right?

alazzaro:
> > > ok, in Debug with gcc 7.5 tas/tensors is leaking memory
> > 
> > 
> > Is it something we are already masking via:
> > https://github.com/cp2k/dbcsr/blob/develop/tools/docker/lsan.supp
> > ?
> 
> right, sorry. Why haven't we integrated that in CMake yet? And where is the upstream bug tracked again?
> It's not a false positive, but a real memleak nonetheless, right?

The original bug was in CP2K (see https://github.com/cp2k/cp2k/issues/1023 ).

src/mm/dbcsr_mm.F
dev-zero: ok, but then I have to do: `2.0*REAL(l_row - f_row + 1)*REAL(l_col - f_col + 1))/REAL(numnodes)*REAL(l_k - f_k + 1)`

hfp:
There are conflicts now and it appears to be stale work. If still desired, one can start over from the closed PR.

Issue #389, opened 2020-11-22
Bug when cublas and cusmm are used at the same time


alazzaro:
See https://github.com/cp2k/cp2k/issues/911

alazzaro:
@oschuett 
For the discussion we had on this bug, I'm not completely sure that your change will make the difference here... Well, probably I'm fully clear how the stream GPU mechanism works in DBCSR...

In the current code used by DBCSR in CP2K, each CPU thread submits to the GPU on a different portion of the C matrix. Then, each thread is attached to a stream (via a round-robin). Now, I understand the cusmm uses atomics, indeed we run simultaneous GPU threads that can operate within a CPU thread, so we need to be thread-safe. That's OK.
Assuming that at some point we get a large kernel, this is submitted by each CPU thread (so independent C-portions) via a cublas call in a loop of calls within the same stream. The code is what you linked:

https://github.com/cp2k/dbcsr/blob/d2dfe6e25efde255f72ac27f542124b69c7af1c2/src/mm/dbcsr_acc_operations.F#L150

Yes, the cublas DGEMM is not thread-safe, but we are running only a cublas dgemm at the time (they serialize within the same stream)... This is not a batched call, where for sure we will have problems.

Even assuming previous cusmm calls competing with the cublas calls, they still run on the same stream and serialize...

Should I assume that the consecutive calls on the same stream somehow are not serialized?
Could you elaborate where we have symultanous cublas dgemm calls accessing to the same C data?

oschuett:
Each OpenMP threads has multiple stack buffers (for different combinations of mnk). Each stack buffer has it's own stream (more or less). Hence, there are multiple streams per OpenMP thread.

Unfortunately, this stream / even part is quite complex. Have a look at Figure 4 in the [book chapter](https://www.cp2k.org/_media/gpu_book_chapter_submitted.pdf).

alazzaro:
Thanks @oschuett , now I understand the stacks and streams relation...
Now, your idea of using the global stream per each thread, i.e. `c_area%memory_type%acc_stream`, will make cublas calls OK. So we have
* cusmm calls OK
* cublas calls OK
What about the mix case, i.e. cublas and cusmm operating on the same C-block?

oschuett:
[...]

PR #400, opened 2020-12-01
Cublas stream bug #389

alazzaro:
Introduce the c_stream for cuBLAS to avoid race-condition due to no thread-safe cublas DGEMM. See #389

### DIFF
--- a/src/mm/dbcsr_mm_accdrv.F
+++ b/src/mm/dbcsr_mm_accdrv.F
@@ -566,6 +566,7 @@ CONTAINS
       CALL acc_event_record(stackbuf%ready, stream=upload_stream)
 
       CALL acc_stream_wait_event(stackbuf%stream, stackbuf%ready)
+      CALL acc_stream_wait_event(c_area%memory_type%acc_stream, stackbuf%ready)
 
       CALL dbcsr_acc_do_mm_stack(params, stackbuf%devmem, stack_size, c_area%data_type, &
                                  a_data=a_area%acc_devmem, &
--- a/src/mm/dbcsr_mm_accdrv.F
+++ b/src/mm/dbcsr_mm_accdrv.F
@@ -575,7 +575,8 @@ CONTAINS
                                  n_max=stack_descr%max_n, &
                                  k_max=stack_descr%max_k, &
                                  def_mnk=stack_descr%defined_mnk, &
-                                 stream=stackbuf%stream, &
+                                 stack_stream=stackbuf%stream, &
+                                 c_stream=c_area%memory_type%acc_stream, &
                                  success=success)
 
       IF (success) THEN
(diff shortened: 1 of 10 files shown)


oschuett (approved):
LGTM

src/mm/dbcsr_mm_accdrv.F
```diff
@@ -566,6 +566,7 @@ SUBROUTINE dbcsr_mm_accdrv_process(this, left, right, params, stack_size, &
       CALL acc_event_record(stackbuf%ready, stream=upload_stream)
 
       CALL acc_stream_wait_event(stackbuf%stream, stackbuf%ready)
+      CALL acc_stream_wait_event(c_area%memory_type%acc_stream, stackbuf%ready)
```
oschuett, line 569: I don't think this is needed because `libsmm_acc_process_blas` uses `param_stack_host`. Hence, it does not have to wait for them to be uploaded to the device.

alazzaro: Let's keep the discussion here...
Your point is correct, there is not need to wait for the upload of the stacks.
Can it be that cublas has to start after the transpose kernel? How to sync that?

alazzaro:
Oh, I think the problem is *after* the cublas call, in these lines:

```
IF (success) THEN
         CALL acc_event_record(stackbuf%calculated, stream=stackbuf%stream)
         CALL acc_stream_wait_event(a_area%memory_type%acc_stream, stackbuf%calculated)
         CALL acc_stream_wait_event(b_area%memory_type%acc_stream, stackbuf%calculated)
         CALL acc_stream_wait_event(c_area%memory_type%acc_stream, stackbuf%calculated)
      ENDIF
```

Basically, the calculated in on the stackbuf%stream, so I think we should sync somehow the c_area%memory_type%acc_stream for cublas... @oschuett any advice is appreciated...

oschuett:
I don't think we have to worry about what happens after cublas, because it calls [StreamSynchronize](https://github.com/cp2k/dbcsr/blob/develop/src/acc/libsmm_acc/libsmm_acc.cpp#L249). Hence, the OpenMP thread will wait for cublas to finish before submitting any new kernels. Btw, there is even [a second](https://github.com/cp2k/dbcsr/blob/develop/src/acc/libsmm_acc/libsmm_acc.cpp#L251) call to StreamSynchronize, which can probably be removed.

However, your point about the transpose kernels is absolutely right!

We have to wait for `b_area%acc_ready` to make sure it's uploaded and transposed. Furthermore, we also have to wait for `a_area%acc_ready` to make sure it's uploaded as well.

Waiting for `stackbuf%ready` has the same effect, it just takes a little longer because it also waits for the stack buffer to be uploaded.

alazzaro:
Then, I would suggest keeping the current code. I will remove the other sync (good catch).

oschuett:
> Then, I would suggest keeping the current code.

Yes, but maybe add some explanatory comments. This event/stream stuff is hairy.

PR #404, opened 2020-12-03
Force thread_dist for C and left matrix to be the same

### LINKED ISSUE #379: CP2K error, seems due to DBCSR-GPU
See https://github.com/cp2k/cp2k/issues/893

alazzaro:
Try to fix #379

### DIFF
--- a/src/mm/dbcsr_mm.F
+++ b/src/mm/dbcsr_mm.F
@@ -394,11 +394,12 @@ CONTAINS
                  output_unit
       INTEGER(KIND=int_8)                                :: my_flop
       LOGICAL :: ab_dense, keep_product_data, keep_sparsity, product_reindex, release_tdist, &
-                 transpose_left, transpose_right, use_dense_mult, use_mempools
+                 transpose_left, transpose_right, use_dense_mult, use_mempools, thread_dist_force
       REAL(KIND=dp)                                      :: cs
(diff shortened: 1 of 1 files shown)


oschuett:
I have no idea how thread_dist works, but just swaping them seems a bit risky.

Maybe we should just always enable densification? Now with #389 fixed there even ought to be a performance gain.

alazzaro:
This is not a problem that densification can solve all the way, it will happen in the sparse case too. Densification mitigates the problem because it forces a re-evaluation of the thread-dist of the densified C matrix, based on the A matrix, which is exactly what I'm enforcing now even without densification... I think we are lucky that the problem didn't appear before (or maybe it is there, let's see what the CP2K CI will tell us) because most of our tests do not have large variability in block dimensions.

To explain better the problem, thread-dist assigns row-blocks to threads, so we avoid race-conditions on C.
So we decompose C and the A in threads. It is clear that both thread distributions for C and A must be the same, otherwise you are gone to miss blocks to store in C, which is exactly the reason for the bug.

alazzaro:
> Maybe we should just always enable densification? Now with #389 fixed there even ought to be a performance gain.

Unfortunately, I did several tests with densification enabled on the GPU. It turns out that only in some cases we gain in performance (but we use more memory). The reason is that 

a) densification takes time (copy of data)
b) we need to have a batched dgemm. 

I will keep the performance discussion for v2.2.

For this reason, densification is still off (we can make a flag to enabled it)

PR #406, opened 2020-12-03
OpenCL based ACC-backend and SMM library

hfp:
This PR just misses the build integration into DBCSR (CMake, etc.), which can be followup PR (*any help appreciated*). Transpose and SMM both work for general matrices, and the code is tested based on stand-alone reproducers.


hfp:
The build system may pull-in kernel source into the [transpose](https://github.com/cp2k/dbcsr/pull/406/files#diff-ca7c198ffac062cc7488f2ec226da484fc2907aebb28c42effbc63db97878a3cR19) and [smm](https://github.com/cp2k/dbcsr/pull/406/files#diff-ca7c198ffac062cc7488f2ec226da484fc2907aebb28c42effbc63db97878a3cR24) locations, which allows to stay independent of search paths otherwise needed for the separate kernel sources (which keeps the executable literally portable). The process of embedding the sources requires some minor processing to bring text files into format of string literals.

Kernel sources are templated such that all cases are handled. There is no limit currently implemented, but such arguments are part of the ACC/LIBSMM interface already (and marked unused in the current implementation). Similarly, processing heterogeneous stacks requires a certain result code such that DBCSR picks up the stack with more general code. This is similar to the CUDA/HIP backend. Btw, certain values are currently hard-coded across backends and might be covered by [ACC/LIBSMM interface](https://github.com/cp2k/dbcsr/blob/develop/src/acc/acc_libsmm.h) in the future (number of parameters per stack aka 3 on device side and 7 on host-side as well the return code to reject heterogeneous stacks, etc.).

hfp:
Now this PR builds with CMake. This work is in the middle of passing DBCSR tests (a great bunch does not pass at the moment).

hfp:
Specifically, `tests/dbcsr_unittest2` passes single-threaded, `tests/dbcsr_unittest4` passes in general, but `tests/dbcsr_unittest1` and `tests/dbcsr_unittest3` are failing.

CMakeLists.txt
```diff
@@ -92,12 +92,16 @@ set_property(CACHE USE_SMM PROPERTY STRINGS blas libxsmm)
...
 option(USE_HIP "Build with HIP support" OFF)
-# USE_CUDA and USE_HIP are mutually exclusive options: we either compile with
-# nvcc OR with hipcc
-if (USE_CUDA AND USE_HIP)
+option(USE_OPENCL "Build with OpenCL support" OFF)
+
+# USE_CUDA, USE_HIP, and USE_OPENCL shall be mutually exclusive options
+if ((USE_CUDA AND USE_HIP)
+    OR (USE_CUDA AND USE_OPENCL)
+    OR (USE_OPENCL AND USE_HIP))
```
dev-zero, line 100: should we maybe change to a `USE_ACCEL=(CUDA|HIP|OPENCL)` or can we anticipate that we are capable of including support for different accelerators in the same binary

hfp: Good suggestion! I will try getting this into this PR or log an issue if not.

CMakeLists.txt
```diff
@@ -186,13 +191,22 @@ if (USE_SMM MATCHES "blas")
...
   message("-- Using libxsmm for Small Matrix Multiplication")
 else ()
   message(FATAL_ERROR "Unknown SMM library specified")
 endif ()
 
-# =================================== GPU backend
+# =================================== GPU backends
+if (USE_OPENCL)
+  if (NOT LIBXSMM_FOUND)
+    message(FATAL_ERROR "LIBXSMM is not found but required for ACC/OpenCL.")
```
dev-zero, line 203: If the requirement for using OpenCL within DBCSR is to use `libxsmm` for the SMM operations, then the message should be: `"The OpenCL acceleration supports requires DBCSR to use libxsmm as the SMM backend, please install libxsmm and specify USE_SMM=xsmm`, if the user can use the OpenCL accel but still use BLAS as the SMM backend then simply repeat the `pkg_check_modules(LIBXSMM ...)` again (CMake should cache the result from earlier and not repeat the call). The other option is to move the `pkg_check_modules(LIBXSMM...)` outside without the `REQUIRED` and then use `if (NOT LIBXSMM_FOUND)` in both cases.

CMakeLists.txt
```diff
@@ -258,7 +270,6 @@ endif ()
 # inspired from
 # https://github.com/ROCm-Developer-Tools/HIP/tree/master/samples/2_Cookbook/12_cmake_hip_add_executable
 if (USE_HIP)
-
```
dev-zero, line 261: I always wonder why people dislike whitespace 😜

src/CMakeLists.txt
```diff
@@ -203,6 +203,25 @@ if (OpenMP_FOUND)
   target_link_libraries(dbcsr PRIVATE OpenMP::OpenMP_Fortran)
 endif ()
 
+# =================================================================================================
+# DBCSR LIBRARY's OPENCL BACKEND
+
+if (USE_OPENCL)
+  target_compile_definitions(dbcsr PRIVATE __DBCSR_ACC)
+  target_link_libraries(dbcsr PUBLIC ${OpenCL_LIBRARY})
```
dev-zero, line 211: you probably want `PRIVATE`

dev-zero:
> This PR just misses the build integration into DBCSR (CMake, etc.), which can be followup PR (_any help appreciated_). 

sorry about dropping the ball here, will take a look at it again next week (please ping me should I forget)

CMakeLists.txt
hfp: Hmm, indeed this is a more helpful error message. In fact, BLAS can be used for host/cpu side SMMs and I was planning to just keep it simply and require LIBXSMM. However, I will try to decouple things.

src/CMakeLists.txt
hfp: I just copied it from the CUDA part and thought it solved some problem for them. I will try PRIVATE.

hfp:
> sorry about dropping the ball here, will take a look at it again next week (please ping me should I forget)

No worries, you already helped with your review. I can take some more time to get all tests running. Though, this is the boring/time-consuming part trying to find where things break. However, it looks good already; I can see it running on my integrated GPU.

src/CMakeLists.txt
dev-zero: hmm, I remember there was a reason it was `PUBLIC` on CUDA, but not why

CMakeLists.txt
dev-zero: Well, it is not necessary to untangle it, there may be good reasons to use LIBXSMM as SMM backend when you already have it (I doubt that an AMD EPYC together with an Intel Xe becomes a common setup such that you would want to use BLIS+OpenCL).

hfp: LIBXSMM works well on AMD. We do CPUID-feature based code-gen/dispatch which yields the best-to-our-knowledge code-path. I doubt that BLIS can compete here as it does not focus on small matrix multiplications (SMMs). CPUID is a necessity for JIT'ted code as we need to figure out what code we want to generate. Further, one can force any code path at runtime even code that fails to run on a given arch (`LIBXSMM_TARGET` environment variable). Btw, we included ARM64 support just now (master) and will upstream this with Linux distributions with our next release.

alazzaro: For the record, I've already tried libxsmm on AMD and it is better than BLIS...

alazzaro:
@hfp Thanks a lot for this PR! This work will definitely go for v2.2. I will give a look and comment next week.

I have a general comment on the reason for having the OpenCL backend.
Andreas did a preliminary OpenCL implementation years ago. At that time, the idea was to have OpenCL for the GPUs, but then we realized that CUDA was the best for Nvidia (no surprise) and the OpenCL implementation became useless (we remove it during the transition from SVN to GIT). For the general case, now we support HIP. Shoshana did nice work so that CUDA and HIP can share almost the entire code. In this way, we can cover the NVIDIA and AMD GPUs (and whatever HIP can support). 

Now, I wonder which target do you have in mind for the OpenCL backend. Is this for the Intel GPU or FPGA?
And what about your OpenMP backend (https://github.com/cp2k/dbcsr/pull/260)?

Sorry for these questions... OpenCL is good to have it for sure, but we should avoid having too many backends to support...

hfp:
The OpenCL backend is meant for upcoming Intel GPUs. I could have done this with DPC++ and I may try this as we go forward. Meanwhile, I believe the OpenCL backend can be useful for FPGA/GPU targets in general (though, Intel FPGAs also allow for DPC++). Anyway, OpenCL is an industry standard and in case of Intel the runtime shares the bits with our Level-0 runtime (https://github.com/intel/compute-runtime). I consider this hence slightly more low-level than DPC++ and wanted to try this first. Regarding non-Intel, this backend should work fine on AMD and we can try/compare when ready.

alazzaro:
OK, thanks for the explanation and I agree with should have it for v2.2.
Next question then: how we do optimize the kernels, autotuning in the same way we do for CUDA/HIP?

hfp:
> And what about your OpenMP backend

The OpenMP backed was meant to be a proof of concept and it was found not suitable for implementing the ACC interface (at least not until a "stream"-clause is in reach). Generally, an API driven approach (as implemented with the ACC interface) is somewhat against a directive based approach (pragmas are naturally scoped with the code side). I think additional/future acceleration in DBCSR can chose the easiest approach whether it is API- or directive based especially if the ACC interface would require to grow. Though, there was no point touching the highly sophisticated Cannon implementation and it is way easier to write a backend.

alazzaro:
> > And what about your OpenMP backend
> 
> The OpenMP backed was meant to be a proof of concept and it was found not suitable for implementing the ACC interface (at least not until a "stream"-clause is in reach). Generally, an API driven approach (as implemented with the ACC interface) is somewhat against a directive based approach (pragmas are naturally scoped with the code side). I think additional/future acceleration in DBCSR can chose the easiest approach whether it is API- or directive based especially if the ACC interface would require to grow. Though, there was no point touching the highly sophisticated Cannon implementation and it is way easier to write a backend.

I definitely agree... the two approaches are somehow orthogonal and I liked a lot your attempt to add ACC-OpenMP... The situation may change in the future, but I agree with your comment.

hfp:
> how we do optimize the kernels, autotuning in the same way we do for CUDA/HIP?

I am deferring the auto-tuning for the time being (this PR). I was looking at the CUDA/HIP approach but found I had to rewrite the benchmark code etc. (that code is unfortunately written using CUDA/HIP for no real reason; it barely uses anything beyond the ACC interface). For auto-tuning OpenCL kernels, I would step back to a fixed set of triplets. My plan is to use the `acc_bench_smm` and `acc_bench_trans` drivers for tuning with perhaps OpenTuner. I have good experience with the latter and it is typically a few lines of code (see [here](https://github.com/hfp/libxsmm/blob/master/samples/transpose/transpose_opentuner.py) or [here](https://github.com/hfp/libxsmm/blob/master/samples/xgemm/xgemm_opentuner.py)). OpenTuner does a decent job wrt Hyperparameter tuning and it can avoid the old way of exhaustive exploration. Btw, especially predictive modeling with the CUDA/HIP auto-tuning approach is rather expensive/complex given the advantage over a fixed set of triplets. For this backend, the OpenTuner approach would simply run the afore mentioned drivers and set some environment variables ("tunables") embedded into the backend (e.g., `ACC_OPENCL_TRANS_WGSIZE`, `ACC_OPENCL_TRANS_INPLACE` just to speak about transpose; I am aware transpose is just B/W bound and there is not much to tune). The SMM code can easily expose "tunables" going forward with the kernel implementation.

alazzaro:
OK Hans, thanks for the other clarification. I agree with the idea of having a static optimizer, i.e. no JIT kernel. We can always think to have a "default" kernel for other kernels or even a fall-back to OpenCL BLAS. Great!

alazzaro:
One more question, this is for @dev-zero too: How do we test the OpenCL backend on the way forward?

hfp:
Well, this backend uses JIT code generation already and the tuning approach would also load a set of optimized parameters (perhaps even GPU specific). The JIT code gen. uses poor men templates by passing some '-DSomething' with OpenCL's build-line (and the kernel is potentially written based on "Something"). Effectively, there is no difference when compared to C++ templates and loading high-level C++ code into CUDA-RT.

dev-zero:
> One more question, this is for @dev-zero too: How do we test the OpenCL backend on the way forward?

Well, an OCI-compatible image with all the required runtime would be nice and if we can run on a CPU as a fallback for the OpenCL-part it becomes a lot easier. If that is indeed possible but only with an Intel CPU we can use our tcopt4-tcopt8 machines as a test runner.

hfp:
Wrt tests, perhaps the HIP image is a good choice? It should carry the OpenCL runtime as well/already.

dev-zero:
sure, but we should find a way to get a test which is also executed to be able to establish a baseline

CMakeLists.txt
hfp: I have implemented `USE_ACCEL` as part of this PR.

CMakeLists.txt
hfp: I have implemented `USE_ACCEL=opencl` and `USE_SMM=libxsmm` to be independent.

src/CMakeLists.txt
hfp: I am using `PRIVATE` now; seems to be OK.

hfp:
DBCSR's unit tests are passing. I have enabled DP and SP for the OpenCL backend (CUDA/HIP only implements DP).

alazzaro:
> DBCSR's unit tests are passing. I have enabled DP and SP for the OpenCL backend (CUDA/HIP only implements DP).

This is very, very, relevant... There was a request on SP, and the reply was "we don't support"
Let me review and we can merge whenever you are ready...

hfp:
I think it is ready except that I broke the ROCm tests (unfortunately, I only have a Vega56 at hand but it's in a macPro and the mildly spoken the OpenCL stack can be better likely for the same card in a PC). I probably need help for fixing this.

The backend should work on AMD, NVidia, and Intel OpenCL implementations. I ran tests on P100, V100, GEN9, and DG1. Additional fixup may be needed for one or the other OpenCL implementation, but this can be PR's as we go. I believe there are enough utilities in the backend to even do some light runtime discovery and fixup like [this](https://github.com/hfp/dbcsr/blob/openclacc/src/acc/opencl/acc_opencl.c#L237) one.

hfp:
One more caveat, I have not revised the hidden `.cp2k/Makefile` and did not yet integrate with CP2K.

alazzaro:
> One more caveat, I have not revised the hidden `.cp2k/Makefile` and did not yet integrate with CP2K.

Ignore the Makefile, v2.2 will be standalone in CP2K... Ideally, I want to still keep the makefile in CP2K to call cmake in DBCSR... As a next stage, DBCSR will be installed by the toolchain...

hfp:
For the record, multi-GPU must be still checked and tested on my side. Currently, `acc_set_active_device` causes to drop the context (if the device actually changes), which is sort of wrong. This should not harm for the moment and will be fixed asap. ATM, I working on SVM support trying to sketch how MKL's GPU implementation can be called. The latter is important as it can effectively save me from tuning the current kernels (they should not be too bad even in the current shape; though they are not outgrown or highly sophisticated).

hfp:
The last update dropped ~300 LOC, which make less to review.

hfp:
I looked into what I did for `acc_set_active_device`, and everything should be fine ("dropping device context") as long as [ACC_OPENCL_THREADLOCAL_CONTEXT](https://github.com/hfp/dbcsr/blob/openclacc/src/acc/opencl/acc_opencl.h#L74) is enabled (which is the case by default).

hfp:
ROCm tests are now passing since `dbcsr_acc_test` is disabled in case of HIP (TODO). This PR can be merged now without regressing existing tests/code.

alazzaro:
LGTM. I leave the final word to @dev-zero

hfp:
@alazzaro Thank you, Alfio!
@dev-zero If you want to take another look, it would be appreciated!

This PR is now ready, and I would like to merge before others touch the same bits like CMake/ROCm (see latest PR). I have introduced `OPENCL_LIBSMM_SMM_BATCHSIZE`, `OPENCL_LIBSMM_SMM_BLOCK_M`, and `OPENCL_LIBSMM_SMM_BLOCK_N` parameters and got acceptable defaults. I validated and exercised the code paths triggered by above three parameters, and DBCSR unit tests are passing on different devices/vendors for both DP and SP (P100, V100, GEN9, DG1).

Next step (follow-on PR) is to add an auto-tuning infrastructure and loading parameter files. This will be rather little amount of add-on code when compared to CUDA/HIP backend. As indicated above, the follow-on PR will lack predicting parameters, and supports (at least first step) only a fixed set of triplets.

src/acc/acc_bench_smm.c
```diff
@@ -20,9 +20,12 @@
 #if !defined(ELEM_TYPE)
 # define ELEM_TYPE double
 #endif
-#if !defined(EPSILON)
+#if !defined(EPSILON) && 0
```
dev-zero, line 23: This looks a bit like a leftover from debugging?
Maybe remove completely?

hfp: Indeed, it it makes little sense to have a fixed tolerance given all parameters/problem sizes; let alone DP vs. SP. I will remove this.

dev-zero:
awesome!

dev-zero:
I'll have to test the tri-state for `USE_ACCEL` with `cmake-gui`: not sure whether the empty string has to be explicitly mentioned or not. If it has to be explicitly mentioned: what would you prefer: the empty string or an explicit `NO` (or `OFF` or `FALSE`)?

hfp:
> 
> 
> I'll have to test the tri-state for `USE_ACCEL` with `cmake-gui`: not sure whether the empty string has to be explicitly mentioned or not. If it has to be explicitly mentioned: what would you prefer: the empty string or an explicit `NO` (or `OFF` or `FALSE`)?

As you like, you can suggest what's desired/aligned with DBCSR's standard (user expectation).

dev-zero:
@hfp the shell script is missing quotes around a lot of vars (to account for spaces, special chars, etc.). Maybe it would be a good idea to add `shellcheck` to the list of linters to run?

dev-zero:
Also, since we keep on adding file formats (I haven't checked whether it is already the case): we should version the file formats (as in embedding the a version number in the data) and verify the version compatibility right from the start.

hfp:
@dev-zero I tried Shellcheck myself and found it was not able to be adopted quickly like fixing Shell scripts in a straightforward fashion. In fact, some fixes I tried caused not only work but broke on BSD, etc. I tried Shellcheck with LIBXSMM's scripts, but feel free to add it to the list of linters.

hfp:
Talking about linters, the CMake formatter reordered the arguments of my script: https://github.com/hfp/dbcsr/blob/openclacc/src/acc/opencl/smm/CMakeLists.txt#L17 (the CSV-file argument was last originally). Since the script was already agnostic with respect to the order of arguments, it was acceptable (but it may be not acceptable for other cases).

hfp:
> 
> 
> Also, since we keep on adding file formats (I haven't checked whether it is already the case): we should version the file formats (as in embedding the a version number in the data) and verify the version compatibility right from the start.

Ahh, you mean the CSV-file and JSON files?  I consider these files internal to the backend. These files are generated as part of the auto-tuning process and if the format evolves (e.g., due to extension), I am sure code changes are local to the backend/libsmm.

hfp:
The last commit added support for auto-tuned parameters embedded into the binary by default. An environment variable can prevent considering these parameters as well as supply an external file with parameters.

I consider the following is still missing, but not necessary for this PR to be merged:

* Wrapper script to tune a series of kernels rather than a single kernel (multiple triplets).
* Documentation incl. HOWTO auto-tune kernels.

dev-zero:
> Ahh, you mean the CSV-file and JSON files? I consider these files internal to the backend. These files are generated as part of the auto-tuning process and if the format evolves (e.g., due to extension), I am sure code changes are local to the backend/libsmm.

As long as there is no chance at all that a user might start copying those around it's fine.

src/acc/opencl/acc_opencl.sh
```diff
@@ -0,0 +1,97 @@
...
+CPP=$(command -v cpp)
+RM=$(command -v rm)
+
+# GNU sed is desired (macOS)
+if [ "" = "${SED}" ]; then
+  SED=$(command -v sed)
+fi
+
+if [ "${BASENAME}" ] && [ "${SED}" ] && [ "${RM}" ]; then
+  for OFILE in $@; do :; done
```
dev-zero, line 22: quotes missing around `$@`

src/acc/opencl/acc_opencl.sh
```diff
@@ -0,0 +1,97 @@
...
+            BNAME=$(${BASENAME} "${IFILE}" .cl)
+            VNAME=opencl_libsmm_source_${BNAME}
+            MNAME=$(echo ${VNAME} | tr '[:lower:]' '[:upper:]')
+            echo "#define ${MNAME} ${VNAME}" >>${OFILE}
+            echo "const char ${VNAME}[] =" >>${OFILE}
+            echo "  \"#pragma OPENCL EXTENSION all: enable\\n\"" >>${OFILE}
+            if [ "${CPP}" ] && \
+               [ "$(${CPP} -dD -P -fpreprocessed ${IFILE} 2>/dev/null >/dev/null && echo OK)" ];
+            then
+              ${CPP} -dD -P -fpreprocessed ${IFILE}
```
dev-zero, line 46: quotes missing around `${IFILE}`

src/acc/opencl/acc_opencl.sh
hfp: Fixed.

src/acc/opencl/acc_opencl.sh
hfp: Fixed.

hfp:
I have not looked into Daint's modules for GPU, but since OpenCL is part of the CUDA installation CI-tests on Daint should be able test `USE_ACCEL=opencl`.

Some environment modules make CUDA available, but miss to extend the CPATH in order to point to an include directory with Nvidia's OpenCL headers. I guess standard headers would serve as well, but that needs something like "yum install opencl-headers".

To adjust `CPATH` in the Shell environment one may `export CPATH=${CUDA_ROOT}/targets/x86_64-linux/include:${CPATH}`, and CMake should figure out the OpenCL bits.

hfp:
The recent change unveiled another init/finalize issue, and #422 was submitted to investigate and confirm the expected behavior.

hfp:
> 
> 
> I'll have to test the tri-state for `USE_ACCEL` with `cmake-gui`: not sure whether the empty string has to be explicitly mentioned or not. If it has to be explicitly mentioned: what would you prefer: the empty string or an explicit `NO` (or `OFF` or `FALSE`)?

I am going to add `""` (two double quotes) to the list of strings, which apparently works in the GUI (offers four choices).

hfp:
I am going to merge this PR when tests are passing. I added Daint-CI tests for OpenCL, but disabled actual tests (just the "build" test). All runtime tests pass locally when working on Daint with similar settings like in CI-scripts. Also, Makefile based build (`acc_bench_trans` and `acc_bench_smm`) passes for OpenCL and CUDA backend (described [here](https://github.com/hfp/dbcsr/tree/develop/src/acc#benchmarks)). Missing bits are only incremental improvements (kernels) and documentation.

alazzaro:
Hans, this is great! Give me few days to review what you did before merging... I assume @dev-zero will take a look as well...

hfp:
ACK

hfp:
Prior to documentation update, here is a quick hint on how to test-drive auto-tuning:

```bash
cd dbcsr/src/acc/opencl
make [DBG=0]
cd smm
./tune_multiply.sh  100  1 1  23, 4 9
```

All arguments for tune_multiply.sh are optional, but the default set of triplet is already ~1400 kernels. The `100` given above is the number of seconds spent per kernel for auto-tuning; typically 300 should be aimed for (5 minutes per kernel). Later implementations may use a more informed value than just a fixed duration per kernel. The second group of arguments is about work-split (number of partitions/nodes, and second the actual part-number). The script displays an estimate for tuning prior to the whole process. The last group of command line arguments (`23, 4 9`) is an example of a triplet specification; each comma-separated group is expanded as series of triplets using the Cartesian product (`23, 4 9` generates 9 triplets).

An auto-tuning session (regardless of using the wrapper like `tune_multiply.sh` or the ordinary tuning script `tune_multiply.py`, produce a (series of) JSON file(s), which are summarized into a CSV file. After auto-tuning, simply rebuild:

```bash
cd dbcsr/src/acc/opencl
make [DBG=0]
```

Now, the backend, or to be more accurate - the LIBSMM library embedded the CSV file-content into the binary/executable. However, an environment variable `OPENCL_LIBSMM_SMM_PARAMS` allows to control adoption of tuned parameters (OPENCL_LIBSMM_SMM_PARAMS=0 disables using tuned parameters, and OPENCL_LIBSMM_SMM_PARAMS=/path/to/my.csv allows to supply a file and thereby overriding the parameters/default embedded into the binary). Both tuning parameters for SP and DP can be mixed and embedded or given per single file.

The auto-tuning simply (re-)uses the [benchmarks](https://github.com/hfp/dbcsr/tree/develop/src/acc#benchmarks) and parses console output like timing and data-type, etc. (benchmarks do not know about auto-tuning and tuning does not know about timing, etc.).

hfp:
Documentation for the OpenCL backend as well as for the OpenCL based LIBSMM was incorporated into Ford documentation. Existing documentation is adjusted to accommodate/distinct the OpenCL backend and OpenCL based LIBSMM (in addition to CUDA/HIP based backend/libsmm_acc). The (auto-)tuning guide for the OpenCL based LIBSMM is still missing but [drafted](https://github.com/cp2k/dbcsr/pull/406#issuecomment-766172349).

hfp:
( I am happy to merge #419 if merged earlier in develop. )

alazzaro:
Thanks Hans!

hfp:
Awesome, it made it -- Thank you!  I am looking forward to perhaps seeing others who can try this. I also look forward to improve the kernels. ATM kernels are relatively simple (although, they look cluttered like my code in general aka preprocessor defs). It would be in fact interesting to tune a limited variety of kernels on AMD GPUs and to see what the gap is compared to the CUDA/HIP backend. Unfortunately, I only have a Vega 56 inside of a macOS based platform (not exactly representative also because of missing DP).

alazzaro:
This is definitely on my plan... But first I have to make CP2K based on the DBCSR-cmake compilation and drop the makefiles inside DBCSR... I have a draft for that...

hfp:
Also note, there is a new test for Daint-CI. I disabled runtime tests, but believe there is only a minor issue. The build test however is fully enabled and I adjusted things for `USE_ACCEL=opencl` to build on Daint and to pick-up Nvidia's OpenCL implementation.

PR #407, opened 2020-12-04
Cleanup

hfp:


src/mpi/dbcsr_mpiwrap.F
```diff
@@ -2136,17 +2135,19 @@ SUBROUTINE mp_get_processor_name(procname, resultlen)
...
 #if defined(__parallel)
-      INTEGER                                            :: ierr
+      INTEGER                                            :: namelen, ierr
 
-      CALL mpi_get_processor_name(procname, resultlen, ierr)
-      IF (ierr /= 0) THEN
+      CALL mpi_get_processor_name(procname, namelen, ierr)
+      IF (ierr .EQ. 0) THEN
+         IF (PRESENT(resultlen)) resultlen = namelen
+      ELSE
```
dev-zero, line 2147: hmm, we could simply drop the whole `mp_get_processor_name` function, it seems unused?

We should definitely implement the linker script for the public API, build a shared lib using `-Wl,--as-needed` and then check which symbols get dropped in the process. Or maybe easier: collect all exported (`PUBLIC`) symbols and match them against `USE ..., ONLY: `?

dev-zero:
sorry, in the meantime d975abb7aa0785353770a7023af36ee84c6eb67d got merged, you may have to revert parts of that one
Othen than that, LGTM and may still make it for v2.1

src/mpi/dbcsr_mpiwrap.F
alazzaro: So, we actually need that function... Currently, we use the `m_hostnm` function directly, for example [here](https://github.com/cp2k/dbcsr/blob/42ad7bc98c88b118d1cbd1a12152b780f809d9b9/src/core/dbcsr_log_handling.F#L560). Ideally, we should replace that call with the more generic MPI call... 
Anyway, I like the idea of the linker to check the unused symbols...

alazzaro:
> sorry, in the meantime [d975abb](https://github.com/cp2k/dbcsr/commit/d975abb7aa0785353770a7023af36ee84c6eb67d) got merged, you may have to revert parts of that one
> Other than that, LGTM and may still make it for v2.1

I would like to close the 2.1 this week, so I tend to consider any new code change (including this PR) for the 2.2...

hfp:
I will try coming back to this PR over the course of this week, not sure tough...

src/mpi/dbcsr_mpiwrap.F
hfp: I have private/future changes that rely on `mp_get_processor_name`. Beside of Alfio's point (it's in-use), I believe this function can be also of future/further use.

hfp:
This PR only misses the Codecov watermark.

alazzaro:
Thanks Hans!

PR #408, opened 2020-12-09
Fix links in readmes/docs

shoshijak:


shoshijak:
going through all the links in the documentation to see if I missed anything before I merge ...

alazzaro:
I've killed Jenkins on Daint, since we don't need it for this PR...

PR #419, opened 2021-01-12
Simplify the CMake ROCm detection

haampie:
This is work in progress and untested atm, but opening a pr anyways to get early feedback.

In SIRIUS and SpFFT we had some more success with `find_packge(...)` to locate ROCm libraries, even when using spack to build the ROCm packages.

A spack install of ROCm is generally a useful way to check your cmake, since it does not have AMD's favorite directory `/opt/rocm`, nor does it have llvm installed in `$ROCM_PATH/llvm`, etc.

Edit: this is done


src/CMakeLists.txt
```diff
@@ -316,7 +311,7 @@ if (USE_HIP)
   hip_add_library(acc STATIC ${DBCSR_ACC_SRCS})
   target_compile_definitions(acc PRIVATE ARCH_NUMBER=${ACC_ARCH_NUMBER})
   target_compile_definitions(acc PRIVATE __HIP)
-  target_compile_options(acc PRIVATE "-fPIC")
+  set_property(TARGET acc PROPERTY POSITION_INDEPENDENT_CODE ON)
```
dev-zero, line 314: `set_target_properties` ?

src/acc/libsmm_acc/CMakeLists.txt
```diff
@@ -48,27 +48,27 @@ else () # i.e. USE_HIP
...
   if (USE_OPENMP)
-    # since HIP is based on clang, not GCC, doing:
-    # target_link_libraries(libsmm_acc PUBLIC OpenMP::OpenMP_CXX) does not work
-    # - it links to GNU's OpenMP (libgomp.so) and not clang's libomp.so
-    target_link_libraries(libsmm_acc PUBLIC HIP_OpenMP)
-  endif ()
+    target_link_libraries(libsmm_acc PUBLIC OpenMP::OpenMP_CXX)
+  endif()
+
+  set_property(TARGET libsmm_acc PROPERTY POSITION_INDEPENDENT_CODE ON)
```
dev-zero, line 61: does `hip_add_library` really not automatically set this for a `SHARED` lib?

dev-zero: at least for CMake it is the default for a shared lib: https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html

haampie: Let me try

src/CMakeLists.txt
haampie: Better

haampie:
@dev-zero thanks for the early suggestions. This PR solves the issue where the device compiler's OpenMP is used for linking, which doesn't really make sense. Right now it's always using the host compiler for OpenMP (and things simplify, cause we can just use `find_package(OpenMP REQUIRED)` only).

haampie:
Do you have opinions about including FindHIP module inside the sources, or should I bring back the HIP_PATH variable again for locating an external FindHIP module?

Currently it's required to find this HIP module through `find_package(HIP MODULE)` because it will otherwise detect `hip-config.cmake`, which is the hip config file (which we need too). From discussions on the internet, there's mixed opinions about having external module files.

dev-zero:
> Do you have opinions about including FindHIP module inside the sources, or should I bring back the HIP_PATH variable again for locating an external FindHIP module?
> 
> Currently it's required to find this HIP module through `find_package(HIP MODULE)` because it will otherwise detect `hip-config.cmake`, which is the hip config file (which we need too). The "problem' is that this module search mode only considers very restricted search paths (not CMAKE_PREFIX_PATH, only CMAKE_MODULE_PATH). From discussions on the internet, there's mixed opinions about having external module files.

ugh, is there any chance upstream is gonna fix this mess within reasonable time?
If yes I'd prefer if ROCm users would have to specify `CMAKE_MODULE_PATH`, that's easy to drop once fixed, doesn't hurt much if it proliferates (despite being not required anymore) and can be adapted to also build an older already released DBCSR against newer HIP.
Including `FindHIP` in our tree means we have to monitor upstream for changes and has the possibility that already released DBCSR versions break with newer ROCm framework versions. From that point of view I would probably prefer the `HIP_PATH` from before as it seemed reasonably easy to maintain.
But it's perfectly possible I am completely wrong here.
`pre-commit install --install-hooks` should save you some headaches ;-)

haampie:
I'll drop the FindHIP module from here.

haampie:
pre-commit install --install-hooks should save you some headaches ;-)

unfortunately this relies on python, which gives me headaches too :D it complains about python 3.6+ to be installed, but python 3.8 is the default on my desktop.

dev-zero:
interesting, what does `which pre-commit` say? My guess would be a `pip install --user pre-commit` on a system where `pip` is for `python2` and instead `pip3 install --user pre-commit` should've been used

haampie:
Oh, I installed it with `snap` since Ubuntu suggested that to me. Maybe it shipped its own python...

src/CMakeLists.txt
```diff
@@ -314,11 +309,10 @@ if (USE_HIP)
   set_source_files_properties(${DBCSR_ACC_SRCS}
                               PROPERTIES HIP_SOURCE_PROPERTY_FORMAT 1)
   hip_add_library(acc STATIC ${DBCSR_ACC_SRCS})
-  target_include_directories(acc PRIVATE ${HIP_PATH}/../include)
   target_compile_definitions(acc PRIVATE ARCH_NUMBER=${ACC_ARCH_NUMBER})
   target_compile_definitions(acc PRIVATE __HIP)
-  target_compile_options(acc PRIVATE "-fPIC")
-  target_link_libraries(acc PUBLIC ${HIPBLAS})
+  set_property(TARGET acc PROPERTY POSITION_INDEPENDENT_CODE ON)
```
dev-zero, line 314:
```suggestion
  set_target_properties(acc PROPERTIES POSITION_INDEPENDENT_CODE ON)
```

haampie: It was only necessary for this object library with timing functions. I've never seen an object lib before in cmake, but I've just left it there and added fPIC.

haampie:
Ok, I just tried compiling everything (rocm ecosystem & dbcsr) from source through spack, and I'm hitting

```
ABORT in dbcsr_lib.F:217 DBCSR compiled w/ threading support while libsmm_acc compiled w/o threading support.
```

Apparently the openmp issue is not entirely solved.

dev-zero:
> Ok, I just tried compiling everything (rocm ecosystem & dbcsr) from source through spack, and I'm hitting
> 
> ```
> ABORT in dbcsr_lib.F:217 DBCSR compiled w/ threading support while libsmm_acc compiled w/o threading support.
> ```
> 
> Apparently the openmp issue is not entirely solved.

Since the verbose makefile flag is set you should be able to check the compiler invocations in the log...

haampie:
Yeah, that's what you'd think. Turns out the HIP module generates its own cmake file which is kinda hidden from the rest of the build log. The TL;DR is: the device compiler was used for certain host source files, so the device compiler gets to link with their openmp lib. I have a nasty workaround, gonna test it now.

haampie:
Hm, I'm not really getting it to work. Also without OpenMP the unit tests fail:

```
HIPRTC ERROR: CompileProgram failed with error HIPRTC_ERROR_COMPILATION
```
Seems like it is thrown whenever a jit program is re-compiled: https://github.com/ROCm-Developer-Tools/HIP/blob/rocm-3.9.0/src/hiprtc.cpp#L501

haampie:
Boy, that was no fun. I should have looked in the issues too, cause #261 is related.

Previously when using just the host compiler (gcc) for openmp everywhere, I got this runtime error:

> Assertion `d != acc_device_none && d != acc_device_default && d != acc_device_not_host' failed

Turns GNU's `libgomp.so` defines OpenACC functions, including `acc_init`, and I believe when the dbcsr fortran function `acc_init` is called, which in turn calls the C-function `acc_init`, it ends up calling a function in libgomp instead of the C-interface of dbcsr's acc lib. Very confusing.

With the current patches I can finally build dbcsr for ROCm using only the host compiler's implementation of OpenMP, which makes sense to me. Previously I would end up with both libomp.so _and_ libgomp.so in the dbcsr lib.

haampie:
Remaining issues:

- [ ] `HIPRTC ERROR: CompileProgram failed with error HIPRTC_ERROR_COMPILATION`
- [ ] Some sources that get compiled wit the host compiler include hip header files that want `__HIP_PLATFORM_HCC__` or `__HIP_PLATFORM_NVCC__` to be defined.

haampie:
@dev-zero, I realized just now that all device code is 100% jitted, is that correct? In that case we can drop hipcc / the device compiler altogether from cmake? Or are there cases in which kernels are compiled ahead of time too?

haampie:
ping @mtaillefumier

haampie:
So, the TL;DR if this PR is:

- Drop searching for the device compiler during cmake configuration, since we only have to link to ROCm libraries
- Seems like rocm 3.5.0 and above has moved to using ROCclr which changed improved their jit/hiprtc* code as it now does not call `hipcc` anymore but only clang directly, which is good. But they don't seem to forward this `-D__HIP` flag, which caused the jit to fail. Fixed by looking for __HIP_ROCclr__.
- The C api is now prefixed with c_dbsr_* because otherwise the Fortran code ended up calling `acc_init` in libgomp.so instead of the C-part of dbcsr :upside_down_face:.
- Added -Wno-error=...deprecation warnings, because the relevant code actually already works around that. (Or maybe we should accept that ROCm just has a different interface and a simple macro for NVCC & HIP doesn't work?)
- Updated the docs to reflect what's to be configured for HIP

haampie:
Tests are passing on Ault btw:

```
Test project dbcsr-project/dbcsr/spack-build-le4geew
      Start  1: dbcsr_perf:inputs/test_H2O.perf
 1/21 Test  #1: dbcsr_perf:inputs/test_H2O.perf .......................   Passed    6.48 sec
      Start  2: dbcsr_perf:inputs/test_rect1_dense.perf
 2/21 Test  #2: dbcsr_perf:inputs/test_rect1_dense.perf ...............   Passed    1.75 sec
      Start  3: dbcsr_perf:inputs/test_rect1_sparse.perf
 3/21 Test  #3: dbcsr_perf:inputs/test_rect1_sparse.perf ..............   Passed    2.64 sec
      Start  4: dbcsr_perf:inputs/test_rect2_dense.perf
 4/21 Test  #4: dbcsr_perf:inputs/test_rect2_dense.perf ...............   Passed    1.73 sec
      Start  5: dbcsr_perf:inputs/test_rect2_sparse.perf
 5/21 Test  #5: dbcsr_perf:inputs/test_rect2_sparse.perf ..............   Passed    2.28 sec
      Start  6: dbcsr_perf:inputs/test_singleblock.perf
 6/21 Test  #6: dbcsr_perf:inputs/test_singleblock.perf ...............   Passed    1.65 sec
      Start  7: dbcsr_perf:inputs/test_square_dense.perf
 7/21 Test  #7: dbcsr_perf:inputs/test_square_dense.perf ..............   Passed    1.64 sec
      Start  8: dbcsr_perf:inputs/test_square_sparse.perf
 8/21 Test  #8: dbcsr_perf:inputs/test_square_sparse.perf .............   Passed    1.85 sec
      Start  9: dbcsr_perf:inputs/test_square_sparse_bigblocks.perf
 9/21 Test  #9: dbcsr_perf:inputs/test_square_sparse_bigblocks.perf ...   Passed    5.31 sec
      Start 10: dbcsr_perf:inputs/test_square_sparse_rma.perf
10/21 Test #10: dbcsr_perf:inputs/test_square_sparse_rma.perf .........   Passed    1.87 sec
      Start 11: dbcsr_unittest1
11/21 Test #11: dbcsr_unittest1 .......................................   Passed   59.63 sec
      Start 12: dbcsr_unittest2
12/21 Test #12: dbcsr_unittest2 .......................................   Passed   22.83 sec
      Start 13: dbcsr_unittest3
13/21 Test #13: dbcsr_unittest3 .......................................   Passed   67.32 sec
      Start 14: dbcsr_unittest4
14/21 Test #14: dbcsr_unittest4 .......................................   Passed    0.56 sec
      Start 15: dbcsr_tensor_unittest
15/21 Test #15: dbcsr_tensor_unittest .................................   Passed   69.53 sec
      Start 16: dbcsr_tas_unittest
16/21 Test #16: dbcsr_tas_unittest ....................................   Passed  112.41 sec
      Start 17: dbcsr_test_csr_conversions
17/21 Test #17: dbcsr_test_csr_conversions ............................   Passed    1.19 sec
      Start 18: libsmm_acc_unittest_multiply
18/21 Test #18: libsmm_acc_unittest_multiply ..........................   Passed   13.97 sec
      Start 19: libsmm_acc_unittest_transpose
19/21 Test #19: libsmm_acc_unittest_transpose .........................   Passed   10.58 sec
      Start 20: libsmm_acc_timer_multiply-autotuned
20/21 Test #20: libsmm_acc_timer_multiply-autotuned ...................   Passed   15.37 sec
      Start 21: libsmm_acc_timer_multiply-predicted
21/21 Test #21: libsmm_acc_timer_multiply-predicted ...................   Passed    0.12 sec
```

haampie:
@dev-zero, this PR is done, can you review?

CMakeLists.txt
```diff
@@ -268,20 +265,7 @@ if (USE_HIP)
...
-          CACHE PATH "Path to HIP installation")
-    else ()
-      set(HIP_PATH
-          $ENV{HIP_PATH}
-          CACHE PATH "Path to HIP installation")
-    endif ()
-  endif ()
-  set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${HIP_PATH}/cmake")
-
+  # ROCm is typically installed in /opt/rocm, so provide that prefix as a
```
dev-zero, line 268: unfinished sentence?

cmake/CompilerConfiguration.cmake
```diff
@@ -41,8 +41,8 @@ Please open an issue at https://github.com/cp2k/dbcsr/issues with the reported c
 endif ()
 
 if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-  set(CMAKE_CXX_FLAGS_RELEASE      "-O3 -g -funroll-loops -Wall -Werror")
-  set(CMAKE_CXX_FLAGS_COVERAGE     "-O0 -g --coverage -Wall -Werror")
+  set(CMAKE_CXX_FLAGS_RELEASE      "-O3 -g -funroll-loops -Wall -Werror -Wno-error=deprecated-declarations")
+  set(CMAKE_CXX_FLAGS_COVERAGE     "-O0 -g --coverage -Wall -Werror -Wno-error=deprecated-declarations")
```
dev-zero, line 45: We should drop the `-Werror` for the release.
And I don't like having `-Wno-error` on the whole project for a subcomponent.

docs/guide/2-user-guide/1-installation/index.md
```diff
@@ -95,10 +95,13 @@ If MPI support is enabled (the default), the C API is automatically built.
...
 
-HIP is a relatively new language, and some issues still need to be ironed out. As a workaround to an [issue](https://github.com/ROCm-Developer-Tools/HIP/pull/1543) in HIP's JIT infrastructure, please set the following if you've built HIP from source:
+For custom installs of HIP 3.9.0 and above, some paths have to be configured to ensure the JIT compiler can locate the HIP runtime and compiler tools
 
 ```bash
-    export HIP_PATH=/opt/rocm/hip
+    export ROCM_PATH=/path/to/hip-3.9.0
+    export HIP_PATH=$ROCM_PATH
+    export LLVM_PATH=/path/to/llvm-amdgpu-3.9.0
+    export HIP_DEVICE_LIB_PATH=/path/to/rocm-device-libs-3.9.0/amdgcn/bitcode
```
dev-zero, line 104: Would it be possible and make sense to get defaults for those at build time (from the build environment) and set them internally, but keep them overrideable with env vars?

src/CMakeLists.txt
```diff
@@ -297,28 +297,19 @@ endif ()
...
 
   # Complete list of GPU-support sources
   set(DBCSR_ACC_SRCS ${DBCSR_HIP_SRCS})
 
   # Compile the rest of the HIP source files into a static library
-  set_source_files_properties(${DBCSR_ACC_SRCS}
-                              PROPERTIES HIP_SOURCE_PROPERTY_FORMAT 1)
-  hip_add_library(acc STATIC ${DBCSR_ACC_SRCS})
-  target_include_directories(acc PRIVATE ${HIP_PATH}/../include)
+  add_library(acc SHARED ${DBCSR_ACC_SRCS})
```
dev-zero, line 309: Do we have to build a full SHARED lib her now?
How does that work with the rest of DBCSR being built as a static library (by default)?
Could we use an `OBJECT` here now (getting only one lib in the end)?

src/acc/libsmm_acc/CMakeLists.txt
```diff
@@ -46,30 +46,22 @@ if (USE_CUDA)
...
-                              PROPERTIES HIP_SOURCE_PROPERTY_FORMAT 1)
-  # hip_add_library does not support OBJECT libraries. For this reason, we are
-  # forced to use a static library
-  hip_add_library(libsmm_acc STATIC ${LIBSMM_ACC_FILES})
-  target_include_directories(libsmm_acc PRIVATE ${HIP_PATH}/../include)
-  target_link_libraries(libsmm_acc INTERFACE "stdc++")
-  target_link_libraries(libsmm_acc PUBLIC ${HIPBLAS})
+  add_library(libsmm_acc SHARED ${LIBSMM_ACC_FILES})
+  target_link_libraries(libsmm_acc PUBLIC roc::hipblas hip::host)
+
```
dev-zero, line 51: similar here: `OBJECT` possible and would it make sense?

tests/CMakeLists.txt
```diff
@@ -235,16 +227,15 @@ if (USE_CUDA OR USE_HIP)
...
 
-      hip_add_executable(${libsmm_acc_test_name} ${libsmm_acc_test})
-      target_link_options(${libsmm_acc_test_name} PRIVATE ${HIP_ARCH_FLAGS})
-
-      target_link_libraries(${libsmm_acc_test_name} dbcsr)
+      add_executable(${libsmm_acc_test_name} ${libsmm_acc_test})
+      target_link_libraries(${libsmm_acc_test_name} dbcsr hip::host)
       target_compile_definitions(${libsmm_acc_test_name} PRIVATE __HIP)
+      if (OpenMP_FOUND)
+        target_link_libraries(${libsmm_acc_test_name} OpenMP::OpenMP_CXX)
```
dev-zero, line 237: I haven't checked, but if the `libsmm_acc_test_name` sources (but it is required by `libsmm_acc`) do not explicitly use OpenMP then `OpenMP::OpenMP_CXX` should be exported by the the library and passed down via `target_link_libraries(... dbcsr)` rather than explicitly.

dev-zero (changes requested):
looks good, OpenMP/CMake-remark might be solved by #406

cmake/CompilerConfiguration.cmake
haampie: >  And I don't like having -Wno-error on the whole project for a subcomponent.
`set_source_files_properties` for one specific source file then?

docs/guide/2-user-guide/1-installation/index.md
haampie: Hm yeah, not sure, cause you'd end up with absolute paths in binaries, which is maybe not what you want when you distribute binaries. If rocm is installed by apt/yum/.. it should not be necessary to set these variables, I can clarify that a bit more.

src/CMakeLists.txt
haampie: Let me try, I took the old comments above this as if it was intended to build a shared lib, but it wasn't supported by hip_add_library at that point in time.

tests/CMakeLists.txt
haampie: I'll check

cmake/CompilerConfiguration.cmake
haampie:
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
...
#pragma GCC diagnostic pop
```
annotations is another way to do it...

dev-zero: sounds good, with a generator expression `$<CXX_COMPILER_ID:GNU>`?

docs/guide/2-user-guide/1-installation/index.md
dev-zero: Might be worth to document it, yes please.
This is the kind of the stuff people randomly stick in their scripts and carry around forever until it starts to break something else...
What happens btw if those vars are not set? Do we have a FAQ somewhere such that people find it when googling the error messages?

cmake/CompilerConfiguration.cmake
dev-zero: ... also the `#pragma` works, might be even cleaner.

haampie:
hm, no clue what is different for the rocm tests this time :/

src/acc/libsmm_acc/CMakeLists.txt
haampie: I've removed the object files, these are source files of 1 lib anyways

docs/guide/2-user-guide/1-installation/index.md
haampie: I'll address this in a separate pr, don't have access to AMD GPUs right now.

cmake/CompilerConfiguration.cmake
haampie: pragma it is

alazzaro:
@haampie 
Thanks a lot for your contribution. I will review this PR this week. Could you rebase with the new develop branch?

haampie:
@hfp rebasing now, but I'm confused about the changes w.r.t. `acc_init` / `acc_finalize`. Does OpenCL actually use OpenACC functions? Did you somehow run into the same problem as I where the intent was to call acc_init as defined by dbcsr, but it inadvertently called the function of the same name in libgomp.so?

alazzaro:
> @hfp rebasing now, but I'm confused about the changes w.r.t. `acc_init` / `acc_finalize`. Does OpenCL actually use OpenACC functions? Did you somehow run into the same problem as I where the intent was to call acc_init as defined by dbcsr, but it inadvertently called the function of the same name in libgomp.so?

I leave to @hfp to reply to your message, I don't know the exact details. I think the conflict is due to the merge of the #406 (which I did today). Sorry for the problem...

hfp:
> 
> 
> > @hfp rebasing now, but I'm confused about the changes w.r.t. `acc_init` / `acc_finalize`. Does OpenCL actually use OpenACC functions? Did you somehow run into the same problem as I where the intent was to call acc_init as defined by dbcsr, but it inadvertently called the function of the same name in libgomp.so?
> 
> I leave to @hfp to reply to your message, I don't know the exact details. I think the conflict is due to the merge of the #406 (which I did today). Sorry for the problem...

I did not run into the issue with libgomp.so as far as I realized/know. What "changes w.r.t. `acc_init` / `acc_finalize`" do mean? Sorry, there are a lot of changes just to make sure I can best explain changes.

hfp:
It looks like `DBCSR_GENERIC_ACC_SRCS` is linked in general, i.e., even with `USE_ACCEL=opencl`:

```cmake
set(DBCSR_GENERIC_ACC_SRCS
    acc/libsmm_acc/libsmm_acc_benchmark.cpp
    acc/libsmm_acc/libsmm_acc_init.cpp
    acc/libsmm_acc/libsmm_acc.cpp
    acc/cuda/acc_cublas.cpp
    acc/cuda/acc_dev.cpp
    acc/cuda/acc_error.cpp
    acc/cuda/acc_event.cpp
    acc/cuda/acc_init.cpp
    acc/cuda/acc_mem.cpp
    acc/cuda/acc_stream.cpp)
```

I believe this is a continuation of refactoring code shared between CUDA and HIP, and promoting it to be "common" for all ACC backends. Since none of the above translation units (listed in `DBCSR_GENERIC_ACC_SRCS`) seem to be protected to only contribute code in case of `__CUDA` or `__HIP`, this should lead to duplicate symbols in case of `USE_ACCEL=opencl` (if I am not mistaken).

We once had separate backends and two interfaces implemented by every backend (one for "acc" and one for "libsmm"). This would have enabled writing the entire auto-tuning for all backends including the "acc/libsmm_acc/libsmm_acc_benchmark.cpp", etc. Instead, things like libsmm_acc_benchmark.cpp depends on CUDA/HIP despite of using not much more than ACC functionality.

Anyhow, I will give it a try and compile/test this PR on my end perhaps I overlooked something.

hfp:
Indeed, `USE_ACCEL=opencl` does not compile (which is not visible in above testset since Daint-CI was not triggered (which build-tests the OpenCL backend).

haampie:
Hi @hfp, I should've left a comment after the rebase. I wasn't yet finished fixing the OpenCL build, let me fix that right now. 

> and promoting it to be "common" for all ACC backends

Is that a good or a bad thing? I think the CUDA / HIP backend have a sufficiently similar API to make it possible, is OpenCL different? I'll just add those `DBCSR_GENERIC_ACC_SRCS` sources conditionally then when USE_ACCEL matches hip|cuda.

> Instead, things like libsmm_acc_benchmark.cpp depend on CUDA/HIP despite of using not much more than ACC functionality.
Do you mean those `ACC_API_CALL` / `ACC_DRV_CALL` / `ACC_DRV` macro's? Seems like they're not very useful anyways, because `acc_hip.h` has a set of small wrappers to call e.g. `hipHostMalloc` from `hipHostAlloc` because the API has changed.

hfp:
> Is that a good or a bad thing?

I believe it makes little sense since functions are implemented different for OpenCL vs. CUDA/HIP (even the latter require at least to toggle different include/header files). Of course, it is on purpose that CUDA and HIP can be toggled by just a prefix for function names since the runtime since intentionally mimics CUDA. Alternatively, one can share the code over all backends but need to introduce compile-time toggles like `__OPENCL` vs. CUDA/HIP equivalent (which does not leave much to share except of file/name).

> Do you mean those ACC_API_CALL / ACC_DRV_CALL / ACC_DRV macro's

Yes, they prevented successful compilation in the first place, but linker errors would have followed as noted earlier.

Your changes make sense now! Sorry for grabbing them too early. Let me know when to retry (if you like).

haampie:
Ok, I've made the OpenCL bits compile with CMake again, but I must say it's still quite a mess...

Now I'll try and run the tuning with AMD's libOpenCL.so

hfp:
> i've also prefixed all your acc_* functions with c_d

Thank you! Yes, there is an "internal" API used in OpenCL backend and OpenCL based LIBSMM.

@haampie I am trying the PR and looks good so. I was also able to merge with my changes since my PR was merged. I found minor thing: `src/CMakeLists.txt:223` has some disturbing curly-brace `}` causing warnings like "ISO C99 requires whitespace after the macro name".

haampie:
Thanks for pointing that out, fixed it.

I realize that I didn't get back to you about

> What "changes w.r.t. acc_init / acc_finalize" do mean?

and I might have dropped your workarounds where you added calls to `acc_init()` / `libsmm_acc_init` in ifdef blocks. I might have to bring that back.

src/acc/acc_bench_smm.c
```diff
@@ -118,20 +116,17 @@ int main(int argc, char* argv[])
 #if defined(_DEBUG)
     fprintf(stderr, "Error: no device found!\n");
 #endif
-#if !defined(__CUDA)
-    CHECK(libsmm_acc_finalize(), NULL);
-#endif
-    CHECK(acc_finalize(), NULL);
```
haampie, line 124: @hfp: this is where I got confused and where I might have undone your fixes for OpenCL?

`c_dbcsr_acc_finalize` is confusingly defined in `cuda/acc_init.cpp`, but the same source file is used for HIP, so I thought the ifdef didn't make sense and removed it. The OpenCL version of this function also call libsmm_acc_finalize, so is this really necessary?

haampie: We should probably move those shared HIP / CUDA sources out of the cuda/ folder.

hfp:
I compiled the PR successfully, but ran into runtime errors thrown by DBCSR code like "DBCSR compiled w/ threading support while libsmm_acc compiled w/o threading support.". I looked at CMake VERBOSE=1 output and compared between develop and rocm, I found two things:

1. Branch "develop" and OpenCL backend in particular (`acc/opencl/*`) is not compiled with `-fopenmp` flag, however the LIBSMM code is compiled with the OpenMP flag (`acc/opencl/smm/*`). This at least was sufficient to avoid the mentioned runtime error, but still wrong.
2. PR "rocm" does not apply the OpenMP flag to any of the backend code, which is wrong too (if not more ;-), and it yields the mentioned runtime error.

hfp:
In fact, if OpenCL BE and LIBSMM will both get the OpenMP flag, lines like [this](https://github.com/cp2k/dbcsr/blob/develop/src/acc/opencl/acc_opencl.c#L134) will cause internal checks to fail. In my branch, I replaced:

```
int result = (0 == omp_in_parallel() ? EXIT_SUCCESS : EXIT_FAILURE);
```

with the following:

```
  int result = ((0 == omp_in_parallel()
# if /*WORKAROUND*/defined(__DBCSR_ACC)
    || 0/*master*/ == omp_get_thread_num()
# endif
    ) ? EXIT_SUCCESS : EXIT_FAILURE);
```

In addition, `libsmm_acc_finalize` (in `opencl/smm/opencl_libsmm.c`) the line `#if defined(_OPENMP) && /*WORKAROUND*/!defined(__DBCSR_ACC)` was replaced with just `#if defined(_OPENMP)`.

This stuff is related to #422.

haampie:
> will cause internal checks to fail

What checks actually? Do the tests fail?

hfp:
My note about https://github.com/cp2k/dbcsr/pull/419#issuecomment-772821420 is only relevant if you run the OpenCL based DBCSR. I think we should get your PR merged asap and then apply any necessary fixes on top (it is getting messy otherwise).

hfp:
> 
> 
> > will cause internal checks to fail
> 
> What checks actually? Do the tests fail?

Everything (real workload or test) fails without applying https://github.com/cp2k/dbcsr/pull/419#issuecomment-772821420 if code under src/opencl gets the OpenMP flag (which it should). Though, this was my mistake in CMake stuff not applying OpenMP flag to src/opencl, I only applied the flag to src/opencl/smm (which was sufficient to not run into this).

haampie:
```
$ ./acc_bench_smm 
./acc_bench_smm 3 30000 23 23 23 1875 18750 18750
Error: no device found!
```
can reproduce when using `make ... DBG=2 OMP=1` in `src/acc/opencl`

hfp:
> 
> 
> ```
> $ ./acc_bench_smm 
> ./acc_bench_smm 3 30000 23 23 23 1875 18750 18750
> Error: no device found!
> ```
> 
> can reproduce when using `make ... DBG=2 OMP=1` in `src/acc/opencl`

Indeed, the `Makefile` is consistent if `OMP=1` is given. I should have noticed this earlier. Anyhow, this is sorted out one after another (your PR first).

haampie:
Let me just run `ctest` with hip support and without openmp to see if it's still fine

hfp:
> 
> 
> > ```
> > $ ./acc_bench_smm 
> > ./acc_bench_smm 3 30000 23 23 23 1875 18750 18750
> > Error: no device found!
> > ```
> > 
> > 
> > can reproduce when using `make ... DBG=2 OMP=1` in `src/acc/opencl`
> 
> Indeed, the `Makefile` is consistent if `OMP=1` is given. I should have noticed this earlier. Anyhow, this is sorted out one after another (your PR first).

Hmm, maybe something else since the stand-alone benchmarks do not call init/finalize in parallel regions. Let's not worry, there is no runtime test (yet) in CI to cover the OpenCL backend.

hfp:
I think this PR should be merged tomorrow as soon as you think it is ready.

haampie:
Oof... it might as well be that there is still maintenance going on with the node I'm on :sweat_smile: both `clinfo` and `rocm_agent_enumerator` stopped listing the GPU, even though there are still entries in `/dev/dri`. Maybe disregard my 'can reproduce' comment.

> I think this PR should be merged tomorrow as soon as you think it is ready.

Let's merge it then! There's definitely some improvements in this PR over `develop`, and all sources now use OpenMP when that's enabled. Just to be sure, did you see https://github.com/cp2k/dbcsr/pull/419/commits/70825c34e8d2713d666fffa25a923a88195c3741 too? It was necessary for me.

alazzaro:
retest this please

alazzaro:
@haampie Daint-CI found some compilation problems, can you take a look?

haampie:
Where should I look?

alazzaro:
retest this please

alazzaro:
OK, we are a step forward...
Now it cannot find the executable (I assume because of a wrong name):

```
      Start 20: /scratch/snx3000/jenkg90/jenkins-g90-DBCSR-681.intel/tests/libsmm_acc_unittest_multiply.cpp
Process not started
 /scratch/snx3000/jenkg90/jenkins-g90-DBCSR-681.intel/tests/libsmm_acc_unittest_multiply.cpp
[permission denied]
20/24 Test #20: /scratch/snx3000/jenkg90/jenkins-g90-DBCSR-681.intel/tests/libsmm_acc_unittest_multiply.cpp ...***Not Run   0.00 sec
      Start 21: /scratch/snx3000/jenkg90/jenkins-g90-DBCSR-681.intel/tests/libsmm_acc_timer_multiply.cpp
Process not started
 /scratch/snx3000/jenkg90/jenkins-g90-DBCSR-681.intel/tests/libsmm_acc_timer_multiply.cpp
[permission denied]
21/24 Test #21: /scratch/snx3000/jenkg90/jenkins-g90-DBCSR-681.intel/tests/libsmm_acc_timer_multiply.cpp ......***Not Run   0.00 sec
      Start 22: libsmm_acc_unittest_transpose.cpp
Could not find executable libsmm_acc_unittest_transpose.cpp
Looked in the following places:
libsmm_acc_unittest_transpose.cpp
libsmm_acc_unittest_transpose.cpp
Release/libsmm_acc_unittest_transpose.cpp
Release/libsmm_acc_unittest_transpose.cpp
Debug/libsmm_acc_unittest_transpose.cpp
Debug/libsmm_acc_unittest_transpose.cpp
MinSizeRel/libsmm_acc_unittest_transpose.cpp
MinSizeRel/libsmm_acc_unittest_transpose.cpp
RelWithDebInfo/libsmm_acc_unittest_transpose.cpp
RelWithDebInfo/libsmm_acc_unittest_transpose.cpp
Deployment/libsmm_acc_unittest_transpose.cpp
Deployment/libsmm_acc_unittest_transpose.cpp
Development/libsmm_acc_unittest_transpose.cpp
Development/libsmm_acc_unittest_transpose.cpp
Unable to find executable: libsmm_acc_unittest_transpose.cpp
```

hfp:
> Oof... it might as well be that there is still maintenance going on with the node I'm on 😅 both `clinfo` and `rocm_agent_enumerator` stopped listing the GPU, even though there are still entries in `/dev/dri`. Maybe disregard my 'can reproduce' comment.

Yes, because the benchmark cannot reproduce the problem I mentioned earlier. It must be something on the node, indeed `clinfo` is the right tool to check. This (clinfo) is btw also noted in the install notes for ACC/OpenCL.

> Let's merge it then! There's definitely some improvements in this PR over `develop`, and all sources now use OpenMP when that's enabled.

The scope of this PR increased quite a bit from "simplify CMake for ROCm" to resolving #261. I will merge the PR when CI passes.

> Just to be sure, did you see [70825c3](https://github.com/cp2k/dbcsr/commit/70825c34e8d2713d666fffa25a923a88195c3741) too? It was necessary for me.

I tested AMD's OpenCL stack only under macOS using a Vega56 card, which is probably quite different from Linux (maybe only the ICD is from AMD whereas the "OpenCL platform" comes from Apple; not sure though). The reason for the double quotes was for two-component typenames like "unsigned int" (I played with different foundational types wrt atomics). Anyhow, OpenCL permits typenames like "uint" or "ulong" (quotes are not necessary).

dev-zero:
@hfp @haampie thank you very much for taking care of this and also debugging and properly fixing the CMake configuration!

hfp:
One more note (just for the record), prefixing functions (`c_dbcsr_`) in both benchmark drivers (`acc_bench_trans.c`, and `acc_bench_smm.c`) dropped some code/function calls (`libsmm_acc_init` and `libsmm_acc_finalize`). I will reintroduce this in an upcoming PR (once this PR is merged).

hfp:
Focus now is to fix CI. [GNU](https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-681/gnu.test.out) and [Intel](https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-681/intel.test.out) run-tests point to the same [issue](https://github.com/cp2k/dbcsr/pull/419#issuecomment-773098126).

haampie:
I pressume I can't trigger ci, but let's see:

retest this please

otherwise can someone do it for me?

alazzaro:
retest this please

CMakeLists.txt
```diff
@@ -183,24 +181,18 @@ endif ()
...
+  pkg_check_modules(deps REQUIRED IMPORTED_TARGET GLOBAL libxsmmf)
+  message(STATUS "Using libxsmm for Small Matrix Multiplication")
 else ()
   message(FATAL_ERROR "Unknown SMM library specified")
 endif ()
 
 # =================================== GPU backends
 if (USE_ACCEL MATCHES "opencl")
-  if (NOT LIBXSMM_FOUND)
+  if (NOT USE_SMM)
```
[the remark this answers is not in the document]
haampie: Yeah, that's a mistake. We do `pkg_check_modules(deps REQUIRED IMPORTED_TARGET GLOBAL libxsmmf)` if `USE_SMM MATCHES "libxsmm"`, so I'll just change this into `if (NOT USE_SMM MATCHES "libxsmm") .. message(FATAL_ERROR ...)`. Relying on variables like `LIBXSMM_FOUND` is something I tried to avoid.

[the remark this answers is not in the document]
hfp: ACK, fine with me.

hfp:
Daint-CI seems to be missing once more...

@haampie did the previous test pass with Daint?  
If so, your latest changes should not affect it, and the PR could be merged...

haampie:
I've run the `cmake` + `make` by hand on Daint and it seems to work, so good to go then.

haampie:
Thanks @hfp :)

alazzaro:
I had no time to review the PR before the merge... 
In any case, I left some comments.

Few other remarks here:
1. Is the Daint-CI happy? It seems we didn't run it at the end...
2. I see you replaced .cu with .cpp. Unfortunately, probably this will break the compilation in CP2K with the Makefile, forcing us to move to the cmake (which is good, we have to do that anyway)
3. adding the suffix `c_dbcsr_acc_` is a good choice, but I would have preferred to have change the Fortran name too...

hfp:
Let's not worry, this became brittle since the scope of the PR increased a lot from just CMake for ROCm to some resolution of #261 (touching a lot of source code rather than just CMake stuff). I guess @haampie is happy to help with any other work needed on top of what we got...

alazzaro:
@hfp definitely, another PR is always welcome 😄

haampie:
Regarding

> I see you replaced .cu with .cpp. Unfortunately, probably this will break the compilation in CP2K with the Makefile, forcing us to move to the cmake (which is good, we have to do that anyway)

I found this a bit funny, as it results in cmake switching to the device compiler for that particular source file. I've now disabled the device compiler entirely (that is, CUDA is not an enabled language, and I'm not using hip_add_library for ROCm) since all device code is compiled at runtime anyways. Wasn't aware this caused upstream issues... I could undo it and then set the language of that particular file to CXX so that it uses the right compiler.

> Is the Daint-CI happy? It seems we didn't run it at the end...
I did run it by hand, only for GNU, and it was fine. But maybe good to run it on develop again?

> adding the suffix c_dbcsr_acc_ is a good choice, but I would have preferred to have change the Fortran name too...

Yeah, first I just added `dbcsr_*` but that conflicted with the fortran function names, so I used `c_dbcsr_*`. If you want we can do also change Fortran names in a separate PR?

alazzaro (commented):
> Regarding
> 
> > I see you replaced .cu with .cpp. Unfortunately, probably this will break the compilation in CP2K with the Makefile, forcing us to move to the cmake (which is good, we have to do that anyway)
> 
> I found this a bit funny, as it results in cmake switching to the device compiler for that particular source file. I've now disabled the device compiler entirely (that is, CUDA is not an enabled language, and I'm not using hip_add_library for ROCm) since all device code is compiled at runtime anyways. Wasn't aware this caused upstream issues... I could undo it and then set the language of that particular file to CXX so that it uses the right compiler.

No, that's OK.

> 
> > Is the Daint-CI happy? It seems we didn't run it at the end...
> 
> I did run it by hand, only for GNU, and it was fine. But maybe good to run it on develop again?
> 

Well, it is running in the new PR with your changes, let's see how it goes :smile:

> > adding the suffix c_dbcsr_acc_ is a good choice, but I would have preferred to have change the Fortran name too...
> 
> Yeah, first I just added `dbcsr_*` but that conflicted with the fortran function names, so I used `c_dbcsr_*`. If you want we can do also change Fortran names in a separate PR?

Yeap, name duplication Fortran-C is an issue with the old GNU compiler (fixed with the new compiler). 
Currently, we ask to have the `dbcsr_` prefix in Fortran only for the functions exposed in the API (check [here](https://github.com/cp2k/dbcsr/blob/develop/CONTRIBUTING.md)). The reason was to avoid changing tons of internal code... But now we have a situation where we probably want to have a compatible naming between C and Fortran... In short, yes ;) but no rush...

On the other side, could you fix the two minor comments I left in the code?

cmake/CompilerConfiguration.cmake
```diff
@@ -41,7 +41,7 @@ Please open an issue at https://github.com/cp2k/dbcsr/issues with the reported c
 endif ()
 
 if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-  set(CMAKE_CXX_FLAGS_RELEASE      "-O3 -g -funroll-loops -Wall -Werror")
```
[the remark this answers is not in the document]
haampie: Forgot to reply, but this was suggested above by @dev-zero in the review. I can bring it back, no problem. Maybe it's not very convenient though if compilers we don't test generate warnings and fail to compile as a result.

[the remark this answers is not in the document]
alazzaro: This is actually a good point.. @dev-zero is there a way to add this flag for the CSCS tests? Something during the cmake call...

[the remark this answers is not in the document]
dev-zero: If timing permits, we should run the CI in debug mode, rather than keep `-Werror` in release mode. The rational behind this comes from my distro work: compilers usually get stricter over time with regard to their warnings, hence things tend to break. For the user a warning is a signal that he might want to check for a newer version (where we ideally already fixed that issue since we test in debug), while it still allows him to build this specific version since he may depend on its exact behaviour.
Hence many source-based distros (Gentoo, Arch AUR) tend to strip the `-Werror`.

To specify custom flags we [currently recommend](https://cp2k.github.io/dbcsr/develop/page/2-user-guide/1-installation/1-cmake-build-recipes.html) to use a custom CMake build type:

```
cmake \
         -DCMAKE_BUILD_TYPE=custom \
         -DCMAKE_C_FLAGS_CUSTOM="-O3 -march=native" \
         -DCMAKE_CXX_FLAGS_CUSTOM="-O3 -march=native" \
         -DCMAKE_Fortran_FLAGS_CUSTOM="-O3 -march=native" \
         ..
```
... this is again something I've seen Linux distros doing: to inject their optimization flags, paths, etc. they started to define custom build types for CMake.
But this would mean we'd have to replicate the currently present DEBUG flags.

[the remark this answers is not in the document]
alazzaro: Then no, let's lave -Werror in the release...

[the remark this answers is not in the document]
haampie: Another option is to add a new build type `ReleaseStrict` that inherits all flags from `Release` and adds `-Werror`, but to me it seems like a lot of boilerplate for something trivial. I think easiest is when managers add a patch to drop `-Werror`.

Issue #20, opened 2018-08-20
Modify parameters_K20X.txt to replace “largeDB” by “largeDB1”


alazzaro:

shoshijak:
this would be fixed by PR#25

PR #425, opened 2021-02-04
OpenCL verbose output and documentation, improved auto-tuning scripts, minor fixes after #419

hfp:
OpenCL-BE/LIBSMM: verbose output and documentation. Improved auto-tuning scripts. Minor fixes after #419.
* Fixed Makefile used to build acc_bench_trans/acc_bench_smm with CUDA (accommodate changes from #419).
* Introduced (runtime-)verbosity level. Print device name (non-zero verbosity).
* Fixed issue (https://github.com/cp2k/dbcsr/pull/419#issuecomment-772793542).
* Renamed ACC_OPENCL_VERBOSE to ACC_OPENCL_DEBUG.
* ACC benchmark drivers: inform if no device was found.
* Improved documentation and documented ACC_OPENCL_VERBOSE.
* Introduced verbose output (time needed for kernel compilation, etc).
* tune_multiply.py: option to only rely on primary objective.
* tune_multiply.py: catch CTRL-C and save configuration.
* tune_multiply.sh: relay result code of failing script.
* tune_multiply.sh: continuation with wrapper script.

### DIFF
--- a/src/acc/cuda/Makefile
+++ b/src/acc/cuda/Makefile
@@ -1,6 +1,6 @@
 INCACC := $(wildcard *.h*) ../acc.h
 SRCACC := $(wildcard *.cpp)
-OBJACC := $(SRCACC:.cpp=.o) acc_cublas.o
+OBJACC := $(SRCACC:.cpp=.o)
 
 GPUSMM := $(wildcard ../libsmm_acc/kernels/*.h*)
 INCSMM := $(wildcard ../libsmm_acc/*.h*) ../acc_libsmm.h \
@@ -130,10 +130,7 @@ test: ../dbcsr_acc_test
 ../libsmm_acc/smm_acc_kernels.h: $(GPUSMM) Makefile ../libsmm_acc/generate_kernels.py ../libsmm_acc/parameters/parameters_$(WITH_GPU).json
 	@cd ../libsmm_acc && $(PYTHON) ../libsmm_acc/generate_kernels.py ../libsmm_acc/kernels
 
-acc_cublas.o: acc_cublas.cu Makefile
-	$(NVCC) $(addprefix -Xcompiler $(NULL),$(CXXFLAGS)) -c $< -o $@
-
-../dbcsr_acc.a: $(OBJACC) acc_cublas.o ../libsmm_acc/libsmm_acc_init.o
+../dbcsr_acc.a: $(OBJACC) ../libsmm_acc/libsmm_acc_init.o
 	$(AR) -rs $@ $^
 
 ../dbcsr_acc_smm.a: $(OBJSMM)
@@ -153,7 +150,7 @@ acc_bench_trans.o: ../acc_bench_trans.c Makefile
 	$(CXX) $^ $(LDFLAGS) -o $@
 
 dbcsr_acc_test.o: ../../../tests/dbcsr_acc_test.c Makefile
-	$(CC) $(CFLAGS) -c $< -o $@
+	$(CC) $(CFLAGS) -I../.. -c $< -o $@
 ../dbcsr_acc_test: dbcsr_acc_test.o ../dbcsr_acc_smm.a ../dbcsr_acc.a
 	$(CXX) $^ $(LDFLAGS) -o $@
 
--- a/.ci/daint.cscs.ch/Jenkinsfile
+++ b/.ci/daint.cscs.ch/Jenkinsfile
@@ -66,11 +66,11 @@ pipeline {
                                 run_batch("0:15:00", "ocl", "build")
                             }
                         }
-//                        stage('test') {
-//                            steps {
-//                                run_batch("1:00:00", "ocl", "test")
-//                            }
-//                        }
+                        stage('test') {
+                            steps {
+                                run_batch("1:00:00", "ocl", "test")
+                            }
+                        }
                     }
                 }
                 stage("Intel") {
--- a/.ci/daint.cscs.ch/Jenkinsfile
+++ b/.ci/daint.cscs.ch/Jenkinsfile
@@ -66,11 +66,11 @@ pipeline {
                                 run_batch("0:15:00", "ocl", "build")
                             }
                         }
-                        stage('test') {
-                            steps {
-                                run_batch("1:00:00", "ocl", "test")
-                            }
-                        }
+//                        stage('test') {
+//                            steps {
+//                                run_batch("1:00:00", "ocl", "test")
+//                            }
+//                        }
                     }
                 }
                 stage("Intel") {
--- a/.ci/daint.cscs.ch/cray.build.sh
+++ b/.ci/daint.cscs.ch/cray.build.sh
@@ -1,13 +1,9 @@
 #!/bin/bash -l
 
 #SBATCH --export=ALL
-#SBATCH --exclusive
 #SBATCH --constraint="mc"
 #SBATCH --partition="cscsci"
 #SBATCH --nodes=1
-#SBATCH --ntasks-per-node=4
-#SBATCH --cpus-per-task=3
-#SBATCH --ntasks-per-core=1 # 1=no HT, 2=HT
 
 set -o errexit
 set -o nounset
--- a/.ci/daint.cscs.ch/cray.build.sh
+++ b/.ci/daint.cscs.ch/cray.build.sh
@@ -4,6 +4,9 @@
 #SBATCH --constraint="mc"
 #SBATCH --partition="cscsci"
 #SBATCH --nodes=1
+#SBATCH --ntasks-per-node=4
+#SBATCH --cpus-per-task=3
+#SBATCH --hint=nomultithread
 
 set -o errexit
 set -o nounset
--- a/.ci/daint.cscs.ch/cray.test.sh
+++ b/.ci/daint.cscs.ch/cray.test.sh
@@ -32,4 +32,4 @@ export OMP_PROC_BIND=TRUE # set thread affinity
 env |& tee -a "${STAGE_NAME}.out"
 
 ulimit -s 256000
-env CTEST_OUTPUT_ON_FAILURE=1 make test ARGS="--timeout 900" |& tee -a "${STAGE_NAME}.out"
+env CTEST_OUTPUT_ON_FAILURE=1 make test ARGS="--timeout 1200" |& tee -a "${STAGE_NAME}.out"
--- a/.ci/daint.cscs.ch/cray.test.sh
+++ b/.ci/daint.cscs.ch/cray.test.sh
@@ -1,13 +1,12 @@
 #!/bin/bash -l
 
 #SBATCH --export=ALL
-#SBATCH --exclusive
 #SBATCH --constraint="gpu"
 #SBATCH --partition="cscsci"
 #SBATCH --nodes=1
 #SBATCH --ntasks-per-node=4
 #SBATCH --cpus-per-task=3
-#SBATCH --ntasks-per-core=1 # 1=no HT, 2=HT
+#SBATCH --hint=nomultithread
 
 set -o errexit
 set -o nounset
@@ -32,4 +31,4 @@ export OMP_PROC_BIND=TRUE # set thread affinity
 env |& tee -a "${STAGE_NAME}.out"
 
 ulimit -s 256000
-env CTEST_OUTPUT_ON_FAILURE=1 make test ARGS="--timeout 1200" |& tee -a "${STAGE_NAME}.out"
+env CTEST_OUTPUT_ON_FAILURE=1 make test ARGS="--timeout 900" |& tee -a "${STAGE_NAME}.out"
--- a/.ci/daint.cscs.ch/gnu.build.sh
+++ b/.ci/daint.cscs.ch/gnu.build.sh
@@ -1,13 +1,9 @@
 #!/bin/bash -l
 
 #SBATCH --export=ALL
-#SBATCH --exclusive
 #SBATCH --constraint="mc"
 #SBATCH --partition="cscsci"
 #SBATCH --nodes=1
-#SBATCH --ntasks-per-node=4
-#SBATCH --cpus-per-task=3
-#SBATCH --ntasks-per-core=1 # 1=no HT, 2=HT
 
 set -o errexit
 set -o nounset
--- a/.ci/daint.cscs.ch/gnu.build.sh
+++ b/.ci/daint.cscs.ch/gnu.build.sh
@@ -4,6 +4,9 @@
 #SBATCH --constraint="mc"
 #SBATCH --partition="cscsci"
 #SBATCH --nodes=1
+#SBATCH --ntasks-per-node=4
+#SBATCH --cpus-per-task=3
+#SBATCH --hint=nomultithread
 
 set -o errexit
 set -o nounset
--- a/.ci/daint.cscs.ch/gnu.test.sh
+++ b/.ci/daint.cscs.ch/gnu.test.sh
@@ -32,4 +32,4 @@ export OMP_PROC_BIND=TRUE # set thread affinity
 # document the current environment
 env |& tee -a "${STAGE_NAME}.out"
 
-env CTEST_OUTPUT_ON_FAILURE=1 make test ARGS="--timeout 900" |& tee -a "${STAGE_NAME}.out"
+env CTEST_OUTPUT_ON_FAILURE=1 make test ARGS="--timeout 1200" |& tee -a "${STAGE_NAME}.out"
--- a/.ci/daint.cscs.ch/gnu.test.sh
+++ b/.ci/daint.cscs.ch/gnu.test.sh
@@ -1,13 +1,12 @@
 #!/bin/bash -l
 
 #SBATCH --export=ALL
-#SBATCH --exclusive
 #SBATCH --constraint="gpu"
 #SBATCH --partition="cscsci"
 #SBATCH --nodes=1
 #SBATCH --ntasks-per-node=4
 #SBATCH --cpus-per-task=3
-#SBATCH --ntasks-per-core=1 # 1=no HT, 2=HT
+#SBATCH --hint=nomultithread
 
 set -o errexit
 set -o nounset
@@ -32,4 +31,4 @@ export OMP_PROC_BIND=TRUE # set thread affinity
 # document the current environment
 env |& tee -a "${STAGE_NAME}.out"
 
-env CTEST_OUTPUT_ON_FAILURE=1 make test ARGS="--timeout 1200" |& tee -a "${STAGE_NAME}.out"
+env CTEST_OUTPUT_ON_FAILURE=1 make test ARGS="--timeout 900" |& tee -a "${STAGE_NAME}.out"
--- a/.ci/daint.cscs.ch/intel.build.sh
+++ b/.ci/daint.cscs.ch/intel.build.sh
@@ -1,13 +1,9 @@
 #!/bin/bash -l
 
 #SBATCH --export=ALL
-#SBATCH --exclusive
 #SBATCH --constraint="mc"
 #SBATCH --partition="cscsci"
 #SBATCH --nodes=1
-#SBATCH --ntasks-per-node=4
-#SBATCH --cpus-per-task=3
-#SBATCH --ntasks-per-core=1 # 1=no HT, 2=HT
 
 set -o errexit
 set -o nounset
--- a/.ci/daint.cscs.ch/intel.build.sh
+++ b/.ci/daint.cscs.ch/intel.build.sh
@@ -4,6 +4,9 @@
 #SBATCH --constraint="mc"
 #SBATCH --partition="cscsci"
 #SBATCH --nodes=1
+#SBATCH --ntasks-per-node=4
+#SBATCH --cpus-per-task=3
+#SBATCH --hint=nomultithread
 
 set -o errexit
 set -o nounset
--- a/.ci/daint.cscs.ch/intel.test.sh
+++ b/.ci/daint.cscs.ch/intel.test.sh
@@ -32,4 +32,4 @@ export OMP_PROC_BIND=TRUE # set thread affinity
 # document the current environment
 env |& tee -a "${STAGE_NAME}.out"
 
-env CTEST_OUTPUT_ON_FAILURE=1 make test ARGS="--timeout 900" |& tee -a "${STAGE_NAME}.out"
+env CTEST_OUTPUT_ON_FAILURE=1 make test ARGS="--timeout 1200" |& tee -a "${STAGE_NAME}.out"
--- a/.ci/daint.cscs.ch/intel.test.sh
+++ b/.ci/daint.cscs.ch/intel.test.sh
@@ -1,13 +1,12 @@
 #!/bin/bash -l
 
 #SBATCH --export=ALL
-#SBATCH --exclusive
 #SBATCH --constraint="gpu"
 #SBATCH --partition="cscsci"
 #SBATCH --nodes=1
 #SBATCH --ntasks-per-node=4
 #SBATCH --cpus-per-task=3
-#SBATCH --ntasks-per-core=1 # 1=no HT, 2=HT
+#SBATCH --hint=nomultithread
 
 set -o errexit
 set -o nounset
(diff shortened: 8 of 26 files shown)


src/acc/cuda/Makefile
```diff
@@ -153,7 +150,7 @@ acc_bench_trans.o: ../acc_bench_trans.c Makefile
 	$(CXX) $^ $(LDFLAGS) -o $@
 
 dbcsr_acc_test.o: ../../../tests/dbcsr_acc_test.c Makefile
-	$(CC) $(CFLAGS) -c $< -o $@
+	$(CC) $(CFLAGS) -I../.. -c $< -o $@
```
haampie, line 153: ah yes, i was meaning to add this too :p forgot about it

hfp: No worries.

hfp:
There will be one more change for this PR, which disables ACC_OPENCL_THREADLOCAL_CONTEXT feature in OpenCL backend.

hfp:
I will also try enabling runtime tests for the OpenCL backend on Daint-CI.

alazzaro:
For some reason, Daint-CI cannot run a test in 900s. @hfp could you try to increase the time limit (1200s?) ?

alazzaro:
> @haampie as a carry-forward from #419, it seems like we have some (unwanted) debug output like:
> 
> ```
> 20/23 Test #20: libsmm_acc_unittest_multiply ..........................***Timeout 900.10 sec
> Thu Feb  4 14:57:50 2021: [unset]:_pmi_alps_init:alps_get_placement_info returned with error -1
> Thu Feb  4 14:57:50 2021: [unset]:_pmi_init:_pmi_alps_init returned -1
> # libsmm_acc has 2649 blocksizes for multiplication
> OK 4 x 4 x 4
> OK 4 x 4 x 5
> OK 4 x 4 x 6
> [...]
> ```
> 
> ... which may cause tests to time-out. I have not noticed the "OK _m_ x _n_ x _k_" output before.

Ah, good spot!

hfp:
> 
> 
> For some reason, Daint-CI cannot run a test in 900s. @hfp could you try to increase the time limit (1200s?) ?

Let me try fixing it as part of this PR.

hfp:
> 
> 
> > @haampie as a carry-forward from #419, it seems like we have some (unwanted) debug output like:
> > ```
> > 20/23 Test #20: libsmm_acc_unittest_multiply ..........................***Timeout 900.10 sec
> > Thu Feb  4 14:57:50 2021: [unset]:_pmi_alps_init:alps_get_placement_info returned with error -1
> > Thu Feb  4 14:57:50 2021: [unset]:_pmi_init:_pmi_alps_init returned -1
> > # libsmm_acc has 2649 blocksizes for multiplication
> > OK 4 x 4 x 4
> > OK 4 x 4 x 5
> > OK 4 x 4 x 6
> > [...]
> > ```
> > 
> > 
> > ... which may cause tests to time-out. I have not noticed the "OK _m_ x _n_ x _k_" output before.
> 
> Ah, good spot!

It seems to come from [tests/libsmm_acc_unittest_multiply.cpp.template](https://github.com/cp2k/dbcsr/blob/develop/tests/libsmm_acc_unittest_multiply.cpp.template#L34), i.e., subsequent calls contained there like `libsmm_acc_benchmark`. I will simply increase permitted runtime to 1200s (perhaps we never built/ran these tests that way?).

haampie:
Not sure if relevant, but these particular tests are executed without the `srun ...` prefix is what I found curious when I looked into the cmake code

alazzaro:
> > > @haampie as a carry-forward from #419, it seems like we have some (unwanted) debug output like:
> > > ```
> > > 20/23 Test #20: libsmm_acc_unittest_multiply ..........................***Timeout 900.10 sec
> > > Thu Feb  4 14:57:50 2021: [unset]:_pmi_alps_init:alps_get_placement_info returned with error -1
> > > Thu Feb  4 14:57:50 2021: [unset]:_pmi_init:_pmi_alps_init returned -1
> > > # libsmm_acc has 2649 blocksizes for multiplication
> > > OK 4 x 4 x 4
> > > OK 4 x 4 x 5
> > > OK 4 x 4 x 6
> > > [...]
> > > ```
> > > 
> > > 
> > > ... which may cause tests to time-out. I have not noticed the "OK _m_ x _n_ x _k_" output before.
> > 
> > 
> > Ah, good spot!
> 
> It seems to come from [tests/libsmm_acc_unittest_multiply.cpp.template](https://github.com/cp2k/dbcsr/blob/develop/tests/libsmm_acc_unittest_multiply.cpp.template#L34), i.e., subsequent calls contained there like `libsmm_acc_benchmark`. I will simply increase permitted runtime to 1200s (perhaps we never built/ran these tests that way?).

Well, no idea.. But I would suggest commenting the output (leave it only for Debug) if it makes the test running faster... Daint-CI has a limited budget for us...

alazzaro:
> Not sure if relevant, but these particular tests are executed without the `srun ...` prefix is what I found curious when I looked into the cmake code

Really? This is a GPU test, cannot run on the frontend node... We provide srun via cmake, i.e.:

`-DMPIEXEC_EXECUTABLE="$(command -v srun)" \`

haampie:
Yeah, I was aware, earlier this morning I learned that `sbatch` executes the bash script on the compute node in fact, so therefore it works. Which made me think it was intentional to not wrap these tests in `[srun command] [test]`

dev-zero:
> Yeah, I was aware, earlier this morning I learned that `sbatch` executes the bash script on the compute node in fact, so therefore it works. Which made me think it was intentional to not wrap these tests in `[srun command] [test]`

Given that the `libsmm_acc` tests all directly use `printf` without checking whether they are on the first node, my guess is that they were intended to be run on single-node only.
That should be rectified I guess (both the code and the CMake command).

Maybe add a wrapper for `add_test(...)` which contains the `if (USE_MPI)` to keep it DRY?

hfp:
> 
> 
> > Yeah, I was aware, earlier this morning I learned that `sbatch` executes the bash script on the compute node in fact, so therefore it works. Which made me think it was intentional to not wrap these tests in `[srun command] [test]`
> 
> Given that the `libsmm_acc` tests all directly use `printf` without checking whether they are on the first node, my guess is that they were intended to be run on single-node only.
> That should be rectified I guess (both the code and the CMake command).
> 
> Maybe add a wrapper for `add_test(...)` which contains the `if (USE_MPI)` to keep it DRY?

If my "hack" (reducing output) goes through with Daint-CI, I will leave this to another PR.

dev-zero:
sure, I just opened #427 to track it

alazzaro:
I wonder why the test cannot complete in 900s anymore. This is the last week output:

https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-679/

The infamous test complete in 629.62s.

hfp:
> 
> 
> I wonder why the test cannot complete in 900s anymore. This is the last week output:
> 
> https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-679/
> 
> The infamous test complete in 629.62s.

Perhaps a diff between the build-logs can help if this is caused by compiler flags or such.

alazzaro:
Well the only changes since last week and now are due to the HIP PR...
@haampie could you take a look and make a diff between the two versions?

hfp:
> 
> 
> Well the only changes since last week and now are due to the HIP PR...
> @haampie could you take a look and make a diff between the two versions?

I just had a look at `libsmm_acc_unittest_multiply` and how the compile/link line changed from 679 and 685. Compilation flags are the same except `-Werror`. The main difference is previously (679) everything possible was linked statically. Further, 679 was relying on `/opt/nvidia/cudatoolkit10.2/10.2.89_3.28-7.0.2.1_2.17__g52c0314/targets/x86_64-linux/lib/stubs` whereas 685 used the non-stubs library directory (probably related to static linkage?). Also, 679 linked against `CMakeFiles/libsmm_acc_unittest_multiply.dir/cmake_device_link.o`, `libcudadevrt`, `librt`, and `libdl` (which are not present on 685's link-line). However, 685 links against `libnvToolsExt`, which was not present on 679's link-line.

alazzaro:
> > Well the only changes since last week and now are due to the HIP PR...
> > @haampie could you take a look and make a diff between the two versions?
> 
> I just had a look at `libsmm_acc_unittest_multiply` and how the compile/link line changed from 679 and 685. Compilation flags are the same except `-Werror`. The main difference is previously (679) everything possible was linked statically. Further, 679 was relying on `/opt/nvidia/cudatoolkit10.2/10.2.89_3.28-7.0.2.1_2.17__g52c0314/targets/x86_64-linux/lib/stubs` whereas 685 used the non-stubs library directory (probably related to static linkage?). Also, 679 linked against `CMakeFiles/libsmm_acc_unittest_multiply.dir/cmake_device_link.o`, `libcudadevrt`, `librt`, and `libdl` (which are not present on 685's link-line). However, 685 links against `libnvToolsExt`, which was not present on 679's link-line.

Wait, libnvToolsExt is used? This is the profiler API that can explain the reason for the slowness... 
However, the code should be protected with a macro...

hfp:
> 
> 
> I will also try enabling runtime tests for the OpenCL backend on Daint-CI.

For reference, the runtime tests on Daint-CI fail like in [build-686](https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-686/ocl.test.out). However, things work fine when manually tested on Daint (user account).

alazzaro:
> > I will also try enabling runtime tests for the OpenCL backend on Daint-CI.
> 
> For reference, the runtime tests on Daint-CI fail like in [build-686](https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-686/ocl.test.out). However, things work fine when manually tested on Daint (user account).

Same configuration? Interesting... Can you try to run a single rank/thread in the CI?

haampie:
I'll have a look tomorrow morning. I'm fine if you temporarily revert the pr

alazzaro:
> I'll have a look tomorrow morning. I'm fine if you temporarily revert the pr

No no, the PR is fine, just need some more checking...

alazzaro:
> > I will also try enabling runtime tests for the OpenCL backend on Daint-CI.
> 
> For reference, the runtime tests on Daint-CI fail like in [build-686](https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-686/ocl.test.out). However, things work fine when manually tested on Daint (user account).

I can reproduce the error on Daint with the same CI configuration. 
I've tried a single rank and thread and it still hit the same problem.
Can you elaborate on "manually tested on Daint (user account)"?

hfp:
> 
> 
> > > I will also try enabling runtime tests for the OpenCL backend on Daint-CI.
> > 
> > 
> > For reference, the runtime tests on Daint-CI fail like in [build-686](https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-686/ocl.test.out). However, things work fine when manually tested on Daint (user account).
> 
> I can reproduce the error on Daint with the same CI configuration.
> I've tried a single rank and thread and it still hit the same problem.
> Can you elaborate on "manually tested on Daint (user account)"?

I will take a look again and document the steps if it runs successfully.

hfp:
> 
> 
> instead of a `checkout` I usually see a `git reset --hard ...` in git syncers, just to ensure that all state is reset in case there was a change (possibly also caused due to interrupted process, a rebase on upstream, etc.)

No, this was not the case. I was confused also. Root cause was I referred previously/initially to a different SHA (which was present since the clone was fresh). However, I updated the SHA and got the error, which was simply caused by the SHA not present (yet) in the repository. So, updating to a newer SHA requires to fetch the new state/SHAs first.

hfp:
> Can you elaborate on "manually tested on Daint (user account)"?

I experimented a bit and maybe CUDA MPS is related. At least `srun --ntasks 1 ...` (or equivalent `salloc`) get things working. I ran multiple MPI-ranks on different systems with CUDA devices but in none of the cases MPS was enabled. I will downgrade the runtime tests for the OpenCL backend to single-rank as a first try.

hfp:
Only CodeCov "fails", i.e., OpenCL runtime tests are now passing with one rank only (see [here](https://github.com/cp2k/dbcsr/pull/425#issuecomment-774088398) for some early reasoning). It seems Daint runs the GPUs in "exclusive" mode which seems to only permit multiple ranks via MPS, i.e., creating a second OpenCL context on the same device (single node!) always fails like "device not available".

dev-zero:
wrt CodeCov: the total coverage consists of different runs/uploads (with, w/o mpi; with w/o omp), and the PR entry often happens too early. We could tune it to [wait for all uploads before posting](https://docs.codecov.io/docs/pull-request-comments#after_n_builds) or [create new posts](https://docs.codecov.io/docs/pull-request-comments#behavior) instead of partially updating the existing one.

wrt single-ranks: there is no extra CMake magic involved when it comes to deciding when to run tests, but the following determines the number of ranks *at build time*:

```
    -DTEST_MPI_RANKS=${SLURM_NTASKS} \
    -DTEST_OMP_THREADS=${SLURM_CPUS_PER_TASK} \
```

As mentioned in #427, the `libsmm_acc` tests are currently not run through `srun` (or `mpirun`) at all.

hfp:
Thank you for following up, Tiziano!

> wrt CodeCov: the total coverage consists of different runs/uploads (with, w/o mpi; with w/o omp), and the PR entry often happens too early. We could tune it to [wait for all uploads before posting](https://docs.codecov.io/docs/pull-request-comments#after_n_builds) or [create new posts](https://docs.codecov.io/docs/pull-request-comments#behavior) instead of partially updating the existing one.

I wonder if we should drop CodeCov? I think it produces mostly noise (nothing actionable). Primarily, we are happy if the project receives a contribution, secondly, we only take PRs and then review/guide the code and format. I believe this process would never pass something unreasonable into the code base. Overall I have not seen anyone of us judging a contribution by the percentage of coverage. With the current code base, it seems a coverage of ~60% is intrinsic to the project or code base. Perhaps it is even possible to just calculate the percentage (but no threshold making things red and sending emails).
 
> wrt single-ranks: there is no extra CMake magic involved when it comes to deciding when to run tests, but the following determines the number of ranks _at build time_:
> 
> ```
>     -DTEST_MPI_RANKS=${SLURM_NTASKS} \
>     -DTEST_OMP_THREADS=${SLURM_CPUS_PER_TASK} \
> ```
> 
> As mentioned in #427, the `libsmm_acc` tests are currently not run through `srun` (or `mpirun`) at all.

I just noticed that the OpenCL runtime test not only passes (see [above](https://github.com/cp2k/dbcsr/pull/425#issuecomment-774419906) for the root cause), but unfortunately ran [less tests](https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-696/gnu.test.out) (19) compared to [other](https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-696/gnu.test.out) (23) runtime tests. The only difference seems to be the rank count. I understand that some tests are currently not running per `srun`, but should this not apply to all runtime tests equally no matter what the requested rank-count was?

alazzaro:
@hfp I assume it is ready for review, I will do it today...

> I wonder if we should drop CodeCov? 

We should keep it, actually I plan to introduce new tests (long project) and add the coverage for the GPU. I like @dev-zero idea...

> It seems Daint runs the GPUs in "exclusive" mode which seems to only permit multiple ranks via MPS, i.e., creating a second OpenCL context on the same device (single node!) always fails like "device not available".

Should we open a ticket with CSCS? I can try on some other machines, I wonder what's the rationale behind that... I can assume not so many people are using OpenCL.

> unfortunately ran less tests (19) compared to other (23) runtime tests.

Interesting... I actually see the same number of tests...

dev-zero:
> I wonder if we should drop CodeCov? I think it produces mostly noise (nothing actionable). Primarily, we are happy if the project receives a contribution, secondly, we only take PRs and then review/guide the code and format. I believe this process would never pass something unreasonable into the code base. Overall I have not seen anyone of us judging a contribution by the percentage of coverage. With the current code base, it seems a coverage of ~60% is intrinsic to the project or code base. Perhaps it is even possible to just calculate the percentage (but no threshold making things red and sending emails).

The "sending emails" part is done automatically by GH when the CodeCov-bot adds the comment.
I agree that it should be tweaked further to generate less noise, but if we only let it report a percentage in the list of tests without further note, nobody is going to look at it anymore at all.

So, I would propose that I change it to wait for all reports before sending it first and changing the relative threshold to something which allows to add some uncovered fixes without error but when adding larger parts they have to be covered, and a lower bound of 60% because that's what we established now.

> I just noticed that the OpenCL runtime test not only passes (see [above](https://github.com/cp2k/dbcsr/pull/425#issuecomment-774419906) for the root cause), but unfortunately ran [less tests](https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-696/gnu.test.out) (19) compared to [other](https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-696/gnu.test.out) (23) runtime tests. The only difference seems to be the rank count. I understand that some tests are currently not running per `srun`, but should this not apply to all runtime tests equally no matter what the requested rank-count was?

Are you sure it's not this line here:

https://github.com/cp2k/dbcsr/blob/dcbc5f6cf77d5516266f5c0b9bc7c91c94f3b67f/tests/CMakeLists.txt#L216

... which should probably be just `if (USE_ACCEL)` again.

hfp:
Regarding multiple ranks on a single card, I guess this can be adjusted with the SMI utility and there is likely no performance regression associated with toggling it from exclusive to shared (and MPS can do whatever it does). However, this is reconfiguring Daint and may not be a realistic request. MPS on the other does not support OpenCL (as per some answer in some Nvidia forum). Another option could be to test multiple ranks using a single rank per multiple nodes, which should work.

hfp:
> which should probably be just if (USE_ACCEL) again.

Thanks for finding, I overlooked this completely!  I guess it should stay since these tests are specific to CUDA/HIP and the implementing of auto-tuning for this backend. If that was written using the ACC interface it would have been easier, but on the other hand we can play with the OpenTuner based auto-tuning and perhaps it is fruitful to have something else.

hfp:
If this passes, I probably want to add an error message explaining a failure when creating the OpenCL context or at least hint the MPS/SMI exclusive thing.

alazzaro:
> If this passes, I probably want to add an error message explaining a failure when creating the OpenCL context or at least hint the MPS/SMI exclusive thing.

OK, please let me know when it is time to review...

hfp:
@alazzaro the PR is ready for review.

alazzaro:
@hfp For my understanding is the solution proposed here 

https://gerrit.gromacs.org/c/gromacs/+/5729/
https://gerrit.gromacs.org/c/gromacs/+/5780/7

I cannot find anything else on the topic (OpenCL with multiprocesses)...

hfp:
> 
> 
> @hfp For my understanding is the solution proposed here
> 
> https://gerrit.gromacs.org/c/gromacs/+/5729/
> https://gerrit.gromacs.org/c/gromacs/+/5780/7
> 
> I cannot find anything else on the topic (OpenCL with multiprocesses)...

I found different stuff including to prefer `clCreateContextFromType` ("due to a bug") over `clCreateContext`. I tried this and only got a different error code like `CL_DEVICE_NOT_AVAILABLE` rather than `CL_INVALID_DEVICE`. Your finding looks more promising and potentially more applicable. I wonder if this would only affect the backend's code base or if this is a more general change? Currently, the BE code does not use MPI or broadcasts/shares anything. From the BE's perspective, all device discovery is process-specific and local.

alazzaro:
OK, then let's keep a single rank per device and open another issue for further investigation, probably employing the intranode broadcast. It seems that this is a common issue, other codes (CENAERO and SPECFEM3D) have similar problems, but Gromas seems to have found a solution...
I will review today.

hfp:
I checked on a Daint/GPU node like `srun nvidia-smi -q -d COMPUTE"` and it shows:

```text
==============NVSMI LOG==============

Timestamp                           : Mon Feb  8 11:20:52 2021
Driver Version                      : 440.33.01
CUDA Version                        : 10.2

Attached GPUs                       : 1
GPU 00000000:02:00.0
    Compute Mode                    : Exclusive_Process
```

hfp:
> probably employing the intranode broadcast.

They did a broadcast initially and the fix was to discover devices on a per rank basis. We already do the right thing then and it seems to be all about the [exclusive mode](https://github.com/cp2k/dbcsr/pull/425#issuecomment-775038210), but let's investigate over the course of today...

alazzaro:
LGTM, Daint-CI with Intel failed to submit the job. I've restarted the job.

So, I suggest opening an issue for the OpenCL and multiple ranks per device, as we discussed (unless you have a very last-minute solution, I would keep the discussion for a different PR).

I can have some naive questions (apologies in advance):

1. Can we use OpenCL and CUDA at the same time? Let me explain better what I have in mind: let's assume we are running on nvidia gpu with OpenCL and we ask for a kernel that is not implemented, can we fall-back to the CUDA side? I can assume it will be a mess to organize memory and so on... Another possible solution would be to have OpenCL for single precision (great that you have implemented it) and CUDA for DP (but of course we can go for all with OpenCL...)
2.  Concerning the SP part, we probably have to hack the fortran code to allow the SP part for ACC. I can remember (but maybe I'm wrong) that it is only DP support in several places...
2. What about the CUBLAS backend, is there anything similar in OpenCL?
3. Do you have a performance comparison CUDA/OpenCL or HIP/OpenCL?

alazzaro:
New tests on Daint-CI wenf all fine...

hfp:
> LGTM, Daint-CI with Intel failed to submit the job. I've restarted the job.

I will merge asap.

> So, I suggest opening an issue for the OpenCL and multiple ranks per device, as we discussed (unless you have a very last-minute solution, I would keep the discussion for a different PR).

ACK

Thank you for the good questions! I will try answer them one by one (see below).

> Can we use OpenCL and CUDA at the same time? Let me explain better what I have in mind: let's assume we are running on nvidia gpu with OpenCL and we ask for a kernel that is not implemented, can we fall-back to the CUDA side? I can assume it will be a mess to organize memory and so on... Another possible solution would be to have OpenCL for single precision (great that you have implemented it) and CUDA for DP (but of course we can go for all with OpenCL...)

Of course, it's doable, but likely quite some effort since we never accounted for multiple backends.

> Concerning the SP part, we probably have to hack the fortran code to allow the SP part for ACC. I can remember (but maybe I'm wrong) that it is only DP support in several places...

Joost quite some time ago (obviously) confirmed SP at least for CPU/LIBXSMM (out of the box), but I cannot remember what/how he tested. Since a lot of code in CP2K is hard-coding DP, it was probably just plain/straight-forward QS test (maybe Water ;-). For SP in general, we probably want to gather workforce and identify if/where this is beneficial and what to enable first. This also touches enabling low/mixed-precision with SP as viable special case. DBCSR stand-alone "workloads" and all (unit-)tests already work with SP. On DBCSR's side, we can think if something else should offered like "internal type" w/ copy-in/out from higher precision, etc. Maybe like the feature about reduced MPI-traffic based on SP.

> What about the CUBLAS backend, is there anything similar in OpenCL?

I have implemented SVM support already, i.e., LIBSMM can leverage calls to some other implementation like MKL. This might be similar to the CUDA/HIP/SMM_ACC (except they need this host-side stack; not sure if other GEMM-args are fed from the host or if this also relies on device-pointers like SVM). So far, this might be useful for (very) large kernels, but otherwise (CUDA/HIP) this is inefficient since the stack is processed on the host with all singular SMMs queued on the device-side. I would rather like to call `dgemm_batch`, however, this does not specify what happens for duplicated C-indexes (beside of taking an array of pointers). I may ask MKL team to revisit such case. Btw, LIBXSMM implements parallelized batches SMMs with C-matrix sync (which is used by benchmark drivers). To somwwhat answer your original question, calling MKL for instance with data already populated on the device will be easy given SVM is enabled (currently disabled in OpenCL backend).

> Do you have a performance comparison CUDA/OpenCL or HIP/OpenCL?

At the moment, I believe HIP/OpenCL should deliver the best insight with respect to what this backend is capable of. @haampie wanted to look at some AMD GPUs; perhaps I can help/do this as well (if accessible per my CSCS account). Otherwise (Nvidia devices), the gap using OpenCL vs CUDA is reaching up to 2x with current kernels. I will improve kernels, which are by no means sophisticated yet on the OpenCL side. To get to your question, you can send me an email; I have numbers for P100, V100, A100, and perhaps others.

alazzaro:
Please merge and thanks for the replies!
One point to keep in mind concerning SP: of course DBCSR allows SP and DP (real and complex). However, only DP is available on the GPU. In the Fortran code, we have some flags to check for that... What Joost did was only for the CPU...

PR #426, opened 2021-02-04
Update some versions in the docs

haampie:
CMake 3.17 is now a requirement, and the Daint installation info was outdated.

Should we also add `spack install` instructions? There is no mention of it, but basically what I want is:

```shell
spack install dbcsr +rocm              # will build a lot of things including llvm-amdgpu and rocblas
spack install dbcsr +cuda              # will just download cuda
spack install dbcsr +opencl ^cuda      # possible since yesterday
spack install dbcsr +opencl ^amdopencl # there is an AMD developer wrapping this package, to be released soon
```

### DIFF
--- a/docs/guide/2-user-guide/1-installation/1-cmake-build-recipes.md
+++ b/docs/guide/2-user-guide/1-installation/1-cmake-build-recipes.md
@@ -158,8 +158,8 @@ selected BLAS/LAPACK library accordingly (e.g. drop the `_mpi`, or `_mp`).
 
 ### CCE and libsci-cray
 
-Verified on CSCS' Piz Daint with CCE 9.0.2 and cray-libsci 19.06.1,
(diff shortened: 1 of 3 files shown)


alazzaro:
@dev-zero any comment on the spack part?

dev-zero:
> CMake 3.17 is now a requirement, and the Daint installation info was outdated.
> 
> Should we also add `spack install` instructions? There is no mention of it, but basically what I want is:
> 
> ```shell
> spack install dbcsr +rocm              # will build a lot of things including llvm-amdgpu and rocblas
> spack install dbcsr +cuda              # will just download cuda
> spack install dbcsr +opencl ^cuda      # possible since yesterday
> spack install dbcsr +opencl ^amdopencl # there is an AMD developer wrapping this package, to be released soon
> ```

Sure!

ACK.

PR #431, opened 2021-02-08
Remove error borders in docs example code

haampie:
Bit of a trivial PR, but there are ugly looking borders around every line of example code here: https://cp2k.github.io/dbcsr/develop/page/2-user-guide/1-installation/index.html. So my guess is you have to specify the language to something.


hfp:
I some times use "text" if it is not "bash" like:

```text
some text
which is not
Bash
```

PR #432, opened 2021-02-09
Docker image maintenance

### LINKED ISSUE #430: ghi: move ROCm image to our hub.docker.com org
The ROCm build tests currently uses an image from @shoshijak, we should migrate that to our Docker hub and rebuild/update the image periodically like we do with the others.

Note: the [Docker Hub Repo](https://hub.docker.com/r/dbcsr/build-env-rocm-ubuntu-18.04) is already there and GHI should have push access, so the only things remaining are:

* [ ] add job entry in `.github/workflows/docker-build-env.yml` for the ROCm Docker in `tools/docker/Dockerfile.build-env-rocm`
* [ ] change the reference docker repo in `.github/workflows/testing-linux.yml`

haampie:
Closes #430 

- Bump cmake patch version
- Bump the rocm image to use 4.0.1 (seems like 4.0 == 4.0.0)
- Periodically build the ROCm Docker image
- Start using the new image

### DIFF
--- a/tools/docker/Dockerfile.build-env-rocm
+++ b/tools/docker/Dockerfile.build-env-rocm
@@ -1,43 +1,24 @@
-FROM rocm/dev-ubuntu-18.04:latest
+FROM rocm/dev-ubuntu-20.04:4.0.1
+
+ENV DEBIAN_FRONTEND=noninteractive
 
 # install compilers, libraries & co
-RUN apt-get update
-RUN apt-get install -y \
-        locales \
-        gfortran \
-        gcc-7 \
-        g++-7 \
-        llvm-7-dev \
-        llvm-7-tools \
-        mpich \
-        libomp-7-dev \
+RUN apt-get update -qq \
+    && apt-get install --no-install-recommends -qq \
+        ca-certificates \
+        gfortran g++ gcc \
+        git \
+        hipblas \
         libmpich-dev \
-        libopenblas-dev \
-        wget
-
-# install rocm libraries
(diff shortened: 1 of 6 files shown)


tools/docker/Dockerfile.build-env-rocm
```diff
@@ -1,43 +1,24 @@
...
-
-# install cmake
-RUN wget https://github.com/Kitware/CMake/releases/download/v3.17.0/cmake-3.17.0-Linux-x86_64.tar.gz
-RUN tar -xzvf cmake-3.17.0-Linux-x86_64.tar.gz
-ENV PATH="/cmake-3.17.0-Linux-x86_64/bin:${PATH}"
+        wget \
+        python3-pip \
+    && rm -rf /var/lib/apt/lists/* \
+    && wget -qO- "https://cmake.org/files/v3.17/cmake-3.17.5-Linux-x86_64.tar.gz" | \
+       tar --strip-components=1 -xz -C /usr/local \
```
dev-zero, line 22: just discovered this one here: https://apt.kitware.com/ ... what do you think?

haampie: Hm yeah, I think it can't hurt to install a specific version and I'm too lazy to figure out how you would pin it with apt

haampie:
Should this be merged @dev-zero ?

dev-zero:
@haampie  sorry for not responding :( I went a different way in the end by using the Kitware Ubuntu repository instead (and since `gcc:latest` is based on Debian Testing it is recent enough wrt CMake and Ninja). Also, building ROCm is now also integrated, but for hosting the images we switched to ghcr.io: https://github.com/orgs/cp2k/packages

PR #433, opened 2021-02-10
ocl: expose multiple tiles/subdevices as devices

hfp:
Attempt to (optionally) split devices into subdevices (`ACC_OPENCL_DEVSPLIT`). This enables CP2K/DBCSR to load MPI-ranks onto subdevices (simply discovered as devices as per `c_dbcsr_acc_get_ndevices`). Note such device is treated as an aggregate device by default and `ACC_OPENCL_DEVSPLIT` is completely optional.

* Raised ACC_OPENCL_DEVICES_MAXCOUNT to better accommodate subdevices.
* Minor rephrase; removed red boxes around verbatim section (documentation).
* Documented ACC_OPENCL_DEVSPLIT.

Unrelated: avoid warning about (invalid) "#pragma OPENCL EXTENSION all: enable" (it can only be "disable" according to OpenCL spec.). However, for some OpenCL stacks this comes handy and without this warning. Main point is most platforms enable all extensions by default but at least one not.

### DIFF
--- a/src/acc/opencl/README.md
+++ b/src/acc/opencl/README.md
@@ -16,9 +16,10 @@ An application of compile-time settings (and perhaps a valuable contribution) mi
 
 Runtime settings are made by the means of environment variables (implemented in `acc_opencl.c`). There are variables for chosing an OpenCL device:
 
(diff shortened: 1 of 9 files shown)


hfp:
Daint seems to be still on maintenance...

PR #434, opened 2021-02-19
oclacc: refined GPU order, make use of atomic_add when possible

hfp:
Enabled FP atomic-add when permitted/supported (device specific)

* Introduced c_dbcsr_acc_opencl_device_name to confirm the name of the given device.
* Determine host-device uniform memory (c_dbcsr_acc_opencl_info_devmem).
* Refined handling ACC_OPENCL_DEVTYPE (cpu, gpu, acc, all, other).
* Refined pruning device list (all kinds of devices).
* Adjusted sorting GPUs (device order).

Improved tune_multiply.py

* Avoid potential error due to reading file still open from write.
* Extended error/exception handler to cover file-read errors.
* Made TYPEID a numeric value (int).

### DIFF
--- a/src/acc/opencl/smm/tune_multiply.py
+++ b/src/acc/opencl/smm/tune_multiply.py
@@ -44,7 +44,7 @@ class SmmTuner(MeasurementInterface):
             match = None
         if (match is not None) and match.group(1) and match.group(2):
             self.typename = match.group(2)
-            self.typeid = match.group(1)
+            self.typeid = int(match.group(1))
         else:
             sys.tracebacklimit = 0
             raise RuntimeError(
@@ -128,14 +128,14 @@ class SmmTuner(MeasurementInterface):
             return Result(time=float("inf"), accuracy=0.0, size=100.0)
 
     def merge_into_csv(self, filenames):
-        """merge all JSONs into a single CSV file"""
+        """merge all JSONs into a single CSV-file"""
         if self.args.csvfile:
             merged = dict()
             for ifilename in filenames:
                 with open(ifilename, "r") as ifile:
-                    data = json.load(ifile)
                     try:
-                        key = (data["TYPEID"], data["M"], data["N"], data["K"])
+                        data = json.load(ifile)
+                        key = (int(data["TYPEID"]), data["M"], data["N"], data["K"])
                         value = (
                             data["GFLOPS"],
                             data["BS"],
@@ -152,12 +152,10 @@ class SmmTuner(MeasurementInterface):
                             print(
                                 "Worse result "
                                 + ifilename
-                                + " ignored when merging CSV file"
+                                + " ignored when merging CSV-file"
                             )
-                    except KeyError:
-                        print(
-                            "Malformed " + ifilename + " ignored when merging CSV file"
-                        )
+                    except (json.JSONDecodeError, KeyError):
+                        print("Failed to merge " + ifilename + " into CSV-file")
                         pass
             if bool(merged):
                 with open(self.args.csvfile, "w") as ofile:
@@ -214,6 +212,7 @@ class SmmTuner(MeasurementInterface):
             with open(ofilename, "w") as ofile:
                 json.dump(config, ofile)
                 ofile.write("\n")  # append newline at EOF
+                ofile.close()
                 print(
                     "Result achieving "
                     + str(self.gflops)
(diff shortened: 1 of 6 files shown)


src/acc/opencl/smm/tune_multiply.py
```diff
@@ -128,14 +128,14 @@ def run(self, desired_result, input, limit):
...
-        """merge all JSONs into a single CSV file"""
+        """merge all JSONs into a single CSV-file"""
         if self.args.csvfile:
             merged = dict()
             for ifilename in filenames:
                 with open(ifilename, "r") as ifile:
-                    data = json.load(ifile)
                     try:
-                        key = (data["TYPEID"], data["M"], data["N"], data["K"])
+                        data = json.load(ifile)
```
dev-zero, line 137: since you don't need the file anymore after `json.load` I usually do this to reduce indentation levels and make it a bit clearer (Python is not as strictly scoped as C/C++)

```python
with open(ifilename, "r") as ifile:
    data = json.load(ifile)
try:
    key = ...
```

src/acc/opencl/smm/tune_multiply.py
```diff
@@ -152,12 +152,10 @@ def merge_into_csv(self, filenames):
...
                                 + ifilename
-                                + " ignored when merging CSV file"
+                                + " ignored when merging CSV-file"
                             )
-                    except KeyError:
-                        print(
-                            "Malformed " + ifilename + " ignored when merging CSV file"
-                        )
+                    except (json.JSONDecodeError, KeyError):
+                        print("Failed to merge " + ifilename + " into CSV-file")
```
dev-zero, line 158: the `pass` afterwards is therefore not needed anymore (Python does not need an explicit "ignore this exception")

src/acc/opencl/smm/tune_multiply.py
```diff
@@ -214,6 +212,7 @@ def save_final_config(self, configuration):
             with open(ofilename, "w") as ofile:
                 json.dump(config, ofile)
                 ofile.write("\n")  # append newline at EOF
+                ofile.close()
```
dev-zero, line 215: the context manager (which you get by using `with open(...)` will also try to close the file when leaving scope (but is usually able to cope with an already closed file). I'd rather reduce the indent level to trigger it:

```python
with open(ofilename, "w") as ofile:
    json.dump(config, ofile)
    ofile.write("\n")
print(f"Result achieving {self.gflops}")
```

with regard to string concatenation: I would try to avoid it as best as possible (since slow).
The cleanest solution if you can target Python 3.6 and later are f-strings: `f"content: {some_var}"`.
If you have to support Python 3.5 and earlier I'd rather use `"{}".format(some_var)` rather than string concatenation.

src/acc/opencl/smm/tune_multiply.py
hfp: Good to know, thank you!

src/acc/opencl/smm/tune_multiply.py
hfp: > will also try to close the file when leaving scope

That I know, but I wanted to call `merge_into_csv` in underneath of the with-block. However, `merge_into_csv` reads that otherwise still opened file (which can/will fail).

Probably, I should call `merge_into_csv` after the with-block...

src/acc/opencl/smm/tune_multiply.py
hfp: Hmm, let me change the structure a bit more (good point). I moved `data = json.load(ifile)` underneath of `try` to capture read-errors.

hfp:
Note about CI status: `libsmm_acc_unittest_multiply` timed out as part of the Intel test/run.

alazzaro:
> Note about CI status: `libsmm_acc_unittest_multiply` timed out as part of the Intel test/run.

I have to find some time to make that test faster... For the time being, just ignore it...

src/acc/opencl/smm/tune_multiply.py
hfp: I have replaced string concatenation with formatting a string.

hfp:
> > Note about CI status: `libsmm_acc_unittest_multiply` timed out as part of the Intel test/run.

> I have to find some time to make that test faster... For the time being, just ignore it...

Thank you!  As we already know, it is not related to Intel toolchain (GNU run test outdated also this time).

PR #437, opened 2021-03-17
makedep.py: match CP2K's makedep.py

hfp:
* Removed superfluous fhandle.close().
* Open files UTF8-encoded (CP2K).
* Fixed some typos.

### DIFF
--- a/.cp2k/makedep.py
+++ b/.cp2k/makedep.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# -*- coding: utf-8 -*-
 
 import re
 import sys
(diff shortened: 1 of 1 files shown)


hfp:
Daint-CI seems to have a regression of the environment: `CMake 3.17 or higher is required.  You are running version 3.10.2` (see [build-714](https://object.cscs.ch/v1/AUTH_40b5d92b316940098ceb15cf46fb815e/dbcsr-artifacts/logs/build-714/gnu.build.out)). I believe CMake version was sufficient recently and updated to recent needs.

alazzaro:
@hfp You are right, Daint was updated yesterday and now there is a new module to load:

cdt-cuda/20.11

However, I see that the CMake is still old:

CMake/3.14.5(default)

I will inquire with CSCS if we can have new CMake installed...

hfp:
Let's have a separate PR (if necessary) to account for updated Daint (build scripts, etc.).

alazzaro:
Replay by CSCS:

```
You should be able to load the CMake 3.18.4 module with

module load /apps/daint/UES/jenkins/7.0.UP02/gpu/easybuild/modules/all/CMake/3.18.4

```

Issue #416, opened 2021-01-05
cmake: add -fallow-argument-mismatch if MPICH is detected with GCC 10+


dev-zero:
until https://github.com/pmodels/mpich/issues/4300 is solved we need to add `-fallow-argument-mismatch` when building our MPI wrapper. The most portable and clean solution would be to `try_compile` a snippet triggering the issue when gcc 10+ is used and if it is triggered, add the flag. The other option would be to unconditionally add the flag when MPICH and gcc 10+ are used. The latter has the disadvantage that we manually have to update the check once MPICH resolves the issue

PR #438, opened 2021-03-17
updated daint-ci

hfp:
Account for an [updated](https://github.com/cp2k/dbcsr/pull/437#issuecomment-801154121) Daint-CI (CMake, etc), and counter `-Werror` with newer GNU Compiler (certain translation units).

( Unrelated: updated LIBXSMM. )

### DIFF
--- a/.ci/daint.cscs.ch/cray.build.sh
+++ b/.ci/daint.cscs.ch/cray.build.sh
@@ -12,13 +12,9 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-module load daint-gpu cudatoolkit
+module load daint-gpu cudatoolkit cdt-cuda
 module load /apps/daint/UES/jenkins/7.0.UP02/gpu/easybuild/modules/all/CMake/3.18.4
 module unload cray-libsci_acc
-# make sure a recent GCC is available as NVCC backend:
-#   nvcc does not automatically use Cray's CC as backend
-#   and the OS' default gcc-4.8 hangs when building libcusmm_benchmark
-module load gcc
 module list
 
 set -o xtrace  # do not set earlier to avoid noise from module
--- a/.ci/daint.cscs.ch/cray.build.sh
+++ b/.ci/daint.cscs.ch/cray.build.sh
@@ -12,7 +12,8 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-module load daint-gpu cudatoolkit CMake/3.18.4
+module load daint-gpu cudatoolkit
+module load /apps/daint/UES/jenkins/7.0.UP02/gpu/easybuild/modules/all/CMake/3.18.4
 module unload cray-libsci_acc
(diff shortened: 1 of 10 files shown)


alazzaro:
@hfp CSCS introduced a new module for the gpu:

```
module load cdt-cuda
```

alazzaro:
Concerning the `-Wno-error`, we had already a discussion about that and we really want to avoid it.
The problem is introduced in GCC 10.x (however ctd-gpu will load GCC 9.x, since CUDA doesn't support GCC 10 yet).
So, could you revert your commits?
For the GCC 10 rank issue, the problem comes with some MPI (this is the case for the Cray-mpich installed on Daint). @dev-zero proposed [this solution](https://github.com/cp2k/dbcsr/issues/416).

hfp:
> Concerning the `-Wno-error`, we had already a discussion about that and we really want to avoid it.

I am applying `-Wno-error` specifically for the MPI wrapper and one of the examples. Even with GCC 10.x, we preserve `-Werror` for the entire source code base, and still produce warnings for the issue in question (log output). Do you think this solution is acceptable?

> The problem is introduced in GCC 10.x (however ctd-gpu will load GCC 9.x, since CUDA doesn't support GCC 10 yet).
> So, could you revert your commits?

ACK, I will adjust the build scripts. I think we can nevertheless keep the above relaxation about warning-errors for those who want to use GCC 10.x (without CUDA).

> For the GCC 10 rank issue, the problem comes with some MPI (this is the case for the Cray-mpich installed on Daint). @dev-zero proposed [this solution](https://github.com/cp2k/dbcsr/issues/416).

ACK, makes sense. I remember there was also a proposal to wrap things in C. The MPI-F08 module is also supposedly type-safe, but that's another discussion (similar to a C based wrapper).

hfp:
( Coverage dropped by 5% with no source change while Codecov has its own environment. )

dev-zero:
> ( Coverage dropped by 5% with no source change while Codecov has its own environment. )

Codecov is still missing at least one of the tests (hence no post here from the bot and the coverage diff)

alazzaro:
> > Concerning the `-Wno-error`, we had already a discussion about that and we really want to avoid it.
> 
> I am applying `-Wno-error` specifically for the MPI wrapper and one of the examples. Even with GCC 10.x, we preserve `-Werror` for the entire source code base, and still produce warnings for the issue in question (log output). Do you think this solution is acceptable?
>

This is OK, but please extend the check to only GCC >=10.x. The Daint test is to really cover all possible problems, we want to check all warnings... Then, we will make the change more automatic by adopting the @dev-zero solution...

hfp:
> This is OK, but please extend the check to only GCC >=10.x. The Daint test is to really cover all possible problems, we want to check all warnings... Then, we will make the change more automatic by adopting the @dev-zero solution...

ACK. I am just waiting for current Daint-CI to finish.

PR #459, opened 2021-06-03
Remove gpu arch passed to hiprtc

gsitaram:
JIT compilation fails in DBCSR with newer rocm versions (>=4.1)
because there is a mismatch detected between the target id passed to
hiprtcCompileProgram and what hiprtc already passes by default.

hiprtcCompileProgram now honors compile options passed to it. It
also by default has always passed the target ids of the default
device to the compiler, so we don't have to pass this architecture
string as a compile option again. This also helps avoid any
changes to DBCSR when target ids are modified for future AMD GPUs.


hfp:
LGTM. Though, the CMake option to select the architecture for AMD GPUs can probably go as well (if your are confident it is not needed elsewhere). If it goes, it can simplify the build process for AMD GPUs.

gsitaram:
> s

Thanks for merging this PR. I am not sure if we can live without the cmake arch option though.. I'll test and get back.

hfp:
( You are right, I think the arch toggle for CMake is still used to select the correct set of tuned parameters. )

PR #471, opened 2021-07-07
ocl: introduced SHARED_C along with tunable parameter

hfp:
* Issue warning if opencl_libsmm_read_params fails (code path for embedded parameters).
* Introduced ZERO to avoid type-conversion even from literal value (PTX).
* Improved controlling code-paths, and adjusted defaults.
* Fixed allowed range of tuned parameters (AA, and AB).
* Fixed workshare for SHARED_A and SHARED_S.
* Simplified partial workshares.

### DIFF
--- a/src/acc/opencl/smm/kernels/multiply.cl
+++ b/src/acc/opencl/smm/kernels/multiply.cl
@@ -11,8 +11,19 @@
 #else
 # define UNROLL(N)
 #endif
+#if defined(INTEL)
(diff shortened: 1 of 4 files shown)


hfp:
For some reason Daint-CI was not triggered.

hfp:
Also, the squash-merge at least seems to trigger tests again based on GitHub actions (we seem to waste cycle with retesting the same code/diff again).

PR #472, opened 2021-07-09
Fix CP2K Makefile to build DBCSR correctly with HIP support

gsitaram:
When linking CP2K with OpenBLAS that is built with g++ with OpenMP support and enabling DBCSR's HIP backend, if hipcc is used to build DBCSR's cpp files, then in the final CP2K executable we see two OpenMP runtimes linked. One (libgomp.so) that OpenBLAS pulled in, and another (libomp.so) that DBCSR pulled in. To avoid this, it is best to build DBCSR using the same host compiler as CP2K and add a flag to indicate that the HIP_PLATFORM is AMD. This commit and an equivalent commit to CP2K will rely on the setting of USE_ACCEL to determine which compiler and flags to use for compiling DBCSR's cpp files for AMD devices.

I tested on AMD and NVIDIA devices, but not with the OpenCL backend. I changed how the OpenCL source files are included in the build so that they don't get pulled in when building the HIP backend. Please test and let me know if more modifications are necessary.

The only requirement is the setting of USE_ACCEL in CP2K's arch file when building. This should not change how DBCSR is built standalone.

### DIFF
--- a/.cp2k/Makefile
+++ b/.cp2k/Makefile
@@ -30,8 +30,9 @@
 # FCFLAGS  => Fortran compilation flags
 # LDFLAGS  => Linker flags
 # LIBS     => Libraries
-# ACC      => ACC can be nvcc (CUDA), hipcc (HIP), or non-empty string (OpenCL)
+# ACC      => ACC can be nvcc (CUDA), mpicc or g++ (HIP), or non-empty string (OpenCL)
 # ACCFLAGS => ACC flags
+# USE_ACCEL => hip, cuda or non-empty string (OpenCL)
 # GPUVER   =>
 #          - for CUDA, possible values correspond to NVIDIA GPUs:
 #            possible values are K20X, K40, K80, P100, V100
@@ -131,7 +132,7 @@ CXXFLAGS += -D__DBCSR_ACC
 endif
 
 # If compiling with nvcc
-ifneq (,$(findstring nvcc,$(ACC)))
+ifeq (cuda,$(USE_ACCEL))
 override ACCFLAGS += -D__CUDA
 FCFLAGS += -D__CUDA
 CXXFLAGS += -D__CUDA
@@ -143,14 +144,10 @@ ifeq (,$(findstring -Xcompiler,$(ACCFLAGS)))
 override ACCFLAGS += -Xcompiler="$(CXXFLAGS)"
 endif
 # If compiling with hipcc
-else ifneq (,$(findstring hipcc,$(ACC)))
+else ifeq (hip,$(USE_ACCEL))
 override ACCFLAGS += -D__HIP
 FCFLAGS += -D__HIP
 CXXFLAGS += -D__HIP
-#if "--amdgpu-target" has not yet been set in ACCFLAGS
-ifeq (,$(findstring --amdgpu-target,$(ACCFLAGS)))
-override ACCFLAGS += --amdgpu-target=$(ARCH_NUMBER)
-endif
 # OpenCL backend
 else
 # OBJ_SRC_FILES already includes all *.c files (OpenCL backend is C based)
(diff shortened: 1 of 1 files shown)


hfp:
I will pull the patch and try-out/verify in a local environment, and get back to this PR begin of WW29.

gsitaram:
> I will pull the patch and try-out/verify in a local environment, and get back to this PR begin of WW29.

Thank you, Hans. I appreciate your help.

hfp:
I am testing CP2K/master (55d3731eea70b965a4015ad0e31caf946183de86) with DBCSR/develop (87d180b9ed02752968d2994af12b442b30a3dcd0) and code merged from this PR.

alazzaro:
I will take a look this week too, please do not merge...

.cp2k/Makefile
```diff
@@ -183,24 +180,24 @@ LIBSMM_ACC_ABS_DIR := $(shell find $(SRCDIR) -type d -name "libsmm_acc")
...
-    # Exclude autotuning files
-    OBJ_SRC_FILES += $(shell cd $(SRCDIR); find . -type f ! -name "tune_*_exe*_part*.cu" ! -name "tune_*_exe*_main*.cu"  -name "*.cu")
-  # if compiling with hipcc
-  else ifneq (,$(findstring hipcc,$(ACC)))
-    # All *.cpp files belong to the accelerator backend common between CUDA/HIP
-    OBJ_SRC_FILES += $(shell cd $(SRCDIR); find . -type f ! -name "acc_cuda.cpp" ! -name "acc_hip.cpp" -name "*.cpp")
-    OBJ_SRC_FILES += $(LIBSMM_ACC_DIR)/../hip/acc_hip.cpp
-  # OpenCL backend: OBJ_SRC_FILES already includes all *.c files
-  endif
+	ifeq (cuda,$(USE_ACCEL))
```
hfp, line 187: I recommend removing tabs like https://github.com/hfp/dbcsr/commit/193e77555ce94ea6da3940f9602b3aea986484b5. Using tabs outside of rules for indentation purpose can cause problems depending on the version/verbosity of GNU Make.

gsitaram: This is done. Please check.

.cp2k/Makefile
```diff
@@ -183,24 +180,24 @@ LIBSMM_ACC_ABS_DIR := $(shell find $(SRCDIR) -type d -name "libsmm_acc")
 
 ALL_PKG_FILES := $(shell find $(SRCDIR) -name "PACKAGE")
 OBJ_SRC_FILES  = $(shell cd $(SRCDIR); find . ! -name "dbcsr_api_c.F" ! -name "dbcsr_tensor_api_c.F" -name "*.F")
-OBJ_SRC_FILES += $(shell cd $(SRCDIR); find . -name "*.c")
+OBJ_SRC_FILES += $(shell cd $(SRCDIR); find . -path ../src/acc/opencl -prune -type f -name "*.c")
```
hfp, line 183: I have some issues to build with OpenCL now, but do not worry. I will send my fixes on top of yours. I think pruning is not necessary since the related translation units are simply empty if `__OPENCL` is not defined.

hfp: In fact, your introduction of `USE_ACCEL` for the Makefile is sensible and in sync with what's introduced with CMake. I will need to fix-up with OpenCL backend in any case simply to match `USE_ACCEL` with `opencl` rather than falling into the "else-case". Further, OpenCL BE can now pickup own parameter files, which is simply not implemented yet in the Makefile.

.cp2k/Makefile
hfp: Looks good to me, Thank you!

hfp:
> I will take a look this week too, please do not merge...

ACK. When you completed the review and merged this PR, I will go ahead and have some (minor) changes too.

PR #476, opened 2021-07-27
CUDA / HIP files refactoring

alazzaro:
Most of the CUDA and HIP files are in common (switching via macros).


alazzaro:
@hfp I'm refactoring the acc part for CUDA and HIP. The current design (introduced by Shoshana) was to keep (most) files in the CUDA directory, even if they are shared with HIP. To make it clear, I've introduced a cuda_hip directory.

Now, I see you have introduced the following [Makefile](src/acc/cuda/Makefile), but I'm not sure it is related to only CUDA. What it is? I don't think it is for production, so it seems something internal (performance comparison?).

alazzaro:
@hfp I have also merged @gsitaram changes on the makefile for the CP2K compilation. 
Finally, I had the time to review the changes there.
Unfortunately, you went too far with respect to the original plan for this file, which was to keep it in line with the CP2K installation instructions. Currently, DBCSR has OpenCL, which is not documented in CP2K....

My current plan is to move that makefile as part of the CP2K repository (at the end of the day, we don't test it in DBCSR anymore) so that we have a tight bind with CP2K. As a first action, now the ARCH file is required and I've removed the default flags.

I have a question on the OpenCL part:

```
# ACC      => ACC can be nvcc (CUDA), mpicc or g++ (HIP), or non-empty string (OpenCL)
```
what do you mean by "non-empty string"? Something like `ACC=1` to enable OpenCL?

alazzaro:
- [ ] Move CUDA and HIP files in a common directory
- [ ] Simply CP2K makefile
- [ ] Move CP2K makefile to the CP2K repo

hfp:
> @hfp I have also merged @gsitaram changes on the makefile for the CP2K compilation.

Thank you, Alfio!

> Currently, DBCSR has OpenCL, which is not documented in CP2K....

Indeed, it's only documented for DBCSR/develop. Also, it's still being worked on. It should not affect any deployment with CP2K. In fact, I am checking this is true as well as (undocumented) OpenCL backend works.

> My current plan is to move that makefile as part of the CP2K repository (at the end of the day, we don't test it in DBCSR anymore) so that we have a tight bind with CP2K. As a first action, now the ARCH file is required and I've removed the default flags.

Sounds sensible. I will give it a try and do my duty on top of this PR (if necessary).

> I have a question on the OpenCL part:
> 
> ```
> # ACC      => ACC can be nvcc (CUDA), mpicc or g++ (HIP), or non-empty string (OpenCL)
> ```
> 
> what do you mean by "non-empty string"? Something like `ACC=1` to enable OpenCL?

I will just send my changes on top of the HIP state. Gina introduced `USE_ACCEL` similar to what I introduced for CMake (DBCSR). I will bring my work in line with CUDA/HIP. In the end, we have some options to play with before moving into the new era. For the latter, I will definitely look into the "universal kernel" idea as discussed by Ole during the DevMeeting (or as he pointed out earlier in DBCSR's developer discussions).

alazzaro:
Thanks Hans, then I will proceed with my changes and ask you to review them. 
Next, we can make an RC and move to CP2K wide testings...

alazzaro:
> Now, I see you have introduced the following [Makefile](https://github.com/cp2k/dbcsr/tree/cuda_hip_refactor/src/acc/cuda/Makefile), but I'm not sure it is related to only CUDA. What it is? I don't think it is for production, so it seems something internal (performance comparison?).

@hfp what about this Makefile?

hfp:
> @hfp I'm refactoring the acc part for CUDA and HIP. The current design (introduced by Shoshana) was to keep (most) files in the CUDA directory, even if they are shared with HIP. To make it clear, I've introduced a cuda_hip directory.
> 
> Now, I see you have introduced the following [Makefile](https://github.com/cp2k/dbcsr/tree/cuda_hip_refactor/src/acc/cuda/Makefile), but I'm not sure it is related to only CUDA. What it is? I don't think it is for production, so it seems something internal (performance comparison?).

It is used to build the transpose and SMM benchmark drivers without building DBCSR. I am using both OpenCL's and CUDA's Makefile for tests and comparison. If you can leave the CUDA related Makefile with the native CUDA based implementation that would be great. I just have not taken the time (and likely won't) to extend this to HIP. Also, don't worry breaking it, I will take care after your changes.

PR #497, opened 2021-10-20
Print error log when JIT compilation of kernels fail

gsitaram:
It is convenient to see the source being compiled and the errors if JIT compilation fails for some reason. 

@hfp, could you please help check if similar changes are necessary for the OpenCL backend?


hfp:
@gsitaram Thank you for the note! I will have a look. ATM the OpenCL BE has this functionality already (collecting compile log if JIT compilation failed). However, most "root cause of errors" is not performed if `-DNDEBUG` is present. I will check if this can be improved or better aligned with the CUDA/HIP backend.

alazzaro:
I wonder, do we really have such cases? I thought that if we cannot JITing the kernel the code would stop... Apparently it does now, it is running on the CPU fallback...

hfp:
Yes, if it fails, it falls back to CPU code-path. However, the condition about falling back is likely upfront to the JIT-compilation. Running into such an error is still "unexpected" and hence it can be useful to print the compilation log. I think any verbosity (even in case of an error) might be not the best solution for a library especially if it is built around error/return codes.

( The OpenCL BE is quiet even in case of errors unless `-DNDEBUG` is missing. )

PR #499, opened 2021-10-25
Simpler hip backend

gsitaram:
This PR attempts to simplify the DBCSR HIP backend with the following changes (not ready to merge yet, but I am creating the PR to get them tested more thoroughly):

- Remove the use of posterior streams, use only one type of stream. In the current form, we only have priority streams but they are actually low priority, blocking streams. In my tests, higher priority, non-blocking streams seemed unnecessary.
- There is a unique stream assigned per CPU thread for a given process. This removes race conditions seen when running the HIP backend with the original code.
- A device synchronize after the copy of left and right matrices and before the start of the multiply parallel region removes the need for repeatedly checking for event completions in the beginning of every iteration of the loop.
- Copy of stack indices are now done on the same stream as the kernel launches (for kernels that don't use rocBLAS) as they have to be done in order anyway.
- A stream synchronize (by each CPU thread) at the end of the multiply parallel region (i.e., in accdrv_finalize) ensures that the data (C host buffer)  is ready before it is transferred over MPI. This also removes the need for three additional wait on event calls after the kernel completes.

I intend to run the same path with the CUDA backend also and if there is no performance hit, we can unify the code paths. I would prefer not to use #if defined macros before merging if possible.

### DIFF
--- a/src/acc/cuda_hip/acc_dev.cpp
+++ b/src/acc/cuda_hip/acc_dev.cpp
@@ -28,6 +28,11 @@ extern "C" int c_dbcsr_acc_get_ndevices(int *n_devices){
   return 0;
 }
 
+/****************************************************************************/
+extern "C" void c_dbcsr_acc_device_synchronize(){
+  ACC_API_CALL(DeviceSynchronize, ());
+  return;
+}
 
 /****************************************************************************/
 extern "C" int c_dbcsr_acc_set_active_device(int device_id){
--- a/src/acc/cuda_hip/acc_dev.cpp
+++ b/src/acc/cuda_hip/acc_dev.cpp
@@ -29,9 +29,9 @@ extern "C" int c_dbcsr_acc_get_ndevices(int *n_devices){
 }
 
 /****************************************************************************/
-extern "C" void c_dbcsr_acc_device_synchronize(){
+extern "C" int c_dbcsr_acc_device_synchronize(){
   ACC_API_CALL(DeviceSynchronize, ());
-  return;
+  return 0;
 }
 
 /****************************************************************************/
--- a/src/acc/dbcsr_acc_device.F
+++ b/src/acc/dbcsr_acc_device.F
@@ -40,13 +40,17 @@ MODULE dbcsr_acc_device
 
       END FUNCTION acc_set_active_device_cu
 
+      FUNCTION acc_device_synchronize_cu() RESULT(istat) &
+         BIND(C, name="c_dbcsr_acc_device_synchronize")
+         IMPORT
+         INTEGER(KIND=C_INT)                      :: istat
+
+      END FUNCTION acc_device_synchronize_cu
+
       SUBROUTINE acc_clear_errors_cu() &
          BIND(C, name="c_dbcsr_acc_clear_errors")
       END SUBROUTINE acc_clear_errors_cu
 
-      SUBROUTINE acc_device_synchronize_cu() &
-         BIND(C, name="c_dbcsr_acc_device_synchronize")
-      END SUBROUTINE acc_device_synchronize_cu
    END INTERFACE
 
 #endif
@@ -101,8 +105,13 @@ CONTAINS
    END SUBROUTINE dbcsr_acc_clear_errors
 
    SUBROUTINE acc_device_synchronize()
+      !! Fortran-wrapper for waiting for work on all streams to complete
+
 #if defined (__DBCSR_ACC)
-      CALL acc_device_synchronize_cu()
+      INTEGER                                  :: istat
+      istat = acc_device_synchronize_cu()
+      IF (istat /= 0) &
+         DBCSR_ABORT("acc_device_synchronize failed")
 #else
       DBCSR_ABORT("__DBCSR_ACC not compiled in")
 #endif
--- a/src/acc/dbcsr_acc_device.F
+++ b/src/acc/dbcsr_acc_device.F
@@ -15,7 +15,7 @@ MODULE dbcsr_acc_device
 
    IMPLICIT NONE
 
-   PUBLIC :: dbcsr_acc_get_ndevices, dbcsr_acc_set_active_device, dbcsr_acc_clear_errors 
+   PUBLIC :: dbcsr_acc_get_ndevices, dbcsr_acc_set_active_device, dbcsr_acc_clear_errors
    PUBLIC :: acc_device_synchronize
 
    PRIVATE
--- a/src/acc/dbcsr_acc_device.F
+++ b/src/acc/dbcsr_acc_device.F
@@ -15,7 +15,8 @@ MODULE dbcsr_acc_device
 
    IMPLICIT NONE
 
-   PUBLIC :: dbcsr_acc_get_ndevices, dbcsr_acc_set_active_device, dbcsr_acc_clear_errors
+   PUBLIC :: dbcsr_acc_get_ndevices, dbcsr_acc_set_active_device, dbcsr_acc_clear_errors 
+   PUBLIC :: acc_device_synchronize
 
    PRIVATE
 
@@ -42,6 +43,10 @@ MODULE dbcsr_acc_device
       SUBROUTINE acc_clear_errors_cu() &
          BIND(C, name="c_dbcsr_acc_clear_errors")
       END SUBROUTINE acc_clear_errors_cu
+
+      SUBROUTINE acc_device_synchronize_cu() &
+         BIND(C, name="c_dbcsr_acc_device_synchronize")
+      END SUBROUTINE acc_device_synchronize_cu
    END INTERFACE
 
 #endif
@@ -94,5 +99,13 @@ CONTAINS
       DBCSR_ABORT("__DBCSR_ACC not compiled in")
 #endif
    END SUBROUTINE dbcsr_acc_clear_errors
+
+   SUBROUTINE acc_device_synchronize()
+#if defined (__DBCSR_ACC)
+      CALL acc_device_synchronize_cu()
+#else
+      DBCSR_ABORT("__DBCSR_ACC not compiled in")
+#endif
+   END SUBROUTINE acc_device_synchronize
 
 END MODULE dbcsr_acc_device
--- a/src/acc/libsmm_acc/kernels/smm_acc_common.h
+++ b/src/acc/libsmm_acc/kernels/smm_acc_common.h
@@ -13,7 +13,7 @@
 #define __HIP
 #endif
 #endif
-#if defined(__HIP) && !defined(__HIP_PLATFORM_NVCC__)
+#if defined(__HIP) && !defined(__HIP_PLATFORM_NVCC__) && !defined(__HIPCC_RTC__)
 # include <hip/hip_runtime.h>
 #endif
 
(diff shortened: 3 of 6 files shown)


gsitaram:
@hfp Hi Hans, 
In this PR, I simplified the HIP and CUDA backend code for DBCSR in order to solve a race condition that was only manifesting when the HIP backend was used. In particular, submitting work to the same stream from multiple CPU threads seemed to be the issue. I don't see incorrect results when I separated work from different CPU threads into different streams.  
The question for you is the following. Does the OpenCL backend also use this code? If not, my changes, that assume that this is used only for HIP and CUDA, are sufficient.

gsitaram:
@hfp, thank you for your feedback. If I can squeeze some time, I would like to test your OpenCL backend on AMD GPUs. Will reach out if I need assistance.

Your comment also made me think of something. For now, to keep it simple, I have removed the priority assignment to the streams created.  Although the variable name is "priority_streams", they are just blocking no-priority streams. If someone thinks that the variable must be renamed, please let me know. I can do it.

hfp:
> Although the variable name is "priority_streams"

From my point of view that's just fine. It can remind us of the past implementation (given your code is merged) and keeps the knowledge of what was a "priority stream".

> they are just blocking no-priority streams.

I guess you still mean in-order streams with work enqueued and executed in order?

Previously, the priority was perhaps meant to prefer some streams vs others like hinting the runtime to prefer dispatching work from higher priority streams. I think streams are not blocking wrt host execution (I hope). However, `c_dbcsr_acc_device_synchronize` blocks on the host much like `c_dbcsr_acc_stream_sync` except the former does it on a per-device basis, i.e., for all streams pointing to the device that is currently active (which is potentially thread-specific). Your nice PR now "synchronizes" work only on a per-thread basis which means no/less internal locking needed (and potentially no data race).

hfp:
@gsitaram Here is a short guide on how to tune a kernel (for any GPU using the OpenCL backend): https://cp2k.github.io/dbcsr/develop/page/2-user-guide/4-gpu/index.html. This works in a stand-alone fashion (no need to build entire DBCSR, etc.). The driver program (`src/acc/acc_bench_smm`) used during auto-tuning not only works with the OpenCL backend but also with the CUDA/HIP backend (it solely depends on DBCSR's ACC-interface). Having said the latter you can for instance compare directly the CUDA/HIP and OpenCL based implementation at least using the aforementioned driver (same/artificial input, driver code).

gsitaram:
> > Although the variable name is "priority_streams"
> 
> From my point of view that's just fine. It can remind us of the past implementation (given your code is merged) and keeps the knowledge of what was a "priority stream".
> 
> > they are just blocking no-priority streams.
> 
> I guess you still mean in-order streams with work enqueued and executed in order?
Yes, the stream semantics remain as is. In the case of priority streams, in DBCSR, those were created with a higher priority and using the "hipStreamNonBlocking" flag which would make them non-blocking with respect to the NULL stream. In my PR, I did not use either the priority or this flag. So, these are all just regular streams where work enqueued to a stream is executed in order without blocking other streams or the host. If any work is enqueued to the NULL stream, then it will synchronize work in all these other streams.. but we are taking care to avoid doing that.
> 
> Previously, the priority was perhaps meant to prefer some streams vs others like hinting the runtime to prefer dispatching work from higher priority streams. I think streams are not blocking wrt host execution (I hope). However, `c_dbcsr_acc_device_synchronize` blocks on the host much like `c_dbcsr_acc_stream_sync` except the former does it on a per-device basis, i.e., for all streams pointing to the device that is currently active (which is potentially thread-specific). Your nice PR now "synchronizes" work only on a per-thread basis which means no/less internal locking needed (and potentially no data race).
Yes exactly. I find that if a CPU thread enqueues all of its work on it's own stream, then we don't seem to have the race condition (haven't found incorrect results yet.. which means this code needs more testing :) ). So, at the end of the multrec_multiply loop, each CPU thread only has to synchronize it's own stream. 
The device synchronize before the loop begins is to avoid the need for repeated hipStreamWaitEvent() calls to ensure A, B and C areas are ready.  Every API call adds overhead, so I wanted to try and keep it simple.
Again, we need this code to be tested thoroughly for correctness and for any performance regressions on AMD or Nvidia GPUs.

@hfp, thanks for the info regarding autotuning. I will take a look.

hfp:
> Again, we need this code to be tested thoroughly for correctness and for any performance regressions on AMD or Nvidia GPUs.

Just a "heads up" - the upcoming release of CP2K (v9.1) is targeted for the end of this year. If we assume this PR goes into DBCSR 2.2 and in turn goes into CP2K 9.1, it would be good to get it into CP2K/master/main asap. CP2K runs a broader set of GPU tests as well as performance tests (I think Daint-CI is down ATM or not running for DBCSR).

I will pull this PR by begin of next week (if it's not merged then), and test some workloads like "before" and "after". @alazzaro what about a new RC by end of next week the latest?

alazzaro:
> 
> I will pull this PR by begin of next week (if it's not merged then), and test some workloads like "before" and "after". @alazzaro what about a new RC by end of next week the latest?

Yes, I agree, this is a good target.

hfp:
To simplify the process of performance comparison, I suggest using `tests/dbcsr_perf` with inputs drawn from [tests/inputs](https://github.com/cp2k/dbcsr/tree/develop/tests/inputs). Specifically, I am using `tests/inputs/test_H2O.perf` with an increased matrix size (2208->11500) but less repetitions (50->3).

So far (only preliminary), I used a legacy system with two CPUs and only one P100 and this configuration yields ~25% less performance using this PR when compared to the current master/settled synchronization mechanism. I will further look at systems with two+two CPUs and GPUs including V100 and A100.

oschuett:
While you are at it, maybe you could also run some cp2k bechmarks, e.g.  [H2O-64.inp](https://github.com/cp2k/cp2k/blob/master/benchmarks/QS/H2O-64.inp)?

I've made another attempt to investigate the slowdown we're seeing since https://github.com/cp2k/cp2k/pull/1640. However, it's really difficult to get reproducible results. It seems that after a few minutes of load the nodes become unstable. So, it might be that https://github.com/cp2k/cp2k/pull/1640 only triggered a dormant issue by putting more load on the GPUs.

hfp:
> While you are at it, maybe you could also run some cp2k bechmarks, e.g. [H2O-64.inp](https://github.com/cp2k/cp2k/blob/master/benchmarks/QS/H2O-64.inp)?
> 
> I've made another attempt to investigate the slowdown we're seeing since [cp2k/cp2k#1640](https://github.com/cp2k/cp2k/pull/1640). However, it's really difficult to get reproducible results. It seems that after a few minutes of load the nodes become unstable. So, it might be that [cp2k/cp2k#1640](https://github.com/cp2k/cp2k/pull/1640) only triggered a dormant issue by putting more load on the GPUs.

In fact, I ran H2O-256 already but switched to DBCSR's own benchmark since the results achieved with the former had some occasional fluctuation (which I just wanted to avoid for this evaluation). I though going for `dbcsr_perf` is more pure and maybe allows a more clear conclusion. I will take some time over the course of this week to get some more conclusive finding also with CP2K.

oschuett:
It occurred to me that we might be suffering from bad GPU assignments. I've opened https://github.com/cp2k/cp2k/issues/1743 to track this.

Unfortunately, I can't investigate further because the GPU Cloud is currently not available.

hfp:
> It occurred to me that we might be suffering from bad GPU assignments. I've opened [cp2k/cp2k#1743](https://github.com/cp2k/cp2k/issues/1743) to track this.
> 
> Unfortunately, I can't investigate further because the GPU Cloud is currently not available.

That was my suspicion too. I am about to wrap up some additional verbose output for the OpenCL backend showing the active thread-local devices for each PID. I think on per-process basis (intra-process), CP2K/DBCSR can set active device in a round-robin fashion for every thread which should be good enough given that normally at least one rank per socket/device is used. However, controlling visible devices on a per-process basis is typically out-of-scope for an application like MPS-controlled in case of NVidia and some MPI-builtin (device-)affinity in case of Intel (and perhaps others).

hfp:
@gsitaram My recommendation for now (and to include this work into the next RC) is to toggle between device-based synchronization (your PR) and the current solution. This way the evaluation can continue. The actual setting can even reuse one of the established preprocessor settings/variables (like instead of being just defined or `1`, it can be defined to be `2` or whatever). Anyway, it may need some more work for DBCSR/CMake, and CP2K/exts/build_dbcsr/Makefile.

My results finally rely on H2O-64 on a dual-socket system with one V100 per socket. The workload was scaled like 4, 6, 8, and 16 ranks just for the sake of showing some overheads.

|   | 4 | 6 | 8 | 16 |
|---|---|---|---|---|
| cursync | 78s | 71s | 69s | 58s |
| devsync | 80s | 73s | 69s | 66s |

The timings shown above are walltime (end-to-end). For the `dbcsr_multiply_generic` entry (profile), it looks like:

|   | 4 | 6 | 8 | 16 |
|---|---|---|---|---|
| cursync | 28s | 37s | 35s | 31s |
| devsync | 29s | 39s | 36s | 40s |

If you have some different conclusion and/or measurements, this would be interesting to know.

gsitaram:
@hfp, I ran other tests (H2O-LS-DFT NREP2, NREP4) on A100s, and I see ~1% lower performance on average with the code in my PR. I ran on 1, 4 and 8 GPUs. My tests involved running multiple ranks on each GPU but only within a single node. I used 16 ranks in some tests. It seems like your tests ran on multiple nodes, and `dbcsr_multiply_generic` encompasses MPI + CPU + GPU activity. Should we be looking at `multiply_cannon_multrec` instead?

One thing to note is that we have to use this code path for AMD GPUs. If you like, I could take the code back to one of my [previous commits](https://github.com/cp2k/dbcsr/pull/499/commits/d7776aff1be93e99052ae409930e3c5156c9f9d5) that leaves the CUDA path untouched and just changes the path for HIP. Given this, do we need another compiler flag? I am not super confident about the CMake/Makefile side of things too, so would take a while I guess. Also, I would be unavailable next week, so I wonder if this is something that needs to be done tomorrow. 

If we decide on a new compiler flag, what would you recommend?

hfp:
@alazzaro What do you prefer? Straight merge or conditional compilation? The latter requires CMake build option, etc.

I believe practical performance is on par (only a very minor regression). This PR certainly simplifies the synchronization approach and is supposed to resolve some races when relying on the HIP backend. We still have [problems](https://github.com/cp2k/cp2k/issues/1743#issuecomment-972691536) with multiple GPUs and a simplified approach is the right step.

@gsitaram My [experiment](https://github.com/cp2k/dbcsr/pull/499#issuecomment-972802201) was only single-node.

hfp:
I will merge this PR and @alazzaro can create a new RC. I rather prefer testing a new RC under the hood of CP2K.

hfp:
There will be a follow-on PR:  

* Get OpenCL BE working again. The apparently the complementary code-path of `#if defined(__DBCSR_ACC) && (defined(__HIP_PLATFORM_AMD__) || defined(__CUDA) || defined(__HIP_PLATFORM_NVCC__))` seems to be an invalid configuration state.
* Cleanup DBCSR's configuration from "posterior" and "priority" terminology while compatibility with CP2K (calling `dbcsr_set_config` and `dbcsr_get_config`) is preserved.

PR #531, opened 2021-12-20
baseline versions instead of pinned Python requirements

hfp:
Set baseline versions instead of pinning requirements. Note, there is no trace in Git's log (even prior to directory rename/restructure) telling about reasons for pinning requirements.


dev-zero:
@hfp there's no log because it's quiet common to pin versions in a `requirements.txt`: the `requirements.txt` often specifies a set of dependencies which are supposed to surely work (e.g. a set which have been tested) while a `setup.py` then contains wider version ranges. This `requirements.txt` was likely generated with `pip freeze`, hence also recording more than just the direct dependencies.

The latest blog entry about versioning I found via HN was this one: https://hynek.me/articles/semver-will-not-save-you/
The take-away msg for me was: unmaintained software will break, no matter what.
With this change users will more likely not install vulnerable packages, but we might ship something which does not run anymore (or worse: produce different results) without noticing.

hfp:
>  unmaintained software will break, no matter what.

@dev-zero fully ACK, and my PR/patch was "practically breaking it" since I cannot verify this complex facility. However, we have to mute the security issue in order to stay trusted/responsible and if this code matters, somebody will report an issue (and we can fix it). ATM, nobody ever submitted parameters or ever ran predictive modelling (as it takes nearly as long as the tuning for optimal parameters).

hfp:
> I cannot verify this complex facility [...]

Just realized, I can verify it (but it is a tradeoff where time is spent).

PR #576, opened 2022-03-01
cuda/hip: improved initialization of libsmm_acc

hfp:
* Protect against multiple initialization.
* Protect against change of thread-count.


hfp:
There is an issue about reaching Codecov service hence some failing tests.

PR #590, opened 2022-03-22
Complement enabling ACC timers for non-ACC compilation

hfp:


alazzaro:
Thanks!

hfp:
> Thanks!

No problem (hopefully this fixes https://buildkite.com/intel/dbcsr/builds/804).

alazzaro:
But why the CI doesn't face this error? Is it something we don't build via cmake?

hfp:
> But why the CI doesn't face this error? Is it something we don't build via cmake?

I guess both `acc_bench_smm` and `acc_bench_trans` are not build and exercised (completely stand-alone). Regarding `dbcsr_acc_test`, I think dbcsr library is linked when using CMake based build. So, the issue solved here is mostly related to my Makefile based backend specific builds/tests (OpenCL and CUDA). However, this change introduced a dependency criticized earlier. A setter for timer function-pointers is perhaps preferable and avoiding the backends to depend on DBCSR/frontend.

hfp:
I will eventually send another PR contributing `acc_bench_smm` and `acc_bench_trans` to be build with CMake.

PR #592, opened 2022-03-23
Build acc_bench drivers as part of the tests

hfp:
* Fixed warning about unused variable.

### DIFF
--- a/src/acc/acc_bench_smm.c
+++ b/src/acc/acc_bench_smm.c
@@ -233,14 +233,14 @@ int main(int argc, char* argv[]) {
     ELEM_TYPE *amat_hst = NULL, *bmat_hst = NULL, *cmat_hst = NULL;
     ELEM_TYPE *amat_dev = NULL, *bmat_dev = NULL, *cmat_dev = NULL;
     void* stream = NULL;
-#if defined(__OPENCL)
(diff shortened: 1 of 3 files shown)


hfp:
The additional targets still link against dbcsr library. So, an issue like https://github.com/cp2k/dbcsr/pull/590#issuecomment-1075597791 would be still not reproduced. For the latter, more work is needed like building a dbcsr_acc library (like pre-CMake time).

hfp:
Final push will remove https://github.com/cp2k/dbcsr/commit/a2bedce43c925591ac1de64da9e044897ae8b61d#diff-38405a63679b1bd07fa830ce45808a408bc7f0c367d57ce2a814061fe01e6e03R57.

alazzaro:
No sure I follow the discussion here...  The two acc_bench drivers are built as standalone package via makefile, i.e. they don't depend on the full DBCSR built. I like the idea to have them as part of the tests. But why we want to keep them standalone? Note that in the long run cmake will be the only way to build DBCSR...

alazzaro:
On the other hand, would it be possible to introduce an OpenCL build on the CI (just like we do for HIP)?

alazzaro:
Let's wait on this PR, I have to understand better...

Issue #427, opened 2021-02-04
tests: make libsmm_acc tests run with MPI


dev-zero:
currently the tests output unconditionally and are not being run through the MPI wrappers (`mpirun` or `srun`) in `tests/CMakeLists.txt`

alazzaro:
Comment those tests for the time being (suggested by @hfp !)
See https://github.com/cp2k/dbcsr/pull/595/commits/dec083a1d1432c8ca946500d56b8f21383a1eda4

Issue #422, opened 2021-01-21
ACC init/finalize issues


hfp:
There are two potential issues with DBCSR's init/fini flow wrt ACC interface and LIBSMM:

1. DBCSR calls `acc_init()`, which eventually calls `libsmm_acc_init()` internally (this is a potential issue since the backend (CUDA/HIP, or OpenCL, etc.) then depends on LIBSMM). The latter routine (`libsmm_acc_init`) was only recently introduced (everything here applies similarly to the respective finalize routine also). However, the expected flow (IMHO) is that DBCSR calls `acc_init()` **and** `libsmm_acc_init()` since both respective interfaces are used (and DBCSR does not judge potential internal depedency between ACC and LIBSMM implementation). Then, `libsmm_acc_init()` may call `acc_init()` internally if the ACC interface is used to implement LIBSMM (note, this dependency is somewhat expected). In turn, all init/finalize routines should be safe against multiple/over-initialization.
2. DBCSR calls `dbcsr_multiply_lib_init()` inside of a parallel region, and subsequent (sub-)initialization may be called only by the master thread. For instance, `acc_init()` is called by the master thread only (OMP MASTER construct). However, `acc_finalize()` is apparently called by all threads of a parallel region (workshare). The latter is inconsistent with respect to a "tandem flow" of  init/finalize. The expect behavior (IMHO) is to protect acc_init/finalize and libsmm_acc_init/finalize against multiple threads, i.e., ACC and LIBSMM implementations are only expected to be thread-safe after initialization and before finalization.

dev-zero:
ACK. Possibly related to #421?

hfp:
> 
> 
> ACK. Possibly related to #421?

Could be, at least related in the sense of being an OpenMP/multicore inconsistency.

For OpenCL backend, the first issue from above (1) was unveiled by linker errors and weird dependency/use of libraries/objects in CMake, for the second issue (2) the expectation is/was made explicit (https://github.com/hfp/dbcsr/blob/openclacc/src/acc/opencl/acc_opencl.c#L133).

haampie:
@hfp I don't know if you've seen this, but I prefixed the full C api with dbcsr_c_* here: https://github.com/cp2k/dbcsr/pull/419 to avoid these clashes.

hfp:
@haampie Thank you! I was not aware about this. Also, good to know if end up merging our work together.

[...]

PR #595, opened 2022-03-25
ACC Setdevice reordering

### LINKED ISSUE #591: ACC: orphaned context causing multi-gpu issue
Investigate if creating a context for the currently active device ("init" with device "0") and later switching the device ("set_active_device") is causing orphaned resources like an orphaned context and thereby multi-GPU issues such as imbalance.

alazzaro:
- Setdevice is now called as part of acc_init. This problem was introduced with https://github.com/cp2k/dbcsr/commit/eb1c2688730ec5d1260c0a0cb37f71c1a4c362b6

- It is forbidden to change the device during the DBCSR execution by design, so acc_init is the correct place where to call setdevice

- This PR should fix #591 

- We still have some of the problems discussed in #422 , we use another PR for that...

- Bonus: I'm commenting acc tests on because of #427

### DIFF
--- a/src/acc/PACKAGE
+++ b/src/acc/PACKAGE
@@ -1,5 +1,5 @@
 {
 "description": "Generic accelerator API",
 "archive": "libdbcsr",
-"requires": ["../base", "cuda", "hip", "opencl", "libsmm_acc"]
+"requires": ["../base", "../core", "cuda", "hip", "opencl", "libsmm_acc"]
 }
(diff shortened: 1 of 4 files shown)


hfp:
This looks good to me. I like the Bonus too. Instead of reviewing the code line-by-line, I will test it on a multi-GPU system and see if #591 is resolved.

hfp:
I am working on `makedep error: Dependency forbidden according to package manifest: exts/dbcsr/src/acc/dbcsr_acc_init.F -> exts/dbcsr/src/core/dbcsr_config.F` which appears when integrated into CP2K.

hfp (approved):

hfp:
I can confirm, this PR resolves #591.

PR #601, opened 2022-04-04
Fix libsmm_acc_finalize

oschuett:


alazzaro:
@oschuett Thanks for this PR. I have overlooked this change, which is definitely not right. The finalize has to release all DBCSR components.
Looking in the history, the change was introduced by @hfp with https://github.com/cp2k/dbcsr/commit/341630079f987aea33d4458041bd6a9f5a64e77d. 
@hfp this change introduced a problem with the regression test when the keepalive option is used (see https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-03bbf95e_report.txt). Can we revert the change?

BTW, we can simply the init too...

hfp:
Originally, the fini code was not safe against double-finalization (and similarly for init). I believe it is good to have this protection but the other solution is to avoid double-init/fini in the first place.

oschuett:
The important part is really the `acc_blashandles.clear();`. The rest is just followup cleaning.

alazzaro:
> Originally, the fini code was not safe against double-finalization (and similarly for init). I believe it is good to have this protection but the other solution is to avoid double-init/fini in the first place.

Do you mean via the API? We check for that... (cannot initialize/finalize twice). 
I think you mean the internal functions, i.e. acc_init and acc_finalize. Do we need to check for that? (just like a singleton...) There are several nested calls... On the other hand, if the problem just the cublas init, we can definitely check if the are not already handles and skip otherwise...

alazzaro:
> The important part is really the `acc_blashandles.clear();`. The rest is just followup cleaning.

The clear is what I like (finalize has to clean everything). Indeed, if you call the function twice, the array will be empty and everything will work... And, as I said, we can clean the init too.

oschuett:
> And, as I said, we can clean the init too.

If you'd rather solve this differently or more generally then feel free to close this PR instead of merging it.

hfp:
I will merge this PR. Anything beyond this can be covered by a 2nd PR (follow-up).

PR #606, opened 2022-04-14
Rely on C_SIZE_T when calculating size in Bytes

hfp:


hfp:
This PR is sent on top of f4f25649551a206610f81b0ba17635f3aba78483 since current master fails CI-validation. However, this PR ran into https://github.com/cp2k/dbcsr/issues/607.

PR #610, opened 2022-04-21
gha: fix pre-commit action

dev-zero:


alazzaro:
Unrelated, but this documentation generation is still crashing https://github.com/cp2k/dbcsr/runs/6114486790?check_suite_focus=true ...

Issue #553, opened 2022-02-09
Add CP2K support for A100


shahzebsiddiqui:
I am trying to build CP2K for our NVIDIA A100 system using spack, I reported an issue and according to @dev-zero in https://github.com/spack/spack/issues/28554#issuecomment-1033840686 he suggested we post an issue here so support can be added.

alazzaro:
Autotuning the kernels for A100 is definitely on our schedule. In the meantine, you can use V100 kernels.

dev-zero:
done in #629

PR #629, opened 2022-09-06
Feature/A100 support

dev-zero:


dev-zero:
this resolves  #553

hfp:
I suggest to carry-forward V100 parameters for kernels not present in the A100 set.

PR #633, opened 2022-10-19
Mtaillefumier patch 1

mtaillefumier:
it is a simple fix for a problem I encountered while working on the cp2k cmake build system. DBCSR cmake config file does not search for libxsmmext when libxsmm support is included. It results in a build failure for any cmake project that uses dbscr as a component.


hfp (approved):

hfp:
This issue is a correct observation, it was introduced with LIBXSMM/main branch revising the internal dependencies like `libxsmmf` now depends on `libxsmmext`. I may be able to revisit this dependency by adjusting the Fortran interface somehow, but in any case this change does not harm.

PR #634, opened 2022-11-10
ocl: minor update and prettify

hfp:
* Print compiler version when performing static analysis.
* Improved handling failures.
* Updated LIBXSMM (Daint-CI).
* Prettify.


hfp:
The precommit test failing seems to be a discrepancy/disagreement between versions of `clang-format`.

PR #635, opened 2022-11-20
Workarounds for nvfortran compiler

oschuett:


hfp:
Thank you for the fix!  Precommit is currently broken and unrelated to your PR.

alazzaro:
Out of curiosity, are we able to run CP2K with the nvidia compiler? (former PGI compiler) In the past there were several issues...

hfp:
I have not checked. I can only confirm: the PGI/Nvidia Compiler is part of LIBXSMM's CI and passes our standard set of "quick-tests".

oschuett:
Yes, I've manged to build CP2K with nvfortan:
```
  cd cp2k/exts/dbcsr
  git fetch origin
  git checkout origin/develop

  cd ../../tools/docker/
  docker build -f ./Dockerfile.test_nvhpc ../../
```
Unfortunately, many of the regtests are failing. The bigger issues, though, is that the build itself is very slow. So I think we first have to speed it up before we can start debugging.

alazzaro:
In the past (i.e. 3 years ago) even `-O0` was making tests crashing... In particular, there was a serious issue with INTENT policy in the compiler. Good that you are resuming it, but maybe we should get more help by nvidia...

Talking about compilers, I would like to have NAG because it has the best error checking (see https://www.fortranplus.co.uk/app/download/32186747/fortran_compiler_diagnostic_capability_revision_5.pdf )

oschuett:
> but maybe we should get more help by nvidia...

I was actually contacted by an engineer from the NVIDIA compiler team. They are working on it :-)

> I would like to have NAG...

Yeah, me too. Unfortunately, their website does not even list a price, which usually means it's super expensive. I guess we could try to get a sponsorship, but the easier route is probably to run the tests at some HPC center that already owns a license.

PR #637, opened 2022-12-13
Drop support for the MPI 2 standard

fstein93:
This PR removes the __MPI_VERSION flag to turn off code requiring the MPI 3 standard. Because MPI3 was published 10 years ago, it will be fully supported by all MPI implementations in use.
It is the equivalent to the PR cp2k/cp2k#2438 .


alazzaro:
retest this please

hfp (commented):
Re:OpenCL - Thank you!

PR #654, opened 2023-02-05
Use memory pool also for Cannon send buffers

oschuett:
Some MPI implementations have high overhead when encountering a new buffer.

For the H2O-DFT-LS benchmark this change yields up to 8% speedup with OpenMPI.


oschuett:
@alazzaro, what do you think? I don't want to merge without your approval.

hfp:
I guess this patch has no impact for MPICH/2 like Plain, Cray, or Intel MPI? I think confirming for just one of the MPICH implementations is sufficient.

alazzaro:
yes, it can be merged. I've tested it. For it appears strange to me that we always overlooked it... thanks @oschuett for point it!

PR #656, opened 2023-03-07
Merged tuned parameters from V100 into A100-parameters

hfp:
This complements https://github.com/cp2k/cp2k/pull/2655.


alazzaro:
No, this is not actually what the autotuning was supposed to be.
Merging V100 in A100 means that people will not run autotuning on their side, which is the entire idea behind all the procedure and well documented.
I would propose to follow another solution, i.e. provide a generic kernel for all not existing ones and then make a warning "kernel not found, please autotune it".

hfp:
Note: as far as CP2K goes (https://github.com/cp2k/cp2k/pull/2655), A100 systems used V100 parameters exclusively.

PR #658, opened 2023-04-04
ocl: match devices against tuned parameters based on device-id

hfp:
* Print device-uid as part of the device-name (ACC_OPENCL_VERBOSE).
* Use LIBXSMM_MIN/MAX (instead of MIN/MAX), and rely on LIBXSMM_EXPECT.
* Adjusted default (OPENCL_LIBSMM_SMM_AC, OPENCL_LIBSMM_SMM_LU).
* Updated device-name with device-id (P100/V100 parameters).
* Remain compatible with LIBXMSM 1.17
* Updated LIBXSMM (Daint-CI).


hfp:
Cuda [build](https://github.com/cp2k/dbcsr/actions/runs/4611310100/jobs/8150849186?pr=658) is broken (unrelated to this PR).

PR #660, opened 2023-04-06
ocl: flexible matching of tuned parameters

hfp:
* Match parameters w/ libxsmm_strimatch (if no exact match)
* Introduced --csv-nogflops (tune_multiply.py).
* Added/donated tuned parameters (A100, PVC).
* Revised format of some verbose output.
* Updated LIBXSMM (Daint-CI).


hfp:
Added tuned parameters as [promised](https://github.com/cp2k/dbcsr/pull/656#issuecomment-1458195654).

PR #671, opened 2023-05-12
Add the posibility to build dbcsr shared library

mtaillefumier:
trivial change needed when the shared version of libcp2k is needed. OFF by default

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,7 @@ set(dbcsr_APIVERSION ${VERSION_MAJOR}.${VERSION_MINOR})
 # OPTIONS
 include(CMakeDependentOption)
 
+option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
 option(USE_OPENMP "Build with OpenMP support" ON)
 option(USE_MPI "Build with MPI support" ON)
 cmake_dependent_option(
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -269,9 +269,11 @@ if (WITH_C_API)
   add_library(dbcsr_c ${DBCSR_C_SRCS})
   set_target_properties(dbcsr_c PROPERTIES LINKER_LANGUAGE Fortran)
 
-  set_target_properties(dbcsr_c PROPERTIES VERSION ${dbcsr_VERSION}
-                                           SOVERSION ${dbcsr_APIVERSION}
-                                           POSITION_INDEPENDENT_CODE ON)
+  set_target_properties(
+    dbcsr_c
+    PROPERTIES VERSION ${dbcsr_VERSION}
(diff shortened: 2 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -157,7 +157,8 @@ endif ()
 add_library(dbcsr ${DBCSR_SRCS})
 
 set_target_properties(dbcsr PROPERTIES VERSION ${dbcsr_VERSION}
-                                       SOVERSION ${dbcsr_APIVERSION})
+                                       SOVERSION ${dbcsr_APIVERSION}
+			               POSITION_INDEPENDENT_CODE ${BUILD_SHARED_LIBS})
```
dev-zero, line 161: This is only required if `dbcsr` were a `STATIC` library, which it isn't.

CMakeLists.txt
```diff
@@ -69,6 +69,7 @@ set(dbcsr_APIVERSION ${VERSION_MAJOR}.${VERSION_MINOR})
 # OPTIONS
 include(CMakeDependentOption)
 
+option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
```
dev-zero, line 72: ACK

dev-zero (changes requested):

src/CMakeLists.txt
mtaillefumier: It does not matter actually. Static libraries can also be build with -fPIC (usually even suggested when we mix static and dynamic libraries). I changed it `POSITION_INDEPENDENT_CODE ON` all the time.

hfp: > It does not matter actually. Static libraries can also be build with -fPIC (usually even suggested when we mix static and dynamic libraries). I changed it `POSITION_INDEPENDENT_CODE ON` all the time.

Indeed, very advisable to use Position Independent Code even for all objects even when collected into an archive aka static library. Nowadays, the reason things work without `-fPIC` is that some toolchains apply it by default, e.g., one-step-builds (compile and link into one command). Using `-fPIC` everywhere also technically allows to use the same objects twice (for archive and shared library).

hfp:
@dev-zero Are all requests for change addressed?  If so, let's merge this contribution.

dev-zero:
https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html

So, setting it explicitly does not make sense.

mtaillefumier:
it does as we often link static libraries and dynamic libraries together. Doing so results in potential errors during linking time which is due to the absence of -fPIC during compilation. We can leave it on all the time as the address resolution will happen at linking time if we use archive libraries instead of shared libraries. it has absolutely no influence on performance.

dev-zero (approved):
Ok, I would really prefer people would use one or the other, rather than mix and match, but that's wishful thinking.

PR #677, opened 2023-06-16
Wrap the MPI handles

fstein93:
This PR wraps the MPI handles which used to be plain integers into their own derived types. The handle component is kept private to facilitate the transition to the new mpi_f08 bindings introducing their own types and allow to switch between serial/MPI 90 binding/MPI 2008 bindings. I implemented setters and getters and the comparison operators for convenience. Regarding the constants, the wrapper module will distinguish the handles (which are used for the default initializers) and the actual parameters exposed to the rest of the library. This refactoring improves the type safety within the library.

The new types are used only internally and are not meant to be exposed to users of DBCSR which is why some routines accessible via the dbcsr_api.F module need a wrapper routine to wrap the passed integer handles into the new types. Thus, the external APIs did not change.

Similar changes have already been applied to CP2K. I will continue this PR by adding support for the MPI 2008 bindings which should solve some issues with the old MPI 90 bindings.

### DIFF
--- a/src/core/dbcsr_lib.F
+++ b/src/core/dbcsr_lib.F
@@ -27,7 +27,7 @@ MODULE dbcsr_lib
                             mp_environ, mp_cart_rank, &
                             rm_mp_perf_env, &
                             mp_comm_free, &
-                            mp_get_comm_count
+                            mp_get_comm_count, mp_comm_type, mp_comm_null
    USE dbcsr_mm, ONLY: dbcsr_multiply_clear_mempools, &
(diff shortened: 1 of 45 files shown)


hfp:
@alazzaro Any objection?

( I think this PR is meant to be the starting point for more contributions. )

alazzaro:
I'm sorry, I cannot really follow on that this week... Please go ahead and merge it.

fstein93:
@hfp @alazzaro Gentle ping

hfp:
Thank you, Frederick four your contribution!

PR #678, opened 2023-06-26
Add support for the new mpi_f08 module

fstein93:
The MPI 3.0 standard introduces the mpi_f08 module. This solves most issues arising with the old Fortran module (strong typing, non-blocking communication, allocation of memory, etc.). This PR enables this feature within DBCSR.

Because of experiences with its implementation in CP2K, its use must be explicitly requested by the user by adding the flag `-DUSE_MPI_F08=<ON|OFF>` while calling CMake. It is known that this feature requires more recent compilers (GCC version>10 or Intel compiler) and suitable MPI implementations (OpenMPI and IntelMPI seem to work well, MPICH has issues with Gfortran 11/12). Some of the admins should add or adjust a suitable test for this feature, as this is apparently not possible for me.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,7 @@ include(CMakeDependentOption)
 option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
 option(USE_OPENMP "Build with OpenMP support" ON)
 option(USE_MPI "Build with MPI support" ON)
+option(USE_MPI_F08 "Build with the mpi_f08 module support" OFF)
 cmake_dependent_option(
   WITH_C_API "Build the C API (ISO_C_BINDINGS)" ON "USE_MPI" OFF
 )# the ISO_C_BINDINGS require MPI unconditionally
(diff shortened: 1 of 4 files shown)


hfp:
@dev-zero @alazzaro Not sure if I can add an additional test for `-DUSE_MPI_F08=ON`. As noted above, the test may apply to OpenMPI and/or IntelMPI (not plain/stock MPICH). I guess, GCC/GFortran wise test requirements are fulfilled.

CMakeLists.txt
```diff
@@ -197,6 +198,15 @@ The listed MPI implementation does not provide the required mpi.mod interface. \
 When using the GNU compiler in combination with Intel MPI, please use the \
 Intel MPI compiler wrappers. Check the INSTALL.md for more information.")
   endif ()
+  if (USE_MPI_F08)
+    if (NOT MPI_Fortran_HAVE_F08_MODULE)
+      message(
+        WARNING
```
dev-zero, line 204: Please make this a hard fail. I always find it irritating if a software does some kind of fallback contradicting my explicit intent.

dev-zero (changes requested):

CMakeLists.txt
alazzaro: @fstein93 could you consider @dev-zero change in this PR? I'm happy to merge it afterwards...
For the tests, I will open a specific issue and rely on CP2K tests for the time being...

dev-zero:
wrt tests: I have no issue to hard enable it and stop testing the old wrapper. What is the issue with MPICH and gfortran 11/12?

CMakeLists.txt
fstein93: I will take care of it at the end of this week or begin of next week as I am currently relocating.

fstein93:
> wrt tests: I have no issue to hard enable it and stop testing the old wrapper. What is the issue with MPICH and gfortran 11/12?

It is related to [this bug report](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108056). I first attempted to test it with MPICH/GFortran when I implemented it into CP2K cp2k/cp2k#2486, too. The code compiles but then failes as soon as any kind of communication is initiated. I have never tested it again with later versions of GFortran.

CMakeLists.txt
alazzaro: OK, I did the change in the #687 
Thanks @fstein93 !

Issue #205, opened 2019-07-08
Explicitly reset the active GPU device


oschuett:
Currently, DBCSR assumes that the active GPU device never changes after `dbcsr_init_lib()` has been called. With the arrival of more GPU accelerated libraries in CP2K this assumption will likely break. For example running with two K80 devices lead to several [failures](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-k80-df1fc347_report.txt) which are probably related to interference between SIRIUS and DBCSR.

alazzaro:
That's a good idea... I will include it in v2.0

alazzaro:
Back to this issue, currently we reset the GPU before each multiplication, see

https://github.com/cp2k/dbcsr/blob/f35f901e4460980aa06757294463a1e6308f8dc9/src/mm/dbcsr_mm.F#L429

Should we reset for errors before any GPU calls?

oschuett:
No, it should be enough to call `dbcsr_acc_clear_errors()` only once when DBCSR gains control.
At the same time you should also call `dbcsr_acc_set_active_device()`. Basically, don't assume that the GPU's state is preserved across different DBCSR calls.

Currently, `dbcsr_acc_set_active_device()` is a public method, which is called by CP2K.  However, the trend e.g. with SIRIUS, seems to be that the libraries decide on their own how to assign GPUs to MPI ranks. I think, this indeed the right way to go.

PR #684, opened 2023-06-29
acc: Always set active device before accessing a GPU

oschuett:
I've had for a while the problem that some CP2K regtests would fail with `cudaErrorInvalidResourceHandle` when I ran on two GPUs. I believe this is because the active device gets changed between calls to DBCSR.

Presumably, most users are not affect by this because Slurm et al. set `CUDA_VISIBLE_DEVICES` to pin GPUs to CPU sockets.

### DIFF
--- a/src/acc/dbcsr_acc_devmem.F
+++ b/src/acc/dbcsr_acc_devmem.F
@@ -24,6 +24,8 @@ MODULE dbcsr_acc_devmem
                                acc_stream_cptr, &
                                acc_stream_synchronize, &
                                acc_stream_type
+   USE dbcsr_acc_device, ONLY: dbcsr_acc_set_active_device
+   USE dbcsr_config, ONLY: get_accdrv_active_device_id
 #include "base/dbcsr_base_uses.f90"
 
    IMPLICIT NONE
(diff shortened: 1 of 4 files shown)


dev-zero:
> Presumably, most users are not affect by this because Slurm et al. set `CUDA_VISIBLE_DEVICES` to pin GPUs to CPU sockets.

You would certainly wish for that, but no. The default slurm setting with gres is rather primitive and the more advanced one via nvml, rsmi or openapi which would do this affinity pinning is only available if slurm is built against the respective gpu runtime.

hfp:
I think this does not happen anymore. Rank-device combination was stable in recent versions. However and as Ole said, multi-GPU did not even work in the past (let alone d2d comm), and I used this [script](https://github.com/hfp/xconfigure/blob/master/config/cp2k/multirun.sh) in the past to prefix the CP2K-executable (it uses environment variables exposed by OpenMPI or MPICH/2 to round-robin assign devices). On the other hand, a change in DBCSR can be useful to affinitize devices if the job manager lacks configuration (or is not used). It can be intriguing without tools to find which socket a rank belong to, and which device belongs to a socket, etc.

oschuett:
> The default slurm setting with gres is rather primitive...

Interesting! Then it's indeed strange that not more users have run into this. 

What do you think about this PR? Instead of setting the active device before every call to `acc_interface_*` we could just do it once at the beginning of `dbcsr_multiply`.

alazzaro:
@oschuett I will take a look at the PR next week and hopefully close everything to make a new DBCSR release...

alazzaro:
OK, I think this is a reasonable change and it is going in the direction of the issue https://github.com/cp2k/dbcsr/issues/205

PR #700, opened 2023-09-08
Honor initial CMAKE_Fortran_FLAGS/FFLAGS

opoplawski:
We need to be able to specify extra  Fortran flags.


dev-zero:
@opoplawski ACK, but can you please add this change to the other compiler sections as well?

dev-zero:
Please note that you can always define your own CMake build type as described in the DBCSR manual to add custom flags. But I see why it may be annoying if DBCSR ignores `FFLAGS` completely while all other packages usually take those into account.

opoplawski:
All compilers updated.  Thanks.

PR #701, opened 2023-09-08
Allow specification of an alternate Fortran module install directory

opoplawski:


dev-zero:
@opoplawski ACK, but please add a comment referencing https://gitlab.kitware.com/cmake/cmake/-/issues/19608 and a comment that this is not an official CMake var (yet)

opoplawski:
Comment added.

dev-zero:
awesome, thanks a lot!

PR #767, opened 2024-03-13
Avoid unused const variable/verbose_print warning

hfp:
* Almost false positive aka incorrect warning.
* Shall be fixed regardless.

### DIFF
--- a/src/acc/cuda_hip/acc_dev.cpp
+++ b/src/acc/cuda_hip/acc_dev.cpp
@@ -20,11 +20,6 @@
 #include <stdio.h>
 #include <math.h>
 
-// for debug purpose
-#if defined(__HIP_PLATFORM_NVCC__)
-static const int verbose_print = 1;
-#endif
-
 /****************************************************************************/
 extern "C" int c_dbcsr_acc_get_ndevices(int* n_devices) {
   ACC_API_CALL(GetDeviceCount, (n_devices));
@@ -49,10 +44,8 @@ extern "C" int c_dbcsr_acc_set_active_device(int device_id) {
   // establish context
   ACC_API_CALL(Free, (0));
 
-#if defined(__HIP_PLATFORM_NVCC__)
-  if (verbose_print) {
-    ACC_API_CALL(DeviceSetLimit, (ACC(LimitPrintfFifoSize), (size_t)1000000000));
-  }
+#if defined(__CUDA) || defined(__HIP_PLATFORM_NVCC__)
+  ACC_API_CALL(DeviceSetLimit, (ACC(LimitPrintfFifoSize), (size_t)1000000000));
 #endif
 
   return 0;


alazzaro:
@hfp no sure I understand your change. Originally that function didn't exist with HIP on AMD. I think it exists so we can remove the entire condition...

alazzaro:
Correction, it still [not supported](https://docs.amd.com/projects/HIP/en/latest/doxygen/html/group___global_defs.html#ga02ceb1513c852c4dd1ecf3cc459fda70) with HIP on AMD.

But then the proper condition should include NVIDIA too, something like:
```
#if defined(__CUDA) || defined(__HIP_PLATFORM_NVCC__)
```
and you can directly remove the entire condition...

src/acc/cuda_hip/acc_dev.cpp
```diff
@@ -21,8 +21,8 @@
 #include <math.h>
 
 // for debug purpose
-#if defined(__HIP_PLATFORM_NVCC__)
-static const int verbose_print = 1;
+#if defined(__HIP_PLATFORM_NVCC__) && !defined(ACC_DEV_PRINT_VERBOSE) && 0
+#  define ACC_DEV_PRINT_VERBOSE
```
alazzaro, line 25: I will remove this part entirely....

src/acc/cuda_hip/acc_dev.cpp
```diff
@@ -49,10 +49,8 @@ extern "C" int c_dbcsr_acc_set_active_device(int device_id) {
   // establish context
   ACC_API_CALL(Free, (0));
 
-#if defined(__HIP_PLATFORM_NVCC__)
-  if (verbose_print) {
-    ACC_API_CALL(DeviceSetLimit, (ACC(LimitPrintfFifoSize), (size_t)1000000000));
-  }
+#if defined(__HIP_PLATFORM_NVCC__) && defined(ACC_DEV_PRINT_VERBOSE)
+  ACC_API_CALL(DeviceSetLimit, (ACC(LimitPrintfFifoSize), (size_t)1000000000));
```
alazzaro, line 53: Here I will add a single:
```
#if defined(__CUDA) || defined(__HIP_PLATFORM_NVCC__)
ACC_API_CALL(DeviceSetLimit, (ACC(LimitPrintfFifoSize), (size_t)1000000000));
#endif
```

alazzaro (commented):
follow-up to my last comment, I think we can simplify the code even further...

src/acc/cuda_hip/acc_dev.cpp
hfp: Thanks, will do it accordingly! I simply overlooked the whole condition was meant to actually call this function (I was somehow convinced it's debug-only code).

alazzaro: It comes to plan with only debugging, meaning to restrict the printf on the other kernels where we use the debugging (e.g. https://github.com/cp2k/dbcsr/blob/46c43eda9914f76cb059a816a9439a3eb651070b/src/acc/cuda_hip/acc_event.cpp#L22). Of course, we do set it to `0`.
I think the entire `verbose_print = 1` in this file is really misleading, so better to remove it. Thanks!

hfp:
This PR seems to cause:

>   CUDA RUNTIME API error: DeviceSetLimit failed with error cudaErrorInvalidValue.

( tested on H100 device )

alazzaro:
Is this only H100? Maybe something has changed there?

PR #782, opened 2024-04-22
ocl: demoted verbose messages to higher verbosity-level

hfp:
* Adjusted an internal setting (ZEX_NUMBER_OF_CCS).


hfp:
( If Daint-CI passes, the PR will be merged; see https://github.com/cp2k/dbcsr/issues/783. )

PR #789, opened 2024-05-10
ocl: make backend less verbose at termination

hfp:
* Made OpenCL common/shared code/definitions available.
* Updated tuned parameters.


hfp:
The precommit check seems to have some discrepancy or unwanted behavior as it complains about files which are not changed by the PR in question. Note: clang-format is not stable on a version-to-version basis and enforcing compliance for the entire code base (like files not part of the change-set) is hence not reliable.

PR #800, opened 2024-05-30
ocl: changed license to BSD-3-Clause

hfp:
* Harmonize with CP2K/offload interface/component.


hfp:
I guess the change has no further comments, and I will rebase it asap. A future step can be about moving the code into a separate repository (likely not CP2K's repository).

alazzaro (approved):

alazzaro:
LGTM

hfp:
Thanks! In exchange, I will get CMake to download external dependencies like LIBXSMM and/or a separate OpenCL layer smoothly plus options to customize like relying on prebuilt/local dependencies.

Also, I did not forget the LIBXSMM@ARM: if it's "urgent", you can use LIBXSMM `379d90b6e55c4dd9263574d2066e9b038cb628ea` and it should be fine prior to any "LIBXSMM 2.0".

PR #806, opened 2024-06-17
Initialize member variables of derived types

fstein93:
This PR adds default initializers to all types.

### DIFF
--- a/src/block/dbcsr_block_access.F
+++ b/src/block/dbcsr_block_access.F
@@ -119,15 +119,15 @@ MODULE dbcsr_block_access
    LOGICAL, PARAMETER, PRIVATE :: detailed_timing = .FALSE.
 
    TYPE block_parameters
-      LOGICAL :: tr
-      INTEGER :: logical_rows, logical_cols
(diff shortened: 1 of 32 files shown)


alazzaro:
@fstein93 it appears that this PR doesn't compile with GCC 8 (see https://storage.googleapis.com/cp2k-ci/dashboard_gcc8_report.txt). Any workaround? (I would personally drop GCC 8...)

oschuett:
I wouldn't mind dropping GCC 8 either.

fstein93:
I think Daint runs with an earlier subversion off GCC9, so GCC 8 may cover these kinds of checks. But is up to you to decide on your tests.
The workaround for GCC 8 is similar to [this](https://github.com/cp2k/cp2k/blob/aaf81e18ef1246581368ce870c691a5fa5bac5d3/src/mp2_types.F#L324).

fstein93:
In certain cases, I have to drop the initialisers entirely as GCC8 has issues with some of them. I will drop them with GCC8 in these cases.

PR #813, opened 2024-07-01
Fix GCC8

fstein93:
Follow up to #806


hfp:
FYI, there are macros to support version-checks; see https://github.com/cp2k/dbcsr/blob/develop/src/base/dbcsr_base_uses.f90#L45 (the comment gives an example for GCC and makes typical checks less intriguing ;-)

hfp:
LGTM. However, I wonder how this interferes with the pending decision to drop GCC8?

Given this change, GCC8 can be kept.

Issue #783, opened 2024-04-22
Consider to drop -Werror for tests/configs pulling external APIs/frameworks


hfp:
As we pull latest bits/dependencies into tests, our CI is prone to break automatically. For example, some recent ROCm deprecation seems to cause warnings (turned into errors). This may be wanted but if not, `-Werror` may be disabled for certain configs.

> In file included from /__w/dbcsr/dbcsr/src/acc/cuda_hip/acc_init.cpp:13:
/__w/dbcsr/dbcsr/src/acc/cuda_hip/acc_init.cpp: In function ‘int c_dbcsr_acc_init()’:
/__w/dbcsr/dbcsr/src/acc/cuda_hip/acc_init.cpp:30:40: error: ‘hipError_t hipDevicePrimaryCtxRetain(ihipCtx_t**, hipDevice_t)’ is deprecated: This API is marked as deprecated and may not be supported in future releases. For more details please refer https://github.com/ROCm/HIP/blob/develop/docs/reference/deprecated_api_list.md [-Werror=deprecated-declarations]
   30 |   ACC_DRV_CALL(DevicePrimaryCtxRetain, (&ctx, acc_device));
/__w/dbcsr/dbcsr/src/acc/cuda_hip/../hip/acc_hip.h:33:35: note: in definition of macro ‘HIP_API_CALL’
   33 |     hipError_t result = ACC(func) args; \
      |                                   ^~~~
/__w/dbcsr/dbcsr/src/acc/cuda_hip/acc_init.cpp:30:3: note: in expansion of macro ‘ACC_DRV_CALL’
   30 |   ACC_DRV_CALL(DevicePrimaryCtxRetain, (&ctx, acc_device));
      |   ^~~~~~~~~~~~
In file included from /opt/rocm-6.1.0/include/hip/hip_runtime.h:70,
                 from /__w/dbcsr/dbcsr/src/acc/cuda_hip/../hip/acc_hip.h:13,
                 from /__w/dbcsr/dbcsr/src/acc/cuda_hip/acc_init.cpp:13:
/opt/rocm-6.1.0/include/hip/hip_runtime_api.h:5128:12: note: declared here
 5128 | hipError_t hipDevicePrimaryCtxRetain(hipCtx_t* pctx, hipDevice_t dev);
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /__w/dbcsr/dbcsr/src/acc/cuda_hip/acc_init.cpp:13:
[...]

PR #840, opened 2024-09-05
Avoid Ctx API in HIP backend

gsitaram:
Tested some unit tests and perf_test and they seem to run fine without these API calls on AMD GPUs. Context APIs are now deprecated in newer versions of ROCm, and functionality is satisfied by hipSetDevice() API call. Please review and merge.
cc: @alazzaro


alazzaro:
@gsitaram thanks for this PR, however I would like to ask you 2 changes:

- we need this change only for recent HIP versions (according to https://github.com/cp2k/dbcsr/issues/783#issuecomment-2070790452 this is starting with version 6.1)
- we can revert the change I did at https://github.com/cp2k/dbcsr/pull/792, namely we can remove those lines 

https://github.com/cp2k/dbcsr/blob/3d5549904aaedb5191b3195107d491443fb1ce6c/cmake/CompilerConfiguration.cmake#L54-L56

gsitaram:
@alazzaro, why do we need these ctx APIs for any version of ROCm? I would like to avoid those APIs completely. In that case, only unwinding your fixes in [https://github.com/cp2k/dbcsr/pull/792](https://github.com/cp2k/dbcsr/pull/792)** would be sufficient, and I have done that now.

alazzaro:
> @alazzaro, why do we need these ctx APIs for any version of ROCm? I would like to avoid those APIs completely. In that case, only unwinding your fixes in #792** would be sufficient, and I have done that now.

Personally, I have no idea what those functions do. There were there for the AMD backend since ages, so my assumption was that we need them for the Rocm 5.x... Now, if you tell me that we can drop them for all ROCM versions, then I'm fine with that ;)

gsitaram:
Thank you, @alazzaro . I think we should be fine. Let me know if you encounter issues that may be related to this.

Issue #841, opened 2024-09-10
Slow memory management on Nvidia GPUs


fstein93:
If a DBCSR-heavy calculation in CP2K (LS_SCF) is profiled on NVIDIA GPUs, it turns out that DBCSR spends a lot (most) of the time on allocating/freeing memory on GPUs (tested on H100). PM for additional data. Potentially, this may also be the case on AMD hardware.

hfp:
I have recently ran tests on a GH200 system with the OpenCL backend. The OpenCL backend has support for profiling results to appear in DBCSRs / CP2Ks regular profile (end of execution). The allocations were visible for both host- and GPU-backed memory. Though, this can also depend on the node's configuration like amount of page-lockable memory, etc. Still, the time spent was relatively negligible compared to the total time to solution spent (wall time).

hfp:
These are the prototypes that allow to call CP2K/DBCSR's timer facility for instance in `cuda_hip` sources:
https://github.com/cp2k/dbcsr/blob/develop/src/acc/acc.h#L67-L68

fstein93:
@alazzaro I have sent you a mail.

alazzaro:
> @alazzaro I have sent you a mail.

OK, so I've checked the slides and my understanding is that the problem is appearing on the first multiplications, which is expected. 
We do use [memory pools](https://github.com/cp2k/dbcsr/blob/f4e8c38dd79dedfe63a9615359ddccfa222e08e4/src/acc/dbcsr_acc_devmem.F#L179) with a factor size of 1 (if I recall correctly, we do use memory pools on the CPU too, with a factor size increase of [1.2](https://github.com/cp2k/dbcsr/blob/f4e8c38dd79dedfe63a9615359ddccfa222e08e4/src/core/dbcsr_config.F#L28)). The main function which does memory settings (pools and allocations) is 

https://github.com/cp2k/dbcsr/blob/f4e8c38dd79dedfe63a9615359ddccfa222e08e4/src/data/dbcsr_mem_methods.F#L207

Then, there is a [function](https://github.com/cp2k/dbcsr/blob/f4e8c38dd79dedfe63a9615359ddccfa222e08e4/src/data/dbcsr_data_methods.F#L215) to ensure that size of the buffers is OK.

The part where this function is called is for the C matrix:

[...]

PR #846, opened 2024-09-13
ocl: speedup host-memory allocation on GH200

hfp:


hfp:
This change reduces the duration to allocate a single buffer by approximately 3.5s (tested on GH200 system). No other system has ever shown such high allocation cost for this particular implementation of `c_dbcsr_acc_host_mem_allocate`.

alazzaro:
out of curiosity, is this something related to https://github.com/cp2k/dbcsr/issues/841 ?
I'm checking an idea on how to make the memory pool more effective by introducing a configuration parameter to set a "minimal" allocation size. I will also add some statistics on memory allocations, although we can check the size of the buffers via the MPI size messages...

hfp:
> out of curiosity, is this something related to #841 ? I'm checking an idea on how to make the memory pool more effective by introducing a configuration parameter to set a "minimal" allocation size. I will also add some statistics on memory allocations, although we can check the size of the buffers via the MPI size messages...

This is not directly related (or I did not try to address #841), and I think #841 used the CUDA-based implementation. I have yet to try my reproducer (`acc_bench_smm`) on an Alps-like node using the CUDA backend (I will report back). It's possible the CUDA based host-memory allocation suffers a similar slowness. Though, #841 states "tested on H100" but we also call the H100 tuned parameters "H100" even when collected on GH200.

PR #851, opened 2024-09-24
Specify version of Clang-format

hfp:
- clang-format is not stable version-to-version
- Hook is reformatting code already committed,  
  it may be better to only change touched code


hfp:
If this becomes "green", I will merge the PR. However, `.clang-format` shall be adjusted in a follow-up change. Apparently, not even this config is digested in a stable fashion including errors for certain configuration keys.

hfp:
Syntax of `additional_dependencies` seems to be **not** as documented but rather follows Pip's version spec (`clang-format~=19.1.0`).

PR #874, opened 2024-12-09
Make G2G as default during the compilation...

alazzaro:
.. and control it at runtime via `DBCSR_USE_ACC_G2G` env variable (OFF by default).


alazzaro:
FYI @gsitaram

src/mm/dbcsr_mm.F
```diff
@@ -921,12 +920,6 @@ SUBROUTINE dbcsr_multiply_generic(transa, transb, &
                                      filter_eps=filter_eps, &
                                      flop=my_flop, keep_product_data=keep_product_data)
          END IF
-#else
```
gsitaram, line 924: @alazzaro , are you sure this change removing the else part is ok? Is it ok for some jobs to crash because this code path only supports real_8 data?

alazzaro: We still have the else... it was there even before:
```
         IF (data_type .NE. dbcsr_type_real_8 .OR. (.NOT. dbcsr_cfg%use_acc_g2g%val)) THEN
            ! If G2G is enabled, norms have to be calculated on the GPU.
            ! Since the norms kernel expects only real_8 type data, we
            ! avoid using G2G for all other data types
            CALL multiply_cannon(m2s_left, m2s_right, product_matrix, &
                                 retain_sparsity=retain_sparsity, &
                                 filter_eps=filter_eps, &
                                 flop=my_flop, keep_product_data=keep_product_data)
         ELSE
            CALL multiply_cannon_g2g(m2s_left, m2s_right, product_matrix, &
                                     retain_sparsity=retain_sparsity, &
                                     filter_eps=filter_eps, &
                                     flop=my_flop, keep_product_data=keep_product_data)
         END IF
```
Am I right?

gsitaram: Yes you are absolutely right. I didn't read the changed file carefully. Thanks @alazzaro!

hfp:
Quick Q, is G2G only hinging on GPU-aware MPI?

alazzaro:
> Quick Q, is G2G only hinging on GPU-aware MPI?

Yes, exactly. 
The feature is now promoted to be "a runtime flag", but it is still experimental (there are few things to consider). It will be officially released in 2025...

hfp:
> > Quick Q, is G2G only hinging on GPU-aware MPI?
> 
> Yes, exactly. The feature is now promoted to be "a runtime flag", but it is still experimental (there are few things to consider). It will be officially released in 2025...

Thanks!

I see `__DBCSR_ACC_G2G` also requires to calculate norms on GPU (to keep data in place). In general, I wonder if G2G would work in any case or if there missing transfers. Without norms on GPU, I can think some transfers are missing but is there anything else?

I consider implementing norms on GPU for OpenCL too. I think all contemporary MPIs have GPU support if say pointers are registered, etc.

alazzaro:
> 
> I see `__DBCSR_ACC_G2G` also requires to calculate norms on GPU (to keep data in place). In general, I wonder if G2G would work in any case or if there are missing transfers. Without norms on GPU, I can think some transfers are missing, but is there anything else?

The short answer is: no, it would not work in any case, that's why it is still experimental. For this reason, I've added 31dc51a29d9243cc1d1cb08ad59f3bd6fd8a3773. Indeed, there can be cases where the kernels are too big (for instance, 50x50), so no kernel will be jitted and the library will fall-back to the CPU (without host data!) and fail...

The norms are only one part of the story (and only relevant when we do apply filtering). Another thing @gsitaram introduced: with G2G we move the B-transposed data, so we dont' need to run B-transpose for each step of the multiplication.

Overall, the speed-up on LUMI was quite significant for the H2O-DFT-LS...

alazzaro:
> 
> I see `__DBCSR_ACC_G2G` also requires to calculate norms on GPU (to keep data in place). In general, I wonder if G2G would work in any case or if there are missing transfers. Without norms on GPU, I can think some transfers are missing, but is there anything else?

The short answer is: no, it would not work in any case, that's why it is still experimental. For this reason, I've added 31dc51a29d9243cc1d1cb08ad59f3bd6fd8a3773. Indeed, there can be cases where the kernels are too big (for instance, 50x50), so no kernel will be jitted and the library will fall-back to the CPU (without host data!)...

Then, another thing @gsitaram introduced 

The norms are only one part of the story

hfp:
> Overall, the speed-up on LUMI was quite significant for the H2O-DFT-LS...

Good to know! I will try to get my way through it at some point.

PR #877, opened 2024-12-23
Check block sizes are equal in dbcsr_copy_into_existing_sym

oschuett:


src/ops/dbcsr_operations.F
```diff
@@ -1479,6 +1479,9 @@ SUBROUTINE dbcsr_copy_into_existing_sym(matrix_b, matrix_a)
          IF (.NOT. found) THEN
             CALL dbcsr_data_clear(dst_data)
          ELSE
+            IF (dbcsr_data_get_size(src_data) .NE. dbcsr_data_get_size(dst_data)) THEN
```
hfp, line 1482: I wonder if it's too strict like `.LE` instead of `.NE`? Though, the code in this branch performs `dbcsr_block_partial_copy`.

oschuett: The same check is performed in [dbcsr_copy_into_existing](https://github.com/cp2k/dbcsr/blob/2cdbf1331eb9f0e8265f4ca048cedb125f3e38f9/src/ops/dbcsr_operations.F#L1402) and the regtests are passing. So, I'm fairly sure that this is not an intentional feature.


I stumbled across this by trying to restart an LS_SCF calculation from a density matrix of a different basis set. Instead of complaining about the mismatching block sizes CP2K happly went on to compute garbage :-/

hfp: Thank you for looking into it. LGTM then!

hfp: Ready to be merged, I guess.

alazzaro:
thanks @oschuett . Yes, there are tons of cases where DBCSR assumes things are just right (see for example #37 ). In some cases, the checks are on the critical path, so we need a better way of checking stuff...

PR #878, opened 2025-01-06
ocl: minor fixes and adjustments

hfp:
* Adopt best-matching device even for device-0 if ACC_OPENCL_ACTIVATE is not set or no device is activated.
* Fix: allow zero-sized allocation (c_dbcsr_acc_dev_mem_allocate).
* Improved finding OpenCL header file (CL/cl.h) in Makefile.
* Refactored initialization into init and pre-init.
* Minor rework of environment setup.
* Updated tuned parameters.
* Improved verbose output.


hfp:
Something changed upstream and is addressed by #879. As a side-note, this is another instance of a not-exactly-reproducible test (since it depends on"latest" release rather than being pinned).

PR #879, opened 2025-01-06
HIP: removed unused hipHostAllocDefault

hfp:
- Apparently, some API elements are "mirrored" (TODO: confirm why necessary).
- In particular, "hipHostAllocDefault" now changed "upstream" hence clash.
- Cleanup: removed duplicated prototype (hipHostAlloc).

### DIFF
--- a/src/acc/hip/acc_hip.cpp
+++ b/src/acc/hip/acc_hip.cpp
@@ -15,8 +15,6 @@ hipError_t hipHostAlloc(void** ptr, size_t size, unsigned int flags) { return hi
 hipError_t hipFreeHost(void* ptr) { return hipHostFree(ptr); }
 #endif
 
-unsigned int hipHostAllocDefault = hipHostMallocDefault;
-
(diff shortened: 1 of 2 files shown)


alazzaro:
@gsitaram FYI 
Any comment?

hfp:
( I believe, it was an unused variable [not used anywhere] which then turned into an enum upstream like name clash here )

alazzaro:
> ( I believe, it was an unused variable [not used anywhere] which then turned into an enum upstream like name clash here )

I tend to agree, I think it was there because of an equivalent CUDA version (no sure, need to look at the history). Anyway, the PR is just fine ;)

hfp:
> @alazzaro , @hfp , thank you for addressing this and apologies for missing it. It looks like this change did not make it to the formal DBCSR release that was made recently, am I right?

Yes, you are correct. This change did not make it into v2.8.0.

williamyxl:
Ah, that's why compiler is complaining a HIP defined macro (rocm>=6.3.0) with the same name was being declared as a variable in DBCSR 2.8.0. https://github.com/spack/spack/issues/49544

This is great, looking forward to this update in the next release. Thanks.

PR #890, opened 2025-02-14
cmake: removed standard modules

hfp:
- Minor beautification (CMakeLists.txt).


alazzaro:
@dev-zero could you take a look?
I can remember we had a good reason for using FindBlas, don't we?

hfp:
@dev-zero ( I will do the same for CP2K if you confirm there is no current reason )

dev-zero:
yes, the respective commit message https://github.com/cp2k/dbcsr/commit/f9dca0730aa891273d06aa131321797ab2708724 has the details.

dev-zero:
@hfp if you have a better idea or if this has been fixed upstream, please go ahead. If not, pulling the most recent versions from upstream is the way to go.

hfp:
Thank you all!

( I come back to this PR next week )

dev-zero:
> The FindLAPACK module from upstream changed but it looks like it can be used "as is" (like only its local presence is needed in order to find the locally modified FindBLAS).

correct

> The FindBLAS module from upstream changed significantly as well, however it seems unnecessary work to replicate our patch. I will keep both modules as they are since they seem to work properly in this scope. CP2K's solution btw (which also fixes the OpenMP glitch with OpenBLAS), is yet again of different origin like supplied by some ETHz contribution (looks like no common origin with CMake/upstream).

Ok. At one time I wanted to open a PR at upstream to send the required changes upstream but did not have time to go through all cases and consider a proper design.
Maybe now that upstream has changed their code it might be worth to take a look again.

PR #915, opened 2025-05-07
Fixed looking-up bogus environment variables

hfp:
- Correctly stringize the macro argument instead of relying on wrong behavior.
- The parameter_name argument is expanded inside of quotes (GNU toolchain).
- Disabled fprettify for macro-definition.


alazzaro:
could you give an example of what's wrong here?

src/base/dbcsr_base_uses.f90
```diff
@@ -49,6 +49,15 @@
 #define TO_VERSION3(MAJOR, MINOR, UPDATE) (TO_VERSION2(MAJOR, MINOR) + (UPDATE))
 #define TO_VERSION TO_VERSION2
 
+! Rely on incorrect expansion inside of quotes to avoid issue of handling hash-character.
+!&<
+#if defined(__GNUC__)
```
alazzaro, line 54: no sure I understand this part...
GNU preprocessor is different? This code has been used since a while now, is it something with the new GNU versions?

alazzaro: wow, just found this https://gcc.gnu.org/legacy-ml/fortran/2009-06/msg00131.html...

alazzaro:
LGTM

hfp:
Look at:  
`#define MACRO(ARG) "ARG"`
ARG is *not* supposed to expand inside of quotes. GNU toolchain incorrectly does that. The correct way is to "stringize" ARG using the leading hash sign (just like in C/C++). However, that screws up entire Fortran compilation when on GNU toolchain.

What's solved here is considering the environment variables, e.g., DBCSR_RUN_ON_GPU. For instance Intel Fortran compiler (using the correct macro expansion), failed to look for "DBCSR_RUN_ON_GPU" (`get_environment_variable`) and instead looked up "DBCSR_parameter_name".

src/base/dbcsr_base_uses.f90
hfp: Well, it quietly failed. You would not note if the applications looks up the wrong environment variable. Also, we do not test so many non-GNU Fortran compilers...

hfp:
( This is sort of a "horrible" bug in GNU toolchain )

PR #925, opened 2025-06-07
Update c++ standard

mtaillefumier:
Use the cmake mechanisms to set the c++, c, cuda and hip standards


alazzaro:
OK, I'm checking this PR. Is there any reason why we want to enable:

```
  enable_language(CXX)
  enable_language(C)
```

even if there is no GPU requested support?

PR #931, opened 2025-07-04
CMake/LAPACK/BLAS: removed customized/diverged standard modules

hfp:
- Related: https://github.com/cp2k/dbcsr/commit/f9dca0730aa891273d06aa131321797ab2708724
- Changes like picking the correct OpenBLAS threading layer must be upstreamed (CMake).
- It can be considered micro-management to prevent inconsistent choices/pickup.
- Spack solves this with threads=openmp variant and prevents inconsistency.
- Current CMakeFile, standard modules, and Spack are "good enough".


alazzaro:
unrelated question: 
@hfp do we need to make a new release of DBCSR for the new CP2K?

hfp:
> unrelated question: @hfp do we need to make a new release of DBCSR for the new CP2K?

I think so. Perhaps @oschuett can shed some light on the timeline? I think CMake and Spack packages can still use a bit more time. For instance, I started polishing Spack packages for DBCSR and CP2K. Regarding CMake, I think CP2K's custom modules for all LAPACK/BLAS shall go away for same reasons like this PR. While I am nearly complete with Spack packages, I have only started today with the mentioned removal for CP2K.

hfp:
RPM builds are failing because of:
> rm cmake/FindBLAS.cmake cmake/FindLAPACK.cmake
rm: cannot remove 'cmake/FindBLAS.cmake': No such file or directory
rm: cannot remove 'cmake/FindLAPACK.cmake': No such file or directory

This seems to be specified in `dbcsr.spec`, which is outside of this repository.

hfp:
@alazzaro let's get this reviewed on a finite timeline. On an alternative path, you can pre-release DBCSR in anticipation of an upcoming release of CP2K.

@alazzaro @oschuett  I think, this sort of coupling can drop going forward as soon as the build systems consolidate in CP2K (gmake vs cmake), and as soon as Spack adoption is covering the classic Toolchain.

On a side note, I am working on Spack recipes for CP2K and DBCSR as well as important dependencies like LIBINT and LIBXC to ensure performance and satisfaction carries forward from GNU Make and Toolchain.

alazzaro:
We can merge here, sorry for the long silence...

PR #957, opened 2025-09-19
HIP: fixed compilation error

hfp:
- Error: `printf` was not declared in this scope.
- Appeared when compiling cuda_hip/acc_utils.cpp.


alazzaro:
Interesting, now the CI runs ROCM 7.0 and I see some missing symbols. @gsitaram could you have a look?

alazzaro:
ops @hfp I meant to say "you can merge here" instead of close...

PR #967, opened 2025-10-20
Introduce accelerator-centric mode (acc_centric) for GPU-to-GPU communication via GPU-aware MPI in multiply_cannon (multiply_cannon_g2g)

franzpoeschel:
This PR adds a new configuration option ACC_CENTRIC for `multiply_cannon_g2g()`. This flag enables:

1. Use of GPU-oriented data layout (transposes buffers)
2. Use of peer-to-peer GPU communication via GPU-aware MPI (HIP or CUDA)

Note: This is based on quite an old branch that was never merged back into DBCSR. Nearly every single line ran into merge conflicts when updating to recent dev versions of DBCSR; my main contribution in here is resolving these merge conflicts. I can at the moment not confidently explain every last line of this patch.

TODO:

- [ ] I am currently running CP2K's regtests against this, see how they go
- [ ] Since this is a performance optimization, measure the performance
- [ ] See in-code comment below on assumptions that this patch makes

### DIFF
--- a/src/mm/dbcsr_mm_cannon.F
+++ b/src/mm/dbcsr_mm_cannon.F
@@ -840,9 +840,9 @@ CONTAINS
                               retain_sparsity, &
                               filter_eps, flop, keep_product_data)
       !! Multiplies two DBCSR matrices
-   USE iso_c_binding, ONLY: c_f_pointer
-   USE dbcsr_acc_devmem, ONLY: acc_devmem_cptr
-   USE dbcsr_acc_event, ONLY: acc_event_record
+      USE iso_c_binding, ONLY: c_f_pointer
+      USE dbcsr_acc_devmem, ONLY: acc_devmem_cptr
+      USE dbcsr_acc_event, ONLY: acc_event_record
 
       TYPE(dbcsr_2d_array_type), POINTER                 :: left_set, right_set
          !! set of imaged left matrices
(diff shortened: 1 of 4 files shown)


src/mm/dbcsr_mm_cannon.F
```diff
@@ -1329,6 +1361,73 @@ SUBROUTINE multiply_cannon(left_set, right_set, product_matrix, &
...
+         END DO
+         DO v_ki_left = 1, left_col_nimages
+            CALL dbcsr_data_host2dev(left_buffer_calc%mats(1, v_ki_left)%data_area)
+         END DO
+
+         v_ki_left = 1
+         v_ki_right = 1
+         DO v_ki = 1, max(left_col_nimages, right_row_nimages)
+            IF (v_ki_right .GT. right_row_nimages) THEN
+               DBCSR_ABORT("Failed assumption left_col_nimages >= right_row_nimages")
```
franzpoeschel: In my tests, I was able to trigger this failure. I am not sure how many real-world use cases this assumption excludes from `acc_centric` mode, might want to consider the opposite case, too.

alazzaro:
BTW, are you aware of any MPICH which supports GPU-aware for HIP? 
Unfortunately, we don't have a way to test G2G with the CI...

franzpoeschel:
I have run some benchmarks with this now on two local dev systems and this PR seems to incur a noticeable slowdown in all tests that I ran.
I will run some tests again on an actual HPC system in time, but I am closing this for now.

PR #971, opened 2025-10-29
Deallocate arrays

hfp:
- Not actually necessary (modern Fortran).


alazzaro:
yes, allocatable variables are released when they go out of the scope... so why this change is needed?

hfp:
I had it around for long time and it is now only for consistency like every ALLOCATE get DEALLOCATE. Originally, I wanted to reduce the number of allocated arrays, but code looked to bad when lumping arrays together. What remains (beside of consistency), is deallocation happens likely earlier given automatic cleanup is "at the end of the subroutine".

PR #978, opened 2025-12-10
ocl: support shared virtual memory

hfp:
- SVM can be absent in OpenCL 3.0 (based on OpenCL 2.0).
- System-level fine-grained buffers are prepared.


hfp:
Let me aim for early afternoon (today).

PR #979, opened 2025-12-11
Switch MacOS CI to use MPICH

alazzaro:
Update GCC to v15 to be compatible with mpi.mod file.

### DIFF
--- a/.github/workflows/testing-linux.yml
+++ b/.github/workflows/testing-linux.yml
@@ -98,7 +98,9 @@ jobs:
 
     strategy:
       matrix:
+        use_mpi: [MPI=ON, MPI=OFF]
         use_openmp: [OPENMP=ON]
+        mpi_suffix: [mpich]
 
     steps:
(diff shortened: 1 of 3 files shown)


alazzaro:
Fedora failures are due to https://github.com/cp2k/dbcsr/commit/01a76b27fc6b4a02eacfc3b0654d3bb13d5daa92#commitcomment-172602840

hfp:
> Fedora failures are due to [01a76b2#commitcomment-172602840](https://github.com/cp2k/dbcsr/commit/01a76b27fc6b4a02eacfc3b0654d3bb13d5daa92#commitcomment-172602840)

It seems changing our installation may always break the RPM. The package definition is outside of this repository and it seems to work with globs on top of what our process installs. See also [here](https://github.com/cp2k/dbcsr/commit/01a76b27fc6b4a02eacfc3b0654d3bb13d5daa92#commitcomment-172630302).

PR #987, opened 2025-12-18
[cmake] fix hiprtc dependency

mtaillefumier:
Small fix providing hiprtc target when `find_package(DBCSR)` is called


alazzaro:
Do we need a patch release?

mtaillefumier:
Yes. I am afraid we need a patch release otherwise cp2k won't build with hip without this line. I should have spotted it earlier.

mtaillefumier:
I confirm the patch is needed to build cp2k with or without spack.

alazzaro:
No problem, doing it tonight. Thanks for the contribution!

gsitaram:
@mtaillefumier , thank you for catching that. I should not have missed this file in my update earlier this year. Also learned a lesson that I should try building CP2K too in order to test DBCSR.

Issue #873, opened 2024-12-09
OpenMP Master deprecated


alazzaro:
Follow-up of #857

PR #1009, opened 2026-05-19
Better feature-check under macOS and suppress OpenMP warning

hfp:
- Set CMAKE_TRY_COMPILE_TARGET_TYPE to STATIC_LIBRARY (macOS), compiler test during configure need no linkage.
- Save/restore CMAKE_Fortran_FLAGS around feature checks, to potentially run without -Werror flags.
- Avoid error: ‘master’ construct is deprecated use ‘masked’ (since OpenMP 5.1).
- Check -Wno-error=deprecated-openmp works even if GNU Fortran is 15.1+ (macOS).


hfp:
Related to #873.

hfp:
There were multiple issues; all fixed. The failure building RPMS is unrelated.

PR #1019, opened 2026-06-06
Fix Fedora RPM-build error

Growl1234:
- Initialize tensor-test vectors directly to avoid GCC 16 nonnull warnings from initializer-list assignment that RPM-build treats as error.
- Adjust spec file based on #1018 and considering libxs/libxstream not yet available in Fedora repo.


Growl1234:
The macOS testing failure seems to be existing before and resolved in #1018 ...

I will move these changes into that PR to see how it goes.

PR #1026, opened 2026-07-19
test: restore bounded GPU correctness tests

### LINKED ISSUE #427: tests: make libsmm_acc tests run with MPI

Stardust0831:
## Motivation

The `libsmm_acc_unittest_multiply` and `libsmm_acc_unittest_transpose` executables are built for CUDA and HIP, but their CTest registrations have been disabled because the original tests were not parallelized and were too slow (see cp2k/dbcsr#427). As a result, accelerator builds currently compile the tests without running their correctness checks.

Fixes cp2k/dbcsr#427.

## Changes

- Register the two libsmm_acc correctness tests for CUDA and HIP builds while leaving the timer/performance tests disabled.
- Bound the default smoke coverage to 64 autotuned multiplication kernels, 64 predicted kernels, and 64 transpose pairs.
- Make sampling deterministic and expose zero-valued limits as an explicit full-coverage mode.
- Initialize MPI and the accelerator in the correctness executables, select a device using the node-local MPI rank, and shard cases cyclically across global ranks.
- Aggregate correctness failures across ranks and normalize the process exit status to avoid truncating large error counts.
- Label the tests `gpu;libsmm_acc`, run them serially to avoid GPU contention, and provide configurable rank and timeout settings.

Non-MPI builds continue to execute the same tests directly with one process.
No production DBCSR kernel or accelerator code is changed.

## GitHub Actions red/green validation

I ran the same temporary self-hosted workflow against the code before and after this patch. The workflow itself is kept only on validation branches in my fork and is not part of this PR.

- **Before:** [run 29677536620](https://github.com/Stardust0831/dbcsr/actions/runs/29677536620)
  - source: upstream `develop` plus only the temporary workflow
  - configure: passed
  - build: passed
  - CTest: failed as expected with `Total Tests: 0` and `test_rc=8`
  - Slurm logs were uploaded as the `dbcsr-v100-gha-gpu-before-29677536620`
    artifact
- **After:** [run 29677660083](https://github.com/Stardust0831/dbcsr/actions/runs/29677660083)
  - source files are byte-identical to this PR patch
  - 4 NVIDIA V100 GPUs, CUDA 12.4.1, GCC 14.3, OpenMPI 5.0.10
  - configure: passed
  - build: passed
  - `libsmm_acc_unittest_multiply`: passed in 50.27 s
  - `libsmm_acc_unittest_transpose`: passed in 3.47 s
  - 2/2 tests passed; logs were uploaded as the `dbcsr-v100-gha-gpu-after-29677660083` artifact

## Additional validation

- Full V100 parameter coverage on 32 GPUs across two nodes:
  - 1,235 autotuned and 72,857 predicted multiplication kernels
  - 1,769 transpose pairs
  - zero correctness errors
- Round-robin multi-node placement on 8 GPUs across two nodes verified that node-local ranks select devices 0-3 independently on each node; both tests passed with zero errors.
- Non-MPI CUDA smoke tests passed on one V100.
- The relevant pre-commit hooks pass, including ruff, Black, Python AST, cmake-format, header checks, and clang-format-fypp.

As a separate validation of the MPI/device-sharding approach, I also ran `libsmm_acc_timer_multiply` with a benchmark-only patch on 1, 2, 4, 8, and 16 V100 GPUs. A fixed 1,024-kernel workload reached 1.98x, 3.44x, 4.82x, and 7.59x speedup, respectively, and all CPU checksum validations passed. The timer changes are not included in this PR.

CUDA runtime behavior was validated on V100 GPUs. HIP/ROCm remains covered by the existing compile CI, but I did not have an AMD GPU runner for a HIP runtime test. These are DBCSR backend correctness tests rather than CP2K application
regression inputs. The upstream GitHub Actions accelerator jobs currently compile these executables without access to a physical GPU; this PR makes the tests runnable by GPU-enabled CTest environments but does not add a permanent GPU runner.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -211,6 +211,42 @@ endif ()
 
 # =================================== GPU BACKEND TESTS (CUDA/HIP)
 
+set(DBCSR_LIBSMM_ACC_TEST_SAMPLES
+    64
+    CACHE STRING "Number of predicted libsmm_acc kernels in GPU smoke tests")
+set(DBCSR_LIBSMM_ACC_TEST_MAX_AUTOTUNED
+    64
+    CACHE
+      STRING
+      "Maximum number of autotuned libsmm_acc kernels in GPU smoke tests; zero means all"
+)
+set(DBCSR_LIBSMM_ACC_TEST_MAX_TRANSPOSE_PAIRS
+    64
+    CACHE
+      STRING
+      "Maximum number of libsmm_acc transpose pairs in GPU smoke tests; zero means all"
+)
+set(DBCSR_LIBSMM_ACC_TEST_MPI_RANKS
+    1
+    CACHE STRING "Number of MPI ranks for libsmm_acc GPU smoke tests")
+set(DBCSR_LIBSMM_ACC_TEST_TIMEOUT
+    300
+    CACHE STRING "Timeout in seconds for each libsmm_acc GPU correctness test")
+
+if (DBCSR_LIBSMM_ACC_TEST_MPI_RANKS LESS 1)
+  message(FATAL_ERROR "DBCSR_LIBSMM_ACC_TEST_MPI_RANKS must be at least 1")
+endif ()
+if (DBCSR_LIBSMM_ACC_TEST_TIMEOUT LESS 1)
+  message(FATAL_ERROR "DBCSR_LIBSMM_ACC_TEST_TIMEOUT must be at least 1")
+endif ()
+foreach (libsmm_acc_test_limit
+         DBCSR_LIBSMM_ACC_TEST_SAMPLES DBCSR_LIBSMM_ACC_TEST_MAX_AUTOTUNED
+         DBCSR_LIBSMM_ACC_TEST_MAX_TRANSPOSE_PAIRS)
+  if (${libsmm_acc_test_limit} LESS 0)
+    message(FATAL_ERROR "${libsmm_acc_test_limit} must be non-negative")
+  endif ()
+endforeach ()
+
 # Add custom commands for the test files that need to be generated from a
 # template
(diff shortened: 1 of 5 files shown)


Stardust0831:
The two Fedora Rawhide Packit failures appear unrelated to this PR. Both fail  while linking the existing `dbcsr_test` and `dbcsr_tensor_test` executables  with unresolved `c_dbcsr_*` symbols. PR #1025 shows the same failures on both Rawhide architectures. The CUDA and ROCm MPI/non-MPI build checks for this PR all pass.

Growl1234:
> The two Fedora Rawhide Packit failures appear unrelated to this PR. Both fail while linking the existing `dbcsr_test` and `dbcsr_tensor_test` executables with unresolved `c_dbcsr_*` symbols. PR #1025 shows the same failures on both Rawhide architectures. The CUDA and ROCm MPI/non-MPI build checks for this PR all pass.

I can take care of this. Similar issue also happened with [CP2K](https://github.com/cp2k/cp2k/pull/5557).

Edit: Done with #1027.

alazzaro:
@Stardust0831 thanks for your PR. I must admit, the checks of the GPU kernels are less and less important nowadays since we now provide a generic [kernel](https://github.com/cp2k/dbcsr/commit/c57a4e4f59bb177c4dd078f45537a38f0aa149df). The other limitation is that CI doesn't check GPU execution at all, so there is no way to verify it (I will do it manually in the next days). 
The only scenario where I can see those tests being useful is when people want to apply kernels for GPU X on a different GPU Y (still, this doesn't prevent side effects if the requested kernel is not present).
Please, allow me some time to check your PR.

Stardust0831:
Thanks @Growl1234 for quickly tracking this down and fixing the Fedora Rawhide/GCC 16 C API linking issue in #1027! After rebasing onto the latest `develop` and rerunning CI, all checks on this PR are now passing. Much appreciated!

PR #1027, opened 2026-07-20
Expose DBCSR C API procedures explicitly

Growl1234:
The `dbcsr_api_c` and `dbcsr_tensor_api_c` modules use `PRIVATE` by default, while the `BIND(C)` procedures declared in `dbcsr.h` and `dbcsr_tensor.h` are intended to be exported as the public C API.

With the Fedora Rawhide/GCC 16 RPM build, these symbols are present in `libdbcsr_c.so`, but they are emitted as local symbols and therefore are not available from the dynamic symbol table. As a result, linking the C++ tests against `libdbcsr_c.so` fails with unresolved references to `c_dbcsr_init_lib_internal`, `c_dbcsr_create_new`, `c_dbcsr_t_ndims`, and many other C API entry points.

This PR makes the C API procedures explicitly `PUBLIC`. This documents the intended API boundary and avoids relying on compiler-specific handling of private module procedures with C binding.

### DIFF
--- a/src/dbcsr_api_c.F
+++ b/src/dbcsr_api_c.F
@@ -26,6 +26,96 @@ MODULE dbcsr_api_c
 
    #:include 'data/dbcsr.fypp'
 
+   PUBLIC :: c_dbcsr_clear_mempools, &
+             c_dbcsr_init_lib, &
+             c_dbcsr_finalise_lib, &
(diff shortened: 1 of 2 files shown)


alazzaro:
@Growl1234 makes sense, thanks for this fix

PR #1028, opened 2026-07-20
CMake: Do not search parent directories for Git metadata

Growl1234:
When DBCSR is embedded in another source tree without its own `.git` metadata, the current implementation searches parent directories and may use the enclosing project's Git tags and revision information as DBCSR's version.

Restrict Git metadata discovery to DBCSR's current source directory. If that directory is not a Git working tree or submodule, fall back to the version information provided by the DBCSR source archive.


Growl1234:
Looks like the tests are not that robust...

alazzaro:
@mtaillefumier can you review it? thanks...

mtaillefumier (approved):
These modifications look alright.

PR #1032, opened 2026-08-09
B200 GPU Support

lclemens05:
Autotuned parameter file for NVIDIA B200 Blackwell GPU

[parameters_B200.json](https://github.com/user-attachments/files/30881241/parameters_B200.json)


alazzaro:
@lclemens05 thanks for your contribution!
could you add B200 to the [cmake list of support architecture](https://github.com/cp2k/dbcsr/blob/c67b24681d0a857f74ecb23b66d628ca2cdacaba/CMakeLists.txt#L129) and make sure they are selected during the build?

DCM-Uni-Paderborn (commented):
Thanks for adding the tuned B200 parameter set. I tested the current PR head on an NVIDIA B200 node with CUDA 12.8 while preparing a CUDA-enabled CP2K/Skala build.

Two CMake changes are still needed:

1. Please add `set(GPU_ARCH_NUMBER_B200 100)`. With the current head, configuration prints
   ```text
   GPU target architecture: B200
   Kernel parameters: B200
   GPU architecture number:
   ```
   so the `CUDA_ARCHITECTURES` target property remains empty.
2. Please keep the existing `P100` default and its help text. Selecting B200 should be explicit; changing the global default would alter existing builds unexpectedly.

It would also be useful to add `"parameters_B200.json": "sm_100"` to the `gpu_architectures` mapping in `src/acc/libsmm_acc/kernels/smm_acc.py` so the tooling recognizes the new parameter file consistently.

For comparison, a local DBCSR build with `B200 -> 100` completed successfully and was linked into CP2K. Native-grid Skala smoke tests then passed on one and four B200 GPUs, including deterministic rank-to-device mapping `0:0 1:1 2:2 3:3` and CPU/GPU agreement for energy, forces, and stress.

One independent CUDA 12.8 issue remains: `cuFuncSetSharedMemConfig` is deprecated and DBCSR promotes that warning to an error. I suggest handling that separately rather than suppressing all deprecation diagnostics in this architecture PR.

